PR #4, opened 2022-01-24
Update Energy Total Registers

britkat1980:
Updated some register values, modified tests and have run and passed them in my vscode venv. Hopefully I've done this right...


dewet22:
Very nice, looks great!

PR #6, opened 2022-02-06
Some hybrid inverters start with SD not SA

zaheerm:


dewet22:
Sorry, been traveling so only getting around to this now! Good to know about this - out of curiosity, what is the text description of your model in the GE dashboard? Mine's `Giv-HY5.0`

zaheerm:
> what is the text description of your model in the GE dashboard?

It's Giv-HY5.0 for me.

dewet22:
Interesting, so they seem to be labeled identically. I took your contribution and reworked it a bit last night – hope you don't mind or feel I appropriated it without giving credit (definitely one for the CHANGELOG) but the JSON serialisation is better now (model name instead of serial number prefix) and I fixed the lint warnings at the same time.

zaheerm:
Thanks no issues, I'm on holiday now so didn't get to the suggested linting. I read your changes and they make sense. In fact I started my changes, changing to auto() instead of keeping the strs. I just didn't know what implications it may have had elsewhere in any dependent code.

dewet22:
> It's Giv-HY5.0 for me.

Interesting – talking to the folks at GivEnergy they claim the `SA` model is the 3.6kW Hybrid inverter. Do you know which model you actually have?

zaheerm:
I have a model with a serial number starting SD and the inverter is
definitely a 5 kW inverter.

dewet22:
Ah, I got it the wrong way around - mine starts with SA but is also definitely a 5.0kW Hybrid. Odd, I'll try and understand it a bit better. Out of interest, approximately how old is yours?

zaheerm:
Mine was installed a few weeks ago.

PR #7, opened 2022-03-01
Fix for odd f/w

britkat1980:
Update to allow access to alternative battery charge and discharge energy totals on some versions of f/w


dewet22:
Thanks for this. What's the difference between these registers (IR 105 & 106) and the other battery_total_(dis)charge registers (IR 180 & 181)? My gut feeling is the former is the total for the battery pack and should be equal to the latter for single battery systems. In multi-battery systems, do the latter indicate the cumulative totals across all battery packs?

britkat1980:
For some users registers 180&181 don't return data (0 values), so they need to get battery energy details from 105&106. Not sure why, its due to variances in the firmware. GE haven't been able to really confirm what versions have which registers... So I use some logic in GivTCP to fetch from second register set if the "usual" ones don't return values

dewet22:
Thanks! Do you remember off-hand which models return which values?

britkat1980:
This is the problem... its not specific to any one model, in theory applies to f/w prior to 449 (but even that's not consistent afaik)

dewet22:
Understood. Thanks for the contribution!

PR #16, opened 2022-12-23
Update battery.py

holdestmade:
I have 2 batteries but the list of batteries returns 3, one with a serial of `'          '` (10 spaces)

First battery is correct with a serial number, second returns` '          ' ` (10 spaces) and third has a serial number.

This should hopefully discount that incorrect battery.


dewet22:
Oh, interesting edge case. Thanks for reporting!

PR #42, opened 2023-12-29
Add FA inverter model type

oliver70112:
Added 'FA' to list of supported inverter types.

### DIFF
--- a/givenergy_modbus/model/inverter.py
+++ b/givenergy_modbus/model/inverter.py
@@ -27,6 +27,7 @@ class Model(str, Enum):
             'EA': cls.Gen2,
             'SA': cls.Hybrid,
             'SD': cls.Hybrid,
+            'FA': cls.Hybrid,
         }
         prefix = serial_number[:2]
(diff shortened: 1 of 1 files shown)


dewet22:
Hi @oliver70112 — apologies for the delay in responding to this, it's sat open far too long!

Unfortunately this PR can no longer be merged as-is. The serial-prefix lookup table it targets (`from_serial_number`) was removed in v1.0.0, which refactored model detection to use the device type code register rather than the serial number prefix.

If you're still using an FA-prefix inverter, it would be great to get it properly supported. The quickest way to help would be to report what value `HR(0)` (the device type register) returns on your unit — you can get this by running the library against your inverter and checking `plant.inverter.model`. If the first digit of that code isn't already in the `Model` enum, we can add it; if it is, it may already be working without any changes needed.

Thanks for the contribution regardless!

dewet22:
I'm 99% sure the changes integrated from GivTCP allows this inverter to be detected correctly now. If not, please do shout and I'd be happy to take a look.

PR #133, opened 2026-05-30
fix(model): recognise EMS serial format in is_valid_serial

### LINKED ISSUE #131: Serial validator logs a WARNING on every EMS-format serial
On a live EMS, `givenergy_modbus.model.register` logs a WARNING for the controller's serial on every poll:

```
serial number 'EMS2522018' is valid but does not match expected pattern AA0000A000
```

381 occurrences in a single session ([dewet22/givenergy-hass#52](https://github.com/dewet22/givenergy-hass/issues/52)). EMS controllers use the `EMSYYWWNNN` form, which legitimately doesn't match the standard `AAYYWWANNN` inverter serial pattern — so the warning is noise, not a real problem (the message even says "is valid but…").

### Fix
Teach the serial validator to accept the `EMS`-prefixed format without warning (or drop this particular notice to DEBUG). Pure log-noise reduction; no functional impact.

dewet22:
EMS controllers report their serial as a 3-letter prefix + 7 digits (e.g. `EMS2522018`), with no middle letter, so it does not match the inverter `AA0000A000` shape. `is_valid_serial()` therefore logged a WARNING on every poll: 381 occurrences in one EMS session (dewet22/givenergy-hass#52).

This adds an EMS serial pattern and only warns when a serial matches neither known form, so genuinely-unknown variants still surface but a valid EMS serial stops crying wolf.

Closes #131. Backport to `v2.1` to follow.

### DIFF
--- a/givenergy_modbus/model/register.py
+++ b/givenergy_modbus/model/register.py
@@ -172,18 +172,22 @@ class RegisterDefinition:
 
 
 _SERIAL_PATTERN = re.compile(r"[A-Z]{2}\d{4}[A-Z]\d{3}")
+# EMS controllers use a 3-letter prefix followed by 7 digits (e.g. EMS2522018) and carry
+# no middle letter — a legitimate format that just isn't the inverter AA0000A000 shape.
+_EMS_SERIAL_PATTERN = re.compile(r"[A-Z]{3}\d{7}")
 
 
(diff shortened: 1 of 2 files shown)


dewet22-codex:
I’ve reviewed this and have no concerns to add. I also ran the focused register tests plus the full tox gate locally.

Issue #95, opened 2026-05-28
EMS plant decoding follow-up: meter validity filter, IR(2040,55) verification, inverter cross-check


dewet22:
Follow-up to #86 / #93, opened after walking the wire capture from a real EMS install (see #86 for the captures and [dewet22/givenergy-hass#52](https://github.com/dewet22/givenergy-hass/issues/52) for the topology).

## Context

#93 stops the library issuing inverter-style reads against EMS plant controllers. The wire data also clarified something I'd had wrong: an EMS plant **doesn't** require sub-address inverter walking. Per-inverter telemetry is rolled up into the EMS's own `IR(2040..2094)` block (`inverter_N_power`, `inverter_N_soc`, `inverter_N_temp`, `inverter_N_serial_number`, plus per-meter and plant aggregates). Our `Ems` model already decodes all of it. So the remaining work after #93 is targeted, not architectural.

### 1. Filter empty meter slots in `detect()`

`Client.detect()`'s meter-probing loop appends any address in `0x01..0x08` that *responds* to `IR(60,30)` — but some EMS firmwares acknowledge unused slots with all-zero cache contents (see #86 capture: `0x07`, `0x08` returned zeros while `0x01`, `0x03` returned real voltage/current/power). Those zero-only slots become empty `Meter` objects in `plant.meters` for no benefit.

Suggested: after a probe response, gate the `caps.meter_addresses.append(...)` on `Meter.from_register_cache(cache).is_valid()` (which already checks `v_phase_1 ≠ 0`). Same model the library uses everywhere else; cheap and localised.

### 2. Wire-confirm the EMS `IR(2040,55)` read

#93 leaves the EMS-conditional `IR(2040,55)` append in place but documents it as unconfirmed — the existing capture ran with `capabilities=null`, so the read never fired. A 30-minute passive capture from the same hardware (requested over on hass#52) will let us decode the live values from `Ems.inverter_N_*` / `meter_N_*` against ground truth from the user's plant description (two Giv-AC 3.0 inverters, grid + load CTs, etc.) and either tighten the read range or extend it.

Not blocking — the read works in code today, just hasn't been exercised against hardware yet.

### 3. Cross-check `EMS.inverter_count` / `inverter_N_serial_number` against detect()

When the EMS reports `inverter_count = 2` with their serial numbers as 10-character strings, we could cross-reference those at detect time to:

- Validate the EMS register block parsed sensibly (count and serial-string charsets are easy sanity checks)
[...]

PR #137, opened 2026-05-30
test(fixtures): golden-master classification/topology checks over captures

### LINKED ISSUE #127: Re-verify EMS/AIO detect() classification after the 0x11→0x32 rewrite removal (#121)
Follow-up from #119 / #121.

Before #121, `Plant.update()` rewrote every response arriving at `0x11`/`0x00` to be stored under `0x32`. That masked the fact that EMS and All-in-One controllers actually answer at `0x11`, not `0x32` — and it meant the earlier EMS verification (#95) passed for the wrong reason: the rewrite relabelled the `0x11` capture data as `0x32`, so a replay-based check looked green without ever exercising the real `[0x11]` path. A replay-verification trap, in other words.

#121 removed the rewrite and pointed `detect()` at `register_caches[0x11]` for the DTC. So that green is worth re-confirming honestly now the real path is exercised:

- [ ] `detect()` reads the DTC from `[0x11]` and resolves EMS (`DTC 0x5001` → `Model.EMS`) and All-in-One (`DTC 0x8000` → `Model.ALL_IN_ONE`) correctly.
- [ ] `inverter_address` resolves to `0x11` for those models (not the old `0x32`).
- [ ] the EMS rollup cross-check (`_validate_ems_rollup` / `_ems_rollup_cross_check`) keys off `[0x11]`.
- [ ] classification holds when replayed against the committed fixtures `tests/fixtures/captures/ems_2_inv_3_bat_a/` and `aio_a/`.

Low effort — mostly a fixture-replay assertion pass against the captures added in #120; the classification itself doesn't need live hardware. Noted so it isn't lost now the addressing redesign has landed.

dewet22:
Closes #127 — the golden-master regression layer that was the point of committing real wire captures all along: replay each fixture through the *full* decode machinery and pin the decoded topology, so behaviour can't silently drift.

## Why this matters

`test_addressing_from_captures.py` (landed with #121) asserts which register *address* data lands at. This goes further — it runs the **classification + topology** path: framer → `Plant.update()` (true addressing, no rewrite) → `resolve_model()` → `PlantCapabilities` derivation → typed accessors (`ems`, `inverters`, `hv_stacks`).

That's the exact path the pre-#121 `0x11→0x32` rewrite masked: it relabelled `0x11` capture data as `0x32`, so the earlier EMS check (#95) passed *for the wrong reason* without ever exercising the real path. Pinning classification against real hardware data closes that trap — any future drift in the model/decode machinery now trips here.

## What's asserted per fixture

- **ems_2_inv_3_bat_a** → `Model.EMS` @ `0x11`; rollup decodes `inverter_count=2` and two `ems_rollup`-sourced managed inverters.
- **aio_a** → `Model.ALL_IN_ONE` @ `0x11` (HV); BCU stack separately addressed at `0x70`/`0xA0`/`0x50-0x53`.
- **hybrid_2_bat_a** → `Model.HYBRID_GEN1` @ `0x31` (operational banks at `0x31`, identity-only at `0x11`); LV batteries at `0x32`+.

Verified the assertions actually bite (mutation-checked: flipping an expected value fails the test). Full tox green.

### DIFF
--- a/tests/fixtures/captures/README.md
+++ b/tests/fixtures/captures/README.md
@@ -195,3 +195,13 @@ plant is self-documenting rather than growing a section here:
    and any anomalies (decoder errors, unusual frames). Don't include
    contributor-identifying information. Extend the battery-model table
    above if a new battery family appears.
+5. **Add golden-master coverage.** A new topology scenario must also gain
+   at least one test case in `tests/model/test_fixture_golden_master.py`
+   that replays it through the decode machinery and pins the expected
+   classification + topology (model, inverter address, EMS rollup, HV
+   stack, batteries — whatever the plant exercises). This is the whole
+   point of committing real captures: they're a standing regression
+   net, so a fixture without a golden-master assertion is dead weight
+   that can't catch drift. (Register-landing assertions in
+   `test_addressing_from_captures.py` are complementary, not a
+   substitute.)
--- a/tests/model/test_fixture_golden_master.py
+++ b/tests/model/test_fixture_golden_master.py
@@ -1,0 +1,142 @@
+"""Golden-master regression tests over the committed wire-capture fixtures.
+
+These replay each real-hardware capture in ``tests/fixtures/captures/`` through
+the *actual* decode machinery — framer → ``Plant.update()`` (true device
+addressing, no rewrite) → ``resolve_model()`` → ``PlantCapabilities`` derivation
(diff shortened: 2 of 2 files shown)


dewet22-codex:
I’ve reviewed this and have no concerns to add. I also ran the focused golden-master fixture tests plus the full tox gate locally.

PR #142, opened 2026-05-30
fix(client): raise refresh default budget to 2.0s/1 retry for contended buses

### LINKED ISSUE #132: refresh_plant defaults (1.0s / 0 retries) are fragile under multi-client contention
Observed during live probing on a HYBRID_GEN1 and on the EMS install ([dewet22/givenergy-hass#52](https://github.com/dewet22/givenergy-hass/issues/52)): when other clients (GivTCP, the vendor app, Predbat) poll the same inverter, the stock `refresh` / `refresh_plant` defaults (`timeout=1.0s`, `retries=0`) frequently lose the race and time out — e.g.:

```
Timeout awaiting 2:4/ReadInputRegistersResponse(device_address=0x31 base_register=0) after 2 tries at 1.0s, giving up
```

The inverter serialises requests, so a single shared bus plus an aggressive 1s/no-retry budget produces spurious "register read failed" / "treating plant as unreachable" cycles even though the device is responsive — a patient 3s + retry read succeeds reliably (confirmed via `tests/debug/probe_read_address.py`: 4/4 reads at every address with 3s + 1 retry).

### Not a correctness bug
The coordinator's partial-failure path serves last-known data, so nothing breaks — but the library defaults are tuned for an idle bus. Consider raising the default `refresh` timeout/retries, or documenting recommended values for contended setups (GivTCP/Predbat/app alongside). Worth aligning the library default with what downstreams already pass in places.

dewet22:
Raises the `refresh()` / `refresh_plant()` / `watch_plant()` default request budget from `timeout=1.0s, retries=0` to **`timeout=2.0, retries=1`**, fixing the contended-bus fragility in #132.

## Why

The inverter serialises Modbus requests, so when other clients (GivTCP, the vendor app, Predbat) poll the same unit, the old tight budget on the *hot poll path* lost the race and emitted spurious `register read failed` / `treating plant as unreachable` timeouts even though the device was responsive. Seen live on a HYBRID_GEN1 and on Nick's EMS (givenergy-hass#52); the #132 probe confirmed a 3s + 1-retry read succeeds 4/4 at every address under the same contention.

Odd asymmetry this corrects: `load_config` (config banks, run occasionally) was already patient at `2.0s/3`, while `refresh` (run every poll cycle) was the aggressive one. The hot path had it backwards.

## Change

- `refresh` / `refresh_plant` / `watch_plant` defaults → `timeout=2.0, retries=1` (kept below `load_config`'s 3 retries — refresh runs far more often, so a runaway retry budget there costs more poll latency).
- `load_config` (2.0s/3) and the low-level `send_request_and_await_response` (no default — explicit is correct) unchanged.
- Test pinning the new default budget; `usage.md` tuning section updated with the values + contention rationale.

## Trade-off

A genuinely unreachable inverter now surfaces as failed a bit slower (~4s worst case vs ~1s). The hass coordinator's timeout-tolerance already absorbs transient blips, so this mainly stops crying wolf. Consumers wanting fail-fast can still pass a tighter budget.

Closes #132.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -127,10 +127,17 @@ is not yet confirmed.
 
 `refresh`, `load_config` and `one_shot_command` all accept the same three knobs:
 
-- `timeout` (default 1.0s for refresh, 1.5s for `one_shot_command`) — how long to wait for
-  each response.
-- `retries` (default 0) — number of additional attempts after a timeout.
+- `timeout` — how long to wait for each response. Defaults: `refresh` 2.0s, `load_config`
+  2.0s, `one_shot_command` 1.5s.
+- `retries` — number of additional attempts after a timeout. Defaults: `refresh` 1,
+  `load_config` 3, `one_shot_command` 0.
 - `retry_delay` (default 0.5s) — seconds to wait between a timed-out attempt and the next.
+
+`refresh` defaults to `timeout=2.0, retries=1` because the inverter serialises requests:
+when other clients (GivTCP, the vendor app, Predbat) poll the same unit, a tighter budget
+loses the race and produces spurious "register read failed" timeouts even though the
+device is responsive (#132). If you own the bus exclusively and want genuine failures
+surfaced faster, pass a tighter `timeout`/`retries`.
 
 The retry delay exists because some inverters exhibit multi-second silent windows where
 they stop responding to anything; firing the retry immediately tends to land it inside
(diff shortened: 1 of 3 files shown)


dewet22-codex:
I’ve reviewed this and have no concerns to add. I also ran the focused refresh/load tests plus the full tox gate locally.

Issue #144, opened 2026-05-30
Identify registers for self-heating / winter conditioning / EPS-enable / export-priority (need portal write-captures)


dewet22:
## Context

`givenergy-hass` wants to surface several per-inverter control knobs that GE's own portal exposes (so they're intended-safe to write), but which have **no known Modbus register** — they're not in this library's map, and crucially **not in GivTCP either** (the de-facto reference now GE's docs are gone). GE's portal writes them via its cloud API, so the register can't be inferred from desk research. From the EMS/inverter control-parity handoff (givenergy-hass#52).

These four are the "needs hardware reverse-engineering" category — writing an unknown/guessed register to a live inverter is the genuinely risky case, so we should **not** guess. Each needs a wire capture of the GE portal *writing* the setting to identify the register safely.

## Knobs needing register identification

| Setting | Portal location | Observed | Suggested setter (once register known) |
|---|---|---|---|
| Enable Battery Self-Heating | Inverter → Battery → General | Off | `set_enable_battery_self_heating(bool)` |
| Winter Battery Conditioning | Inverter → Battery → Control | Off | `set_winter_battery_conditioning(bool)` |
| Enable EPS | Inverter → EPS | On | `set_enable_eps(bool)` (likely hybrid-gated) |
| Export Power Priority | Inverter → General | "Load First" (dropdown — enum) | `set_export_power_priority(enum)` |

## How to identify the registers safely

For each: capture the Modbus traffic while the user toggles the setting in the GE portal (or app), then read off the `WriteHoldingRegister` register + value. `tests/debug/replay.py` + a `givenergy-cli capture` session is the tooling; the capture shows the exact register write, which we then add to `WRITE_SAFE_REGISTERS` + a setter + a decoded read-back field, with a register-safety review.

Until a capture pins the register, these stay blocked — no guessing register addresses against live hardware.

## Notes
- **Read-back:** each also needs a decoded field on the `Inverter` model so the hass entity can show current state. Flag any that turn out write-only on the wire.
- **Model gating:** EPS is likely hybrid-only; self-heating may be outdoor-pack-only. Confirm from the captures / model so hass gates entity creation.
- **Battery cutoff (§7 of the handoff):** *not* included here — it most likely maps to the existing `set_battery_power_reserve` (GivTCP's `setBatteryCutoff` routes there). To be confirmed on the hass side against the portal semantics, no new register expected.
[...]

PR #145, opened 2026-05-30
feat(ems): add plant_enabled boolean read-back for Flexi EMS control

dewet22:
Adds the boolean read-back the hass EMS-parity work needs for a "Flexi EMS Control" switch (givenergy-hass#52).

## Gap

`set_ems_plant(bool)` writes HR(2040) (the EMS master-enable register), but the model only exposed `plant_status`, which decodes the **same** register as a coarse `Status` enum (WAITING/NORMAL/...). A HA switch needs a plain on/off boolean to reflect what the setter last wrote — `plant_status` can't cleanly back that.

## Change

Adds `Ems.plant_enabled: bool | None` as a `computed_field` — the boolean view of HR(2040) (non-zero = enabled, `None` if unread). `plant_status` is unchanged (kept as the enum view of the same register).

- **Polling:** no change — HR(2040) is already read for EMS (`load_config` reads HR(2040,36)).
- **Read/write pairing:** `plant_enabled` reads what `set_ems_plant` writes, so the switch state and setter agree.
- Verified against the EMS fixture (running plant → `plant_enabled=True`); empty cache → `None` (preserves the "all-None when empty" model_dump contract). Tests + tox green.

Note on a residual ambiguity (documented in the field docstring): HR(2040) is written as a 0/1 bool by `set_ems_plant` but read as a `Status` enum by `plant_status`. We've no capture of a *disabled* EMS to confirm whether the register is strictly an enable flag or an operational status — but since `plant_enabled` exists precisely to mirror the setter, `bool(HR2040)` is the right backing for the switch either way.

Part of the EMS-parity handoff; the register-unknown items stay in #144.

### DIFF
--- a/givenergy_modbus/model/ems.py
+++ b/givenergy_modbus/model/ems.py
@@ -2,7 +2,7 @@
 
 from typing import ClassVar
 
-from pydantic import ConfigDict, create_model
+from pydantic import ConfigDict, computed_field, create_model
 
 from givenergy_modbus.model.devices import InverterSummary
 from givenergy_modbus.model.inverter import Status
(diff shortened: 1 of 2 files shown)


dewet22-codex:
I’ve reviewed this and have no concerns to add. I also ran the focused EMS model tests plus the full tox gate locally.

dewet22-codex (approved):
I’ve reviewed this and have no concerns to add. I also ran the focused EMS model tests plus the full tox gate locally.

Issue #75, opened 2026-05-16
Move commands onto Inverter so model-specific contracts are enforceable


dewet22:
## Problem

The `givenergy_modbus.client.commands` module is currently a stateless utility — every function builds a `WriteHoldingRegisterRequest` (or list) from arguments, with no knowledge of the inverter it'll target. This was a reasonable simplification when v1 supported a single inverter family, but v2's wider model coverage makes the abstraction leak in three concrete ways already, and the pattern will keep recurring:

### Leak 1: slot_map is inverter-specific

The slot setters (`set_charge_slot`, `set_discharge_slot`, etc.) need `inverter.slot_map` to route the call to the correct register pair. Currently the caller has to thread it through:

```python
commands.set_charge_slot(idx, ts, plant.inverter.slot_map)
```

The caller shouldn't need to know `slot_map` exists. (Addressed partially in d3e4938 by removing the default, which forces correctness but doesn't fix the abstraction leak.)

### Leak 2: many commands are model-specific

Several v2 commands only apply to certain inverters but the library doesn't enforce it:

| Command | Only applies to |
|---|---|
| `set_ac_charge` / `set_force_charge` / `set_force_discharge` | three-phase |
| `set_battery_charge_limit_ac` / `_discharge_limit_ac` | three-phase |
| `set_ems_plant` / `set_export_slot*` | EMS |
| `set_pause_slot*` | models with `BATTERY_PAUSE_*` registers |
| slot setters | model-dependent register layout |

Today a Gen 1 single-phase user can call `commands.set_ac_charge(True)`. The call builds, the wire send happens, the safety check waves it through (1112 is in the global allowlist), the inverter silently ignores. Same footgun shape as the slot_map issue, just slower to manifest.

### Leak 3: WRITE_SAFE_REGISTERS is also model-specific

The global set in `givenergy_modbus/pdu/write_registers.py` lists every register *any* inverter can safely write to — not what *this* inverter can write to. So the safety check passes for any register on any inverter's list, even when the caller's inverter doesn't have that register.

## Proposed shape

Move commands onto the inverter as instance methods, organised by mixin so model-specific commands only exist on the inverter types that support them. The `commands.*` module stays as the underlying primitive layer (useful for tests, internal plumbing) but no longer the recommended call site.

```python
class _InverterCommands:
    """Commands every inverter supports."""
    WRITE_SAFE_REGISTERS: frozenset[int] = frozenset({20, 27, 29, ...})

[...]

PR #152, opened 2026-05-31
feat(model): add is_ac_coupled topology discriminator

dewet22:
Adds an `is_ac_coupled` topology discriminator so downstream consumers can gate AC-coupled-only controls without hardcoding the model set. Requested by `givenergy-hass`, which wants to expose the AC charge/discharge limit controls only on AC-coupled inverters.

## What

- **`AC_COUPLED_MODELS`** (`frozenset({Model.AC, Model.AC_3PH})`) — exported module-level constant in `inverter.py`. Complete as-is: both are coarse families with no specific sub-variants, so an AC DTC like `0x3001` resolves to `Model.AC` via `Model._missing_`.
- **`PlantCapabilities.is_ac_coupled`** — the surface hass gates on (`caps.is_ac_coupled and not caps.is_three_phase` scopes single-phase AC). Mirrors the existing `is_three_phase`.
- **`is_ac_coupled` computed field** on `SinglePhaseInverter` *and* `ThreePhaseInverter`. Duplicated onto the three-phase class rather than inherited — the two inverter classes are parallel (same pattern as `battery_max_power` / `inverter_max_power`), and this is what makes `AC_3PH` correctly read `True`.

## Register hygiene (the `#75` divergence, made explicit)

The `WRITE_SAFE_REGISTERS` comment previously called HR313/314 "ambiguous". Reworded to state the real reason: they're the *single-phase* AC limit registers, but `ThreePhaseInverter` remaps the read-backs to HR1110/1108, so read ≠ write on three-phase AC. A correct three-phase write needs per-model command-register selection (#75) — out of scope here; the consumer excludes three-phase AC via `not is_three_phase`. **No behavioural change** — the write-safe set is byte-identical.

Tests pin this both ways: single-phase read==write (the model read-back reads the same register the command writes), and the deliberate three-phase divergence, so the gap stays explicit rather than drifting silently.

## Tests

`is_ac_coupled` true/false matrix on `PlantCapabilities` and both inverter classes; single-phase register-consistency pin; three-phase divergence pin; updated `model_dump` snapshots. Tox green (py314 + format + mypy + bandit + build + docs).

## Not in scope (deferred)

Three-phase AC limit *writes* — tracked in #75. The consumer scopes to single-phase AC, so this isn't a blocker.

### DIFF
--- a/givenergy_modbus/client/commands.py
+++ b/givenergy_modbus/client/commands.py
@@ -652,8 +652,12 @@ class _InverterCommands:
         def slot_map(self) -> SlotMap: ...
 
     # Universally-applicable subset of pdu.write_registers.WRITE_SAFE_REGISTERS.
-    # Excludes 313/314 (BATTERY_*_LIMIT_AC — ambiguous), 318-320 (pause mode),
-    # 1112/1122/1123 (three-phase), and 2040/2062-2069 (EMS).
+    # Excludes 318-320 (pause mode), 1112/1122/1123 (three-phase), and
+    # 2040/2062-2069 (EMS). Also excludes 313/314 (BATTERY_*_LIMIT_AC): these are
+    # the *single-phase* AC charge/discharge limits, but ThreePhaseInverter remaps
+    # the read-backs to HR1110/1108, so read != write on three-phase AC. A correct
+    # three-phase write needs per-model command-register selection (#75); until that
+    # lands they stay out of the universal write-safe set.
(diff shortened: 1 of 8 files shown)


dewet22-codex:

dewet22-codex (approved):
I’ve reviewed this and have no concerns to add. I ran the focused topology/command tests, a small AC-coupled runtime matrix, ruff, and the full tox gate locally; GitHub checks are green as well.

PR #153, opened 2026-05-31
fix(model): clamp zero month/day in to_datetime to prevent crash

dewet22:
## Summary

- `RegisterCache.to_datetime` called `self.get(m, 1)` to default a missing month register to 1, but the fallback only fires when the key is **absent** — not when it's present with value 0.
- On real AC hardware (confirmed across 3 devices in hass#52: 2× AC inverter + 1× EMS), HR(41) is populated but reads 0, so `datetime.datetime(..., 0, ...)` raises `ValueError: month must be in 1..12, not 0` and crashes the coordinator on every refresh cycle.
- Fix: `self.get(m, 1) or 1` and `self.get(d, 1) or 1` — clamps both month and day to 1 when the register is present but zero, matching the intent of the existing `"safe defaults for zeroes"` docstring.

## Test plan

- [ ] New regression test `test_to_datetime_zero_month_and_day`: builds a `RegisterCache` with month=0 and day=0, asserts the call returns `datetime(2024, 1, 1, 12, 0, 0)` without raising.
- [ ] All 677 existing tests continue to pass.
- [ ] `tox` (py314 + format + lint + build) green locally.

### DIFF
--- a/givenergy_modbus/model/register_cache.py
+++ b/givenergy_modbus/model/register_cache.py
@@ -79,7 +79,7 @@ class RegisterCache(defaultdict[Register, int]):
 
     def to_datetime(self, y: Register, m: Register, d: Register, h: Register, min: Register, s: Register):
         """Combine 6 registers into a datetime, with safe defaults for zeroes."""
-        return datetime.datetime(self[y] + 2000, self.get(m, 1), self.get(d, 1), self[h], self[min], self[s])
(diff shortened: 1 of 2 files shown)


dewet22-codex:

dewet22-codex (approved):
I’ve reviewed this and have no concerns to add. I ran the focused register-cache tests, a direct zero/missing month-day datetime probe, ruff, and the full tox gate locally; GitHub checks are green as well.

Issue #105, opened 2026-05-28
AIO refresh times out — needs real-hardware wire capture to validate device-address (0x32 vs 0x11)


gaima8:
Hey,

For the record, "Failed to connect to the inverter." is from the home assistant plugin, but I've been trying to use modbus directly to connect to my Givenergy AIO now for ~2 days which led me to try home assistant for the first time along with givenergy-hass. But, as I get the same sort of behaviour from both, I figure the issue is more modbus related that home assistant related.

The example python doesn't work against my AIO.

```
root@74f48cfdf3aa:/sol# grep -v ^# modbus.py
import asyncio
from givenergy_modbus.client.client import Client
from givenergy_modbus.client import commands
from givenergy_modbus.model import TimeSlot
from givenergy_modbus.model.inverter import Model

async def main():
    client = Client(host="192.168.20.195", port=8899)
    await client.connect()

    # Read current state first (needed for slot_map)
    await client.refresh_plant(full_refresh=True)
    plant = client.plant


    await client.close()

asyncio.run(main())
root@74f48cfdf3aa:/sol# python3 modbus.py
Timeout awaiting 2:4/ReadInputRegistersResponse(device_address=0x32 base_register=0) after 1 tries at 1.0s, giving up
Traceback (most recent call last):
  File "/sol/modbus.py", line 47, in <module>
    asyncio.run(main())
    ~~~~~~~~~~~^^^^^^^^
  File "/usr/local/lib/python3.14/asyncio/runners.py", line 204, in run
    return runner.run(main)
           ~~~~~~~~~~^^^^^^
  File "/usr/local/lib/python3.14/asyncio/runners.py", line 127, in run
    return self._loop.run_until_complete(task)
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^
  File "/usr/local/lib/python3.14/asyncio/base_events.py", line 719, in run_until_complete
    return future.result()
           ~~~~~~~~~~~~~^^
  File "/sol/modbus.py", line 12, in main
    await client.refresh_plant(full_refresh=True)
  File "/usr/local/lib/python3.14/site-packages/givenergy_modbus/client/client.py", line 471, in refresh_plant
    await self.execute(reqs, timeout=timeout, retries=retries, retry_delay=retry_delay)
  File "/usr/local/lib/python3.14/site-packages/givenergy_modbus/client/client.py", line 710, in send_request_and_await_response
    raise TimeoutError()
TimeoutError
[...]

PR #159, opened 2026-06-01
fix(model): treat residential All-in-One as single-phase (#105)

dewet22:
## Summary

The residential All-in-One (DTC family "8", e.g. `0x8001`) was classified as three-phase. Two consequences, both broken:

1. **Polling** — `refresh()`/`load_config()` requested the per-phase `IR/HR(1000+)` bank the AIO doesn't expose. Every such read error-responds, so `refresh()` failed with `RefreshPartiallySucceeded` — the failure gaima8 hit in #105 once the b5 request-CRC fix got the unit talking.
2. **Decoding** — it was decoded as `ThreePhaseInverter`, whose layout shadows ~30 live fields (battery SoC, AC voltage/frequency, firmware, charge slots, status) onto those same absent 1000-range addresses, so they'd read `None` even if polling were fixed.

## Fix

Remove `Model.ALL_IN_ONE` from both three-phase sets:
- `_THREE_PHASE_MODELS` (`model/plant.py`) — the poll gate behind `is_three_phase`.
- `THREE_PHASE_MODELS` (`model/inverter_threephase.py`) — the decode-layout gate behind `select_inverter`.

The AIO now polls and decodes single-phase, reading its real data from the `IR(0)`/`IR(180)` banks it answers. Its HV battery voltage (307 V), extended 10-slot map and `is_hv` are keyed on the model at the capabilities layer, so they're unaffected by the inverter decode class.

## Evidence

Diagnosed against the reporter's real AIO (#105):
- `HR(0) = 0x8001` → `Model.ALL_IN_ONE` (residential family "8").
- GE spec sheet: 3.6 kW / 16 A continuous, 7.2 kW / 32 A peak — single-phase domestic ratings; owner confirmed 1-phase.
- Decoded the unit's actual register dump both ways: single-phase yields complete, sane values (SoC 4%, 246.1 V, 49.98 Hz, firmware D0.612-A0.612); three-phase yields ~30 nulls.

## Scope

Conservative — only `ALL_IN_ONE` is reclassified. `AIO_COMMERCIAL` stays three-phase. `HYBRID_HV_GEN3` and `ALL_IN_ONE_HYBRID` are suspected single-phase HV, but I've left them untouched pending hardware confirmation rather than guess.

## Tests

- `is_three_phase` true/false sets updated; a guard that the AIO keeps `is_hv` / extended slots while not being three-phase.
- `select_inverter(ALL_IN_ONE)` → `SinglePhaseInverter`.
- `load_config` / `refresh` request-list regressions asserting no `base >= 1000` reads for the AIO.

## One thing I'll close out with the reporter

The failing reads in the capture are all on non-60-aligned bases (1000, 1060…), so I've asked them to also probe the correctly page-aligned 1000-range (960, 1020…) — to be fully sure the bank is genuinely absent rather than just rejected for misalignment. The model-level evidence (single-phase spec + owner confirmation) makes that a formality, but it removes the last doubt, and it would flag a latent alignment issue on the same code path (the gateway poll shares the offset) if anything does come back.

Refs #105.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected ALL_IN_ONE inverter classification to single-phase architecture, ensuring proper register polling and data decoding for this device model.

* **Tests**
  * Added test coverage to verify ALL_IN_ONE inverter decodes correctly as single-phase and prevents three-phase register polling.

### DIFF
--- a/givenergy_modbus/model/inverter_threephase.py
+++ b/givenergy_modbus/model/inverter_threephase.py
@@ -523,12 +523,18 @@ class ThreePhaseInverter(  # type: ignore[valid-type,misc]
         return self.work_time_total_hours  # type: ignore[attr-defined,no-any-return]
 
 
+# Models that decode via the three-phase / 1000-range register layout. The residential
+# ALL_IN_ONE (DTC family "8", e.g. 0x8001) is deliberately absent: it is HV but single-
+# phase, and decoding it as ThreePhaseInverter shadows ~30 live fields (battery_soc, v_ac1,
+# f_ac1, firmware, charge slots, status…) to IR/HR(1000+) addresses it doesn't expose,
+# zeroing them. It carries its data in the single-phase IR(0)/IR(180) banks and so decodes
+# correctly as SinglePhaseInverter. Verified against real AIO register dumps (#105). HV
+# battery voltage, extended slots and is_hv stay model-keyed at the capabilities layer.
 THREE_PHASE_MODELS: frozenset[Model] = frozenset(
     {
         Model.HYBRID_3PH,
         Model.AC_3PH,
         Model.AIO_COMMERCIAL,
-        Model.ALL_IN_ONE,
         Model.HYBRID_HV_GEN3,
         Model.ALL_IN_ONE_HYBRID,
     }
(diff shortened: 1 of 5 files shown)


dewet22:
@coderabbitai review

dewet22-codex:

dewet22-codex (approved):
I’ve reviewed this and have no concerns to add. I checked the AIO classification split through the request-generation, inverter-selection, capability, and golden replay tests, then ran the full tox gate locally. The only remote status still pending when I checked was CodeRabbit.

Issue #158, opened 2026-06-01
Response-CRC algorithm + redaction CRC integrity + faithful replay mock


dewet22:
Spun out of the #105 investigation. Two related, pre-existing defects in CRC/redaction handling, plus a tooling payoff. Sequenced **B-3 → B-2** (the response-CRC algorithm first, since the rest builds on it).

## Background

#105's request-CRC fix (#157) confirmed our request frames had the CRC wrong. While diagnosing, two response-side / tooling gaps surfaced:

1. **We never compute or validate response CRCs.** `TransparentResponse._update_check_code` is a stub that writes back whatever was decoded; `framer.py` notes "it is unclear how a response CRC is calculated or should be verified." So we can't detect corrupted frames *from* the inverter, and we can't re-encode a response with a correct CRC.

2. **`client.redact()` invalidates frame CRCs.** It zeroes serial digits length-preservingly but does **not** recompute the CRC (its docstring wrongly assumes "same length ⇒ CRC stays consistent"). The serial is inside the CRC-covered payload, so every redacted capture frame is self-inconsistent: body says `CH0000G000`, CRC still signs the original. All 6 committed `.log` fixtures are CRC-invalid as a result. Harmless today only because we don't validate response CRCs (defect 1) — but it means the fixtures can't be replayed to a real client that *does* validate.

### B-3 — reverse-engineer the GE response-CRC algorithm (first)
- The request side is now solved (`crc(device_address + function_code + fields)`, CRC16/Modbus, byte-swapped on the wire — see #157 and the FC 0x16 `ReadMeterProductRegistersRequest`). That's a working reference for the algorithm/poly/byte-order.
- Brute-force the **response** CRC span the same way. We hold the **unredacted** serials of the fixture devices, so genuine response frames can be reconstructed and solved against (candidate spans: `device + fn + …register_values`, with/without the serial; both byte orders).
- Implement it in the response `_update_check_code` (currently FIXME stubs in `read_registers.py` / `transparent.py` / `null.py`), enabling response-CRC **computation and validation**.
- Then **fix `redact()` to recompute the CRC** after mutating serial/IPv4 bytes, and re-redact the 6 committed `.log` fixtures so they're internally consistent. Switch the redaction serial scheme from digit-zeroing (which yields *impossible* serials — no valid YYWW) to **notionally-valid** substitutes (e.g. `CH2204G123`) so redacted frames look real to a validating client.

### B-2 — faithful replay mock as a reusable test platform (then)
[...]

Issue #100, opened 2026-05-28
Inverter dongle broadcasts LAN config in plaintext every 3 minutes via undocumented WO-prefix frames


dewet22:
While investigating an EMS-plant install (context: #86, #95), passive 30-minute captures from one of the inverter dongles surfaced a class of frame the library doesn't currently decode and which carries an unencrypted network-config payload.

## Observation

Every ~3 minutes the inverter dongle emits a 70-byte frame on the modbus channel that:

- Does **not** use the transparent-encapsulation marker (`0x8a`) and therefore doesn't carry a modbus function code.
- Uses a source-serial prefix of **`WO`** — a family we haven't seen before (we know `FO`/data adapter, `AB`/response-source, `CE`/inverter, `BE`/`BJ`/`AC`/battery families).
- Contains an ASCII payload of the shape `,<ipv4>,<netmask>,<gateway>\r\n\r\n` — the dongle's own LAN configuration as a CSV.

Decoded layout:

| Bytes | Meaning |
|---|---|
| `5959 0001 0040 01 02` | Header (length = 64 bytes) |
| `574f3030303047303030` | Source serial `WO0000G000` (after CLI redaction) |
| `00 × 8` | Padding |
| `00 × 6` | Leading zeros in payload |
| ASCII CSV | `,<ip>,<netmask>,<gateway>` |
| `\r\n\r\n` | CRLF terminator |
| 2 bytes | CRC |

In the analysed capture, the IP is the inverter's LAN address, the netmask is a `/22`, and the gateway is the home router.

## Prior art and related context

The `WE`/`WF`/`WO` letter family is documented in a different protocol surface — the dongle's WiFi SSID. mundayweb's [GivEnergy security page](https://www.mundayweb.com/html/GivEnergy/Security.html) covers that the dongle broadcasts its own SSID as one of those prefixes followed by the dongle's serial number — a separately documented serial-number disclosure of its own. Tangentially related but distinct: the observation here is that the same letter family also appears as a **source identifier on modbus-channel frames over port 8899**, carrying a network-config CSV which mundayweb's coverage doesn't address.

Cross-checked against captures from a second unrelated installation: the modbus-channel source identifier matches the WiFi SSID format and family letters. Same naming pool, used as a stable dongle-identity tag across both protocol surfaces — modbus frame source and wifi SSID broadcast.

## Privacy / security impact

[...]

PR #160, opened 2026-06-01
fix(pdu/client): response CRC integrity + frame-aware capture redactor (#158)

dewet22:
## Summary

Three pre-existing defects addressed together because they're mechanically linked — the CRC bug hid the redaction bug:

1. **Response CRCs were never computed** — `_update_check_code` on the response side was a no-op stub that echoed back the decoded value. The whole protocol shares one CRC scheme (`CRC16/Modbus over payload[18:]`, byte-swapped), now expressed as a single method on `TransparentMessage`. Confirmed against the real All-in-One corpus (102/102 frames valid, including error responses).

2. **`redact()` invalidated CRCs** — zeroing serial digits changed the bytes inside the CRC-covered span without recomputing the check field. Harmless only because of defect 1. Replaced with a frame-aware `FrameRedactor` that decodes each frame, zeroes exactly the fields whose type (`C.serial`, `LanConfigBroadcast`) marks them sensitive, and re-encodes with a fresh CRC. Undecodable frames and inter-frame garbage are emitted intact with a log message — nothing is dropped.

3. **`LanConfigBroadcast` (#100)** — the WO-prefix dongle LAN-config broadcast (`fid=0x02`, `,ip,netmask,gateway\r\n\r\n` body) was falling through to `InvalidFrame`, leaking the LAN IP every ~3 min in captures. Now decoded as a first-class PDU by a two-byte discriminator; IP fields self-redact with digit-zeroing.

4. **Fixture CRCs and serial bytes** — all committed fixtures now carry self-consistent CRCs. The known `YYWW` manufacture dates were backported into serial fields where previously zeroed by the old all-zeros redactor (e.g. `CE0000G000` → `CE2231G000`), matching what the current `FrameRedactor` would produce. `_redact_extra.py` retired (cross-frame EMS-rollup case is now handled by `FrameRedactor`'s `_SERIAL_GROUPS` lookup).

## Test plan

- [ ] All 678 tests pass; tox fully green (py314, format, lint, build/docs)
- [ ] `test_frame_redactor_*` — envelope + payload serial redaction; invalid frame passthrough (intact + logged); LanConfigBroadcast round-trip
- [ ] `test_redact_serial_*` — `Converter.redact_serial` rule coverage in `test_register.py`
- [ ] Golden-master fixture tests pass with updated serial bytes and CRCs
- [ ] Register-safety review: PASS — CRC change is check-bytes-only, `WRITE_SAFE_REGISTERS` additive-only, no new bare write paths
- [ ] `@coderabbitai review` (non-default branch, triggering manually)
- [ ] Codex review gate


dewet22:
@coderabbitai review

dewet22-codex:

givenergy_modbus/client/client.py
```diff
@@ -41,136 +41,175 @@
...
         return out
 
+    def _redact_frame(self, frame: bytes) -> bytes:
+        from givenergy_modbus.model.register import Converter
+        from givenergy_modbus.pdu import ClientIncomingMessage
+        from givenergy_modbus.pdu.lan_config import LanConfigBroadcast
+        from givenergy_modbus.pdu.read_registers import ReadRegistersResponse
+
+        try:
+            pdu = ClientIncomingMessage.decode_bytes(frame)
```
dewet22-codex: This only decodes client-incoming frames, but capture_frames() feeds the TX path through the same FrameRedactor at client.py:1062. Outgoing requests are ClientOutgoingMessage frames, so they currently hit the exception path and are emitted byte-for-byte. I reproduced that with a ReadHoldingRegistersRequest carrying data_adapter_serial_number="WF1234G567"; the redactor logged it as undecodable and the captured TX bytes still contained WF1234G567. Since capture_frames() promises the sink never sees raw hardware identifiers, I think the redactor needs to either know the direction or try ClientOutgoingMessage before falling back to intact output.

givenergy_modbus/pdu/lan_config.py
```diff
@@ -0,0 +1,134 @@
...
+        redacted_ip = _zero_ip(self.ip)
+        redacted_netmask = _zero_ip(self.netmask)
+        redacted_gateway = _zero_ip(self.gateway)
+        new_csv_raw = f",{redacted_ip},{redacted_netmask},{redacted_gateway}\r\n\r\n".encode("latin1")
+        return LanConfigBroadcast(
+            data_adapter_serial_number=redacted_serial,
+            ip=redacted_ip,
+            netmask=redacted_netmask,
+            gateway=redacted_gateway,
+            check=self.check,
```
dewet22-codex: redact() rewrites _csv_raw with the zeroed IP/netmask/gateway but carries the original trailing check value through unchanged. The current test uses an already-zeroed fixture, so this path is idempotent there; with a real LAN broadcast such as 192.168.1.10, the emitted frame changes the CSV bytes while leaving the last two bytes untouched. If those bytes are meant to be a checksum/CRC, the redacted capture is no longer self-consistent, which undercuts the fixture-integrity goal of this PR. I would either recompute the LAN-config check here, or add an explicit test/comment proving that this field is not tied to the CSV payload.

dewet22-codex (commented):
I found two capture-redaction concerns inline. Local verification otherwise passed: focused capture/PDU/framer/register/golden tests, ruff check/format, and the full tox gate. Remote functional checks were green, with Codecov patch coverage still red when I checked.

givenergy_modbus/model/register.py
```diff
@@ -70,6 +70,39 @@ def string(*vals: int) -> str | None:
...
+        Identical wire form to `string` (concatenated big-endian ASCII), but a distinct
+        converter so the capture redactor can locate serial-bearing registers *by type*
+        — `Def`s tagged `C.serial` — rather than guessing from byte shape. The redaction
+        rule itself lives in `redact_serial`, so "this field is an identifier" and "here's
+        how it's redacted" are declared together on the type, in one place.
+        """
+        return Converter.string(*vals)
+
+    @staticmethod
+    def redact_serial(s: str | None) -> str | None:
```
dewet22, line 86: Not sure that this should sit on Converter – it feels like a utility much more broadly useful.

givenergy_modbus/client/client.py
```diff
@@ -41,136 +41,175 @@
...
+        if hasattr(pdu, "inverter_serial_number"):
+            pdu.inverter_serial_number = Converter.redact_serial(pdu.inverter_serial_number) or ""
+
+        # Redact payload serials in register responses.
+        # A serial is stored across 5 consecutive registers; decode the group as a
+        # string, apply redact_serial, and re-encode back into register values.
+        if isinstance(pdu, ReadRegistersResponse) and not pdu.error:
+            reg_type = "HR" if pdu.transparent_function_code == 3 else "IR"
+            win_base = pdu.base_register
+            win_end = win_base + pdu.register_count
```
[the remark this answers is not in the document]
dewet22: Fixed — using len(pdu.register_values) to determine the window end.

givenergy_modbus/client/client.py
```diff
@@ -41,136 +41,175 @@
...
+    for module in (inverter, battery, ems, gateway):
+        for attr in dir(module):
+            cls = getattr(module, attr)
+            if not isinstance(cls, type):
+                continue
+            lut = getattr(cls, "REGISTER_LUT", None)
+            if not lut:
+                continue
+            for _field, defn in lut.items():
+                if defn.pre_conv is Converter.serial and defn.registers:
```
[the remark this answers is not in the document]
dewet22: Fixed — unwrapping the tuple form before the identity check.

givenergy_modbus/pdu/lan_config.py
```diff
@@ -0,0 +1,134 @@
...
+
+from givenergy_modbus.codec import PayloadDecoder
+from givenergy_modbus.pdu.base import ClientIncomingMessage
+
+_DIGIT_RE = re.compile(rb"\d")
+
+
+def _zero_ip(s: str) -> str:
+    """Zero every digit in a dotted-quad IPv4 string, preserving dots and length."""
+    return _DIGIT_RE.sub(b"0", s.encode("latin1")).decode("latin1")
```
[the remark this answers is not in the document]
dewet22: Fixed — added the None/empty guard.

givenergy_modbus/pdu/lan_config.py
```diff
@@ -0,0 +1,134 @@
...
+    @classmethod
+    def lookup_main_function_decoder(cls, function_code: int) -> "type[ClientIncomingMessage]":
+        """Not used — LanConfigBroadcast is decoded directly, not via lookup."""
+        raise NotImplementedError()
+
+    def redact(self) -> "LanConfigBroadcast":
+        """Return a new instance with the adapter serial and all IP fields zeroed."""
+        from givenergy_modbus.model.register import Converter
+
+        redacted_serial = Converter.redact_serial(self.data_adapter_serial_number)
```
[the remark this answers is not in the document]
dewet22: Fixed — added or "" guard.

givenergy_modbus/client/client.py
dewet22: Good catch — fixed in c028f67. The TX FrameRedactor now uses ClientOutgoingMessage as the decoder, so request frames are decoded and their adapter serials redacted rather than falling through to intact-passthrough. Added a regression test.

givenergy_modbus/pdu/lan_config.py
dewet22: The check field does not follow the CRC16/Modbus(payload[18:], byte-swapped) scheme used by all other GivEnergy frames — verified by testing every candidate span. Its derivation is opaque (firmware-private to the WO-prefix dongle). The field is now preserved verbatim with a docstring and test comment explaining why, and an explicit assertion documenting the passthrough. On a live capture with real IPs the check would be left inconsistent with the rewritten CSV, but without understanding the formula it cannot be corrected.

givenergy_modbus/model/register.py
dewet22: The placement is intentional: C.serial is a distinct converter type that owns both decoding (delegates to string) and its own redaction rule, keeping the two semantics co-located on the type. This mirrors the principle that drove introducing C.serial in the first place — so the capture redactor can locate serial-bearing registers by type rather than by byte shape. A top-level function would sever that connection.

Issue #48, opened 2026-05-11
Field-validate register mappings inherited from britkat's fork


britkat1980:
Hi Dewet, great to see you back on the scene. In your absence I've been using a fork of your previous (unreleased) async client and been updating it to work with a huge range of new kit from Giv.
Now you are back in the saddle with the project, I'm happy for you to pick up some of the new models I created. Things like the HV batteries are completely different...

Check out the for here: https://github.com/britkat1980/givenergy-modbus-async

There's probably lots of stuff in there you won't want, but there's a full set of DTC-> Model mapping, Meters and new structures for the HVBMU etc...

Let me know if you want any help.

dewet22:
Hi @britkat1980 — thanks for keeping this going while I was AWOL, and for the offer. I spent a bit of time going through [givenergy-modbus-async](https://github.com/britkat1980/givenergy-modbus-async) and there's a huge amount of really valuable work in there, especially the device coverage. Let me lay out how I'd like to approach pulling things back upstream.

### Bits I'd like to bring across first

These all live in new files in your fork, so they slot in alongside what's already here without changing anything that already works:

- **HV battery support** — your `hvbcu.py` and `hvbmu.py` for the BCU control unit and BMU modules. This is the biggest gap upstream right now and a lot of people would benefit (AIO, HV Gen3, AIO Hybrid).
- **Meter support** — your `meter.py`, the new `MR` register namespace, and the function code 0x16 response (`ReadMeterProductRegistersResponse`). Upstream just has a `MeterType` enum that doesn't do much.
- **EMS** (`ems.py`) and **Gateway / Gateway2** (`gateway.py`) register tables.
- **Three-phase inverter** register table (`threephase.py`, HR/IR 1000-1420).
- Extra register-decoding helpers and enums (`MeterStatus`, `WorkMode`, `Phase`, `Generation`, etc.) and per-bit error decoding tables.
- New command helpers that don't overlap with existing ones: `set_battery_pause_mode`, pause/export slot setters, `set_enable_rtc`, `set_active_power_rate`, EMS plant enable.

### Bits where we need to chat first

Two of the bigger changes in your fork are good ideas but would break existing users of the library, so I'd like to design how to land them rather than copy them in directly:

[...]

Issue #106, opened 2026-05-28
Tracking: graph-shaped Plant abstraction for multi-device topologies


dewet22:
Tracking issue for the `Plant` model refactor sketched in a working plan document. The current model treats the world as "one inverter with peripherals hanging off it" — flat `dict[device_address → RegisterCache]`, single `PlantCapabilities`, 1:1 `Client` ↔ `Plant`. Real EMS installs (see #86 and the captures committed in #103) break this shape:

- EMS plant controller is a *peer* of the inverters it manages, not their parent
- Batteries belong to their parent inverter, not directly to the plant
- Some devices have their own communication path (separate dongles), some don't (the "blinded inverter" case)
- Direct-data and EMS-rollup views of the same inverter need reconciliation (by serial-number cross-reference)

The intended outcome is a graph-shaped `Plant` that handles all of these correctly while staying incremental — no big-bang v3.0, no breaking [dewet22/givenergy-hass#62](https://github.com/dewet22/givenergy-hass/pull/62)'s capability persistence on the integration side.

## Design intent

The `Plant` metaphor — graph of devices connected via transports, cross-device serial reconciliation — is a generic energy-topology abstraction. Not GE-specific, not Modbus-specific. The same shape applies to mixed-vendor installations (Daikin heat pump + Tesla Powerwall + myenergi Zappi + GivEnergy). Design discipline keeps it hoistable: when a second-manufacturer or second-protocol case appears, extracting becomes a code-org refactor rather than an architectural redesign. Detail in `project_plant_abstraction_direction.md` in project memory.

## Phased rollout

Plan splits the work into phases so each can be merged independently and the integration keeps working throughout. Detailed phase breakdown lives in the local working document; high-level shape:

1. **Target architecture sketch** — `Plant` → `Devices` (subclasses) → `Transport` (abstract), with a `serial_index` for cross-device reconciliation
2. **Unified `Inverter` facade** — direct / blinded / merged data sources behind one type (PR #98 was first stab at this)
3. **`Transport` abstraction** — extract the modbus-specific I/O behind an interface so HTTP / MQTT / etc. can slot in later
4. **Cache-ownership migration** — `RegisterCache` becomes a modbus-impl detail, not a generic-`Device` concern

## Scope

v2.1 line. Plan as currently written isn't committed to the repo — kept as a working document until the phases stabilise enough to track per-phase progress against this issue.

dewet22:
[...]

PR #167, opened 2026-06-02
feat(pdu): extend WRITE_SAFE_REGISTERS with app-confirmed registers

dewet22:
## Summary

The GivEnergy Android app's **Direct Control → Control tab** now exposes the complete user-writable register surface (HR 20–HR 5014), with register number + human name + value range for each entry. We catalogued the full list by running the app against `MockPlant` (no live hardware needed).

The app listing a register as user-editable is authoritative evidence it accepts writes — a far stronger signal than the guesswork the list was originally built on.

**42 registers added.** Cross-checked against GivTCP's `safe_regs` list for completeness; deliberately extends beyond GivTCP where the app confirms registers GivTCP didn't know about.

### What's added

| Group | Registers | Notes |
|---|---|---|
| Battery heaters | 104, 172 | Hardware/batch-gated — write may be rejected per-unit (HR104 rejected on one AC unit); inverter is the arbiter |
| Inverter Parallel Mode | 199 | Was mis-modelled as `enable_standard_self_consumption_logic` (GivTCP lineage); app name is "Enable Inverter Parallel Mode" |
| DC Discharge 10 SOC | 299 | App: "DC Discharge 10 Lower SOC % Limit" |
| Force Off Grid | 331 | Non-damaging but a **sustained islanding state** — not a momentary reboot. Bounded boolean; treat with care at call sites. |
| Smart Load slots 1–10 | 554–573 | 10 start/end timeslot pairs. Same bounded-HHMM class as the existing charge/discharge slots. |
| 3ph Real-Time Control | 1005 | 3ph mirror of HR166 |
| 3ph Battery Reserve | 1078 | App: "Battery Reserve %" |
| 3ph power rates/limits | 1108–1111 | Discharge Power Rate, Discharge Down To %, Charge Power Rate, Charge Up To % |
| 3ph AC charge slots | 1113–1116 | Slots 1 and 2 start/end (the *enables* 1112/1122/1123 were already safe) |
| 3ph DC discharge slots | 1118–1121 | Slots 1 and 2 start/end |
| Restart Hardware | 5010 | Same class as HR163 REBOOT — non-damaging |
| Enable Calculated Load | 5014 | |

### What's held back

- **HR479 "DC Wind CVT Voltage"** — raw voltage setpoint, unbounded 16-bit, no range guard. Admit only alongside a validating `set_*` command.
- **HR1002/1141** — in GivTCP's list but app doesn't expose them, and GivTCP has no write command for either (1141 is an orphaned list entry; 1002 has only a decode Def). Excluded on the app-is-authority principle.

### Also in this commit

- `ruff format` on `mock_plant.py` (b12 format slip — `ruff check` passed but `--check` failed on the PDU-logging args)
- Negative test fixture in `conftest.py` repointed HR199 → HR178 (now that 199 is write-safe), CRC recomputed
- `test_app_confirmed_registers_are_write_safe` locks in the batch

## Test plan

- [x] `tox` green (py314, format, lint, build, 705+ tests)
- [x] `register-safety-reviewer` run twice — once on the main batch, once specifically on the Smart Load addition
- [ ] @coderabbitai review
- [ ] Codex review gate

Refs #48

## Summary by CodeRabbit

* **New Features**
  * Expanded writable register support for battery self-heating, power control parameters (DC/AC charge/discharge), smart load timing, and real-time control functions.

* **Tests**
  * Added validation to ensure app controls align with safe-to-write registers.

* **Chores**
  * Code formatting improvements.


dewet22:
@coderabbitai review

dewet22-codex:

tests/client/test_commands.py
```diff
@@ -671,6 +671,45 @@ def test_ems_commands_are_write_safe():
...
+        1121,  # three-phase slots
+        5010,
+        5014,  # hardware-control block
+        *range(554, 574),  # SMART_LOAD_SLOT_1..10 start/end (HR554-573)
+    }
+    missing = app_confirmed - WRITE_SAFE_REGISTERS
+    assert not missing, f"app-confirmed registers not write-safe: {sorted(missing)}"
+    # HR479 (DC Wind CVT Voltage) is app-writable but held back — an unbounded
+    # voltage setpoint with no range guard; admit only with a validating set_*.
+    assert 479 not in WRITE_SAFE_REGISTERS
```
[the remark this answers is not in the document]
dewet22: Added in c8072da — each register in app_confirmed now calls encode(value=0) and asserts the result, symmetric with the negative fixture that verifies unsafe registers raise.

dewet22:
@deepsourcebot review

givenergy_modbus/pdu/write_registers.py
```diff
@@ -110,6 +110,60 @@
...
     2071,  # EMS_EXPORT_POWER_LIMIT
+    # ------------------------------------------------------------------------
+    # App-confirmed writable registers. Source: GivEnergy Android app "Direct
+    # Control" → Control tab (2026-06-02), the writable-register surface GE
+    # exposes to end users now that the cloud portal is being retired. The app
+    # listing each of these as a user-editable control is authoritative evidence
+    # they are safe to write. See givenergy-modbus#48.
+    104,  # ENABLE_BATTERY_SELF_HEATING — hardware/batch-gated: write may be rejected per-unit
+    172,  # ENABLE_MANUAL_BATTERY_HEATER — likely hardware-gated like 104
+    199,  # ENABLE_INVERTER_PARALLEL_MODE (was mis-modelled as self-consumption logic)
```
dewet22-codex, line 128: I think this also needs the read-side model name to move at the same time, or at least gain a compatibility alias. This PR admits HR199 under the app-confirmed meaning ENABLE_INVERTER_PARALLEL_MODE, but SinglePhaseInverter still decodes the same register as enable_standard_self_consumption_logic. That leaves downstream consumers with one public field name for reads and a different write-side meaning for the exact same HR, which is especially awkward now that raw writes to 199 encode successfully. I would either rename/deprecate the LUT field in this PR, or hold HR199 back until that API migration is included.

dewet22: Agreed the asymmetry is real. The model field rename (enable_standard_self_consumption_logic → enable_inverter_parallel_mode, with a deprecated @property alias so the old name stays readable) is the next PR in the queue — deliberately separated because it is a consumer-visible model API change that hass needs a coordinated handoff for (#106 migration context), while WRITE_SAFE admission is a purely additive PDU-layer change with no API surface. The transition window is short: the rename PR follows immediately.

PR #173, opened 2026-06-02
fix(ems): surface per-slot inverter status as raw hex code (#108)

### LINKED ISSUE #108: EMS rollup bitfield decode: per-slot status fields don't match observed values in real capture
Surfaced while wire-confirming #95 item 2. The `Ems` model packs per-slot statuses into two bitfields and the decoded values don't match the actual plant state.

### `IR(2043)` — per-meter status (8 slots × 2 bits)

From the 30-min capture in `tests/fixtures/captures/ems_plant_a/ems_arm1036_30min.log`:

- `meter_1_status` decodes as DISABLED, but `meter_1_power = −94 W` exactly matches `grid_meter_power` — slot 1 is clearly the active grid meter
- `meter_3_power = 583 W` (the solar/load CT) but `meter_3_status` is not ONLINE either
- `meter_6_status` and `meter_8_status` show ONLINE but their power values are zero (no physical meter at those slots in this plant)

### `IR(2045)` — per-inverter status (4 slots × 3 bits)

- `inverter_1_status` and `inverter_2_status` decode as WAITING, but both inverters are clearly producing (the −94 W grid export and 583 W generation are coming from them)
- `inverter_4_status` decodes as NORMAL despite no physical inverter at that slot (consumers should respect `inverter_count`, not iterate all four slots)

### Hypotheses

1. Bit positions in the model definitions are wrong (offsets / widths off)
2. Bit endianness assumption is wrong (low-to-high vs high-to-low within the register)
3. Status enum values are misaligned with the on-wire encoding

### What's needed

1. Compare decoded statuses against the live GE iOS app for the same plant at a known moment — only consumer known to render these correctly
2. Adjust the bitfield definitions and add a regression test pinned to the committed capture

Out of scope here: the rollup-data fields (power, SoC, temperature, serials) decode fine — this issue is just the packed status bitfields.

### Related

- #95 — surfaced during item 2 wire-confirmation
- #103 — committed the capture being analysed

dewet22:
## Summary

Fixes #108 (the known issue called out in the 2.1.0 release notes). Targets a **2.1.1 patch**.

The EMS rollup's per-slot **inverter** status (`IR(2045)`, 4×3-bit) was decoded through the main inverter `Status` enum, whose values don't match this field's encoding — so a *present, idle* inverter (code `2`) read as "WARNING" and an *empty* slot (code `0`) as "WAITING". The bit extraction was already correct; only the interpretation was wrong.

### What changed

- **Inverter status → hex string.** Surface the raw 3-bit code as a hex string — the house idiom for uninterpreted codes (cf. `device_type_code`, `fault_code`) — rather than mis-applying `Status`. Verified codes: `0` = empty slot, `2` = present/idle (against the committed EMS capture's SoC/power registers). `inverter_count` remains the authoritative presence signal.
- **Meter status unchanged** — it was already correct (`MeterStatus`, verified against the meter power registers); now pinned by the fixture regression test.

### Why hex rather than a new enum

The GivEnergy app abstracts the whole EMS into a single "EMS-PCS" object and never exposes per-slot inverter status, so there's no authoritative label for the codes. A speculative enum would either invent GE's wording or `ValueError` on an unseen code. Hex is honest and robust.

### Verification

- App session (against an EMS-seeded MockPlant) independently confirmed the broad EMS rollup decode — solar 583 W, grid 94 W exporting, battery idle/45%, energies, voltages, temps all matched — and plant status NORMAL. It just doesn't surface per-slot inverter status.
- `tox` green (py314 + format + lint + build).
- `test_ems.py` / `test_devices.py` updated to assert the hex codes; meter statuses still pinned against the capture's power readings.
- Decode-only change — no register-safety review needed.

Refs #108

## Summary by CodeRabbit

* **Bug Fixes**
  * Inverter per-slot status codes are now returned as raw hexadecimal values instead of parsed enum states, as the firmware encoding does not match expected Status values. Affects all four inverter slots in EMS monitoring.

### DIFF
--- a/givenergy_modbus/model/ems.py
+++ b/givenergy_modbus/model/ems.py
@@ -1,5 +1,6 @@
 """GivEnergy EMS (Energy Management System) data model."""
 
+from enum import IntEnum
 from typing import ClassVar
 
 from pydantic import ConfigDict, computed_field, create_model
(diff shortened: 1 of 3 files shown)


dewet22:
@coderabbitai review

dewet22-codex:

dewet22-codex (approved):
I’ve reviewed this and have no concerns to add. I checked the EMS decode path and the managed-inverter projection; I did not run local tox/ruff/tests because CI already covers the routine gate.

Issue #174, opened 2026-06-02
e_load_day (IR35) reads ~0 on GEN single-phase — real consumption-today register exists


dewet22:
## Summary

On a live GEN-class single-phase inverter (`SA2114G...`), `e_load_day` (mapped to
`IR(35)`, `givenergy_modbus/model/inverter.py:647`) reads ~0 while instantaneous load
power (`p_load_demand`, IR(42)) reads correctly (~780 W). The GE Android app's
Read-Only → Energy today shows **Consumption today: 21.40 kWh** for the same inverter,
so a working consumption-today register exists — **IR(35) is the wrong source** (a
GivTCP-era guess), not a hardware limitation.

Reported via givenergy-hass; blocking a user's GivTCP→givenergy_local long-term
statistics migration (the load pair is currently excluded as a stopgap).

## App "Energy today" block (cross-correlation ground truth)

Captured from the app, GEN single-phase inverter, ~21:12 local:

| App label | kWh | App label | kWh |
|---|---|---|---|
| PV generation today | 10.80 | Battery charge today | 5.40 |
| PV string 1 today | 5.10 | Battery discharge today | 3.90 |
| PV string 2 today | 4.40 | **Consumption today** | **21.40** |
| Grid import today | 14.50 | AC charge today | 3.80 |
| Grid export today | 0.10 | | |

Balance: 10.80 + 14.50 + 3.90 − 0.10 − 5.40 = 23.70 ≈ 21.40 (within conversion
losses) — internally consistent, so 21.40 is a real consumption figure.

## Plan

1. Probe the today-energy IR block on a GEN single-phase unit
   (`probe --type ir --base 0 --count 60`) and match against the values above. The
   register reading **214** (21.40 × 10, deci) is the real consumption-today.
2. Re-point `e_load_day` to that register.
3. Check for a consumption-**total** register: single-phase `Inverter` has no
   `e_load_total` def today (only `gateway.py` IR1656/1657 and
   `inverter_threephase.py` IR1398/1399 define it). Add it if one exists.

## Target

2.1.1 patch, alongside #108 (same class: wrong-register / decode correctness).
Needs a live probe to pin the register before the fix can land.

PR #176, opened 2026-06-03
fix(model): rename e_inverter_out_total → e_pv_generation_total + battery_max_power None for unmapped DTC

dewet22:
## Summary

Two small register-correctness fixes prompted by hass / energy-conductor coordination, following on from #174.

### 1. `e_inverter_out_total` (IR45/46) → `e_pv_generation_total`

The same sentinel cross-correlation that proved IR(44) is PV-generation-today (#174) confirmed **IR(45/46) is PV-generation-total**, not inverter AC output (app "PV generation total" = sentinel 294916.60). Renamed on `SinglePhaseInverter`, with a deprecated `e_inverter_out_total` @property alias for one release — mirroring the `e_inverter_out_day` → `e_pv_generation_today` change so hass can migrate its "Inverter Output" entity pair (today + total) coherently in one step.

**Single-phase only.** Unlike the day field, three-phase is *not* symmetric here: `ThreePhaseInverter` has its own native `e_inverter_out_total` (IR1362/3), a genuine register distinct from its PV total (`e_pv_total`, IR1374/5). That field is left untouched and stays non-deprecated. The single-phase IR(45/46) also leaks onto 3ph as `e_pv_generation_total` (unverified there), consistent with the existing `e_pv_generation_today` leak — left for #48.

### 2. `battery_max_power` returns `None` for unmapped DTC (was `0`)

`_DTC_BATPOWER` is an incomplete, hardware-unverified DTC→watts lookup. Defaulting an unknown model to `0` reads as "no battery capacity" to a consumer (e.g. a watt↔percent limit conversion or a dispatch coordinator) when the truth is "rated power unknown". Returns `None` so callers can distinguish unavailable from zero.

## Test plan

- [ ] CI green
- [x] `tox`-env pytest green (741 passed); ruff check + format clean on touched files
- [x] New `test_e_pv_generation_total_rename_and_deprecated_alias` covers single-phase rename+alias and the 3ph native-field / leak behaviour
- [x] `test_battery_max_power` updated: unmapped DTC → `None`
- [ ] Codex review

### DIFF
--- a/givenergy_modbus/model/inverter.py
+++ b/givenergy_modbus/model/inverter.py
@@ -662,11 +662,15 @@ class SinglePhaseInverterRegisterGetter(RegisterGetter):
         # Inverter AC grid-terminal apparent power (VA); pairs with IR(24), not IR(30)
         # (IR43/IR24 → plausible PF ~0.9; IR43/IR30 → impossible). Same node as IR(24).
         "p_grid_apparent": Def(C.uint16, None, IR(43), max=50000),
-        # IR(44) is PV-generation-today (sentinel cross-correlation, #174), not the
-        # inverter AC output. Renamed from "e_inverter_out_day"; the old name is kept
-        # as a deprecated @property alias on both inverter classes for a release.
+        # IR(44) is PV-generation-today and IR(45/46) is PV-generation-total (both
+        # sentinel cross-correlation, #174), not the inverter AC output. Renamed from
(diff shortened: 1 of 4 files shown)


dewet22-codex:

dewet22-codex (approved):
I’ve reviewed this and have no concerns to add. I checked the single-phase rename, the three-phase field shadowing/compatibility behaviour, and the unknown battery-power case; I did not run local tox/ruff/tests because CI already covers the routine gate.

PR #194, opened 2026-06-05
feat(model): add _ThreePhaseCommands and _EmsCommands mixins (#75)

dewet22:
## Summary

Adds model-specific command mixins that expose three-phase- and EMS-only setters as instance methods on the model classes that actually accept them — closing the structural half of #75.

- `_ThreePhaseCommands` composed onto `ThreePhaseInverter` — `set_ac_charge`, `set_force_charge`, `set_force_discharge`; HR(1112/1122/1123) added to its `WRITE_SAFE_REGISTERS`.
- `_EmsCommands` composed onto `Ems` (peer device, deliberately *not* inheriting from `_InverterCommands`) — full EMS plant surface (master enable; charge/discharge/export slots and their endpoint variants; per-slot target SoC; export power limit); EMS-only HR(2040, 2044–2071) allowlist.
- `_InverterCommands` gap-fill: `set_export_priority` and `set_enable_eps` (AC-coupled; registers already in the universal allowlist).
- Deprecated primitives (`enable_charge`, `set_shallow_charge`, etc.) intentionally stay off the mixin.
- AC-limit commands (HR 313/314) and pause-mode commands (HR 318–320) remain `commands.*`-only — model-vs-firmware applicability still pending wire-data disambiguation.

## What's deferred (and why this PR isn't the whole of #75)

This PR is the **structural** half of #75 — surfacing the right methods on the right model types. The **enforcement** half (rejecting setters that don't apply to the detected model at write time, rather than just not exposing them as instance methods) is left for the #106 typed-device work in v2.2. The per-mixin `WRITE_SAFE_REGISTERS` sets I've added are declared but `WriteHoldingRegisterRequest.ensure_valid_state()` still consults the global allowlist — that's intentional.

## Docs

`docs/usage.md` gets a rewrite of the stale "not yet exposed via the inverter API" disclaimer, an "EMS command API" snippet pointing at `plant.ems`, and per-row markers on the command tables (✦ three-phase / ▣ ems / ⛔ commands-only).

## Open question (non-blocking)

`set_export_slot` currently lives only on `_EmsCommands`. The pre-existing docs note in usage.md says export slots are shared with the standalone inverter — if that's accurate the methods might belong on `_InverterCommands` too. I've kept them EMS-only for now (matches the issue body + the EMS-only register addresses 2062–2069); revisit if wire data turns up.

## Test plan

- [x] `uv run --group test pytest` — 769 tests pass (added 37 to `tests/client/test_inverter_commands.py` covering composition, allowlist semantics, method delegation, encode round-trip, and isolation from single-phase)
- [x] `uv run --group test tox` — `py314`, `format`, `lint`, `build` all green
- [ ] No downstream consumer impact expected — mixin methods are additive


dewet22-codex:

docs/usage.md
```diff
@@ -190,9 +190,30 @@ await client.one_shot_command(inverter.set_enable_discharge(True))
 await client.one_shot_command(inverter.set_charge_slot(1, my_timeslot))
 ```
 
-The inverter knows its own `slot_map`, so slot setters don't need it threaded through. The mixin is composed onto both `SinglePhaseInverter` and `ThreePhaseInverter`; the methods listed below ("Available commands") are universally available.
+The inverter knows its own `slot_map`, so slot setters don't need it threaded through. The base `_InverterCommands` mixin is composed onto both `SinglePhaseInverter` and `ThreePhaseInverter`; methods carrying no mixin marker in the tables below are universally available.
```
dewet22-codex: I do not think the unmarked methods can be described as universally applicable yet. For example, `ThreePhaseInverter.set_mode_storage()` comes from `_InverterCommands`, but delegates to the module-level `set_mode_storage()`, which hardcodes `SINGLE_PHASE_SLOTS` and therefore writes HR(56/57,44/45) instead of the three-phase HR(1118–1121) slots. Several reserve/limit helpers similarly still use single-phase register numbers. With this wording, the recommended instance API can write the wrong register family on a three-phase unit. I would keep the affected methods out of the universal surface/mark them commands-only until their delegation is model-aware, or make them use the instance’s `slot_map` and model-specific register mapping.

givenergy_modbus/client/commands.py
```diff
@@ -907,3 +916,150 @@ def set_mode_storage(
...
+
+    Composed onto `ThreePhaseInverter` alongside `_InverterCommands`. The
+    `WRITE_SAFE_REGISTERS` set unions the base allowlist with the three-phase
+    additions, so any future write-time enforcement that consults the mixin's
+    allowlist (rather than the global PDU-level set) gets the right view for
+    a three-phase target.
+    """
+
+    WRITE_SAFE_REGISTERS: ClassVar[frozenset[int]] = _InverterCommands.WRITE_SAFE_REGISTERS | frozenset(
+        {
```
dewet22-codex: This does not yet give future model-aware enforcement the right three-phase view described above. `ThreePhaseInverter` inherits the base slot methods, and its `slot_map` makes those methods emit HR(1113–1116) and HR(1118–1121); it also exposes `set_battery_reserve_soc()`, which emits HR(1078). None of those registers are in this set, so enforcement against `inverter.WRITE_SAFE_REGISTERS` would reject valid commands already exposed on the same object. I would include the native three-phase registers represented by the inherited command surface here and extend the superset test accordingly.

dewet22: Fixed in 107b339. Added the native three-phase registers the inherited model-aware surface emits — 1078 (`set_battery_reserve_soc`), 1113-1116 / 1118-1121 (charge/discharge slots 1-2 via `THREE_PHASE_SLOTS`) — to `_ThreePhaseCommands.WRITE_SAFE_REGISTERS`, and extended the superset test. The slot-3-10 registers (246-289) were already in the base allowlist. Register-safety reviewed before pushing. (The underlying methods sitting on the universal mixin is a separate model-awareness issue, now tracked in #203.)

docs/usage.md
dewet22: Good catch — this is the more important one. You're right that `set_mode_storage` (hardcodes the single-phase discharge slots) and `set_battery_reserve_soc` (three-phase-only HR1078) aren't model-aware, so the "universally applicable" wording was wrong. Fixed the docs in 107b339 to call both out explicitly and steer callers to the slot setters. The underlying model-aware delegation is a bigger job (→ #106), tracked now in #203 — this PR scopes to honest docs + the verified allowlist rather than the refactor itself.

givenergy_modbus/client/commands.py
```diff
@@ -907,3 +915,150 @@ def set_mode_storage(
...
+            1112,  # AC_CHARGE_ENABLE
+            1122,  # FORCE_DISCHARGE_ENABLE
+            1123,  # FORCE_CHARGE_ENABLE
+        }
+    )
```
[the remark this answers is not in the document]
dewet22: Fixed in 107b339 — added 1078, 1113-1116, 1118-1121 to `_ThreePhaseCommands.WRITE_SAFE_REGISTERS` (register-safety reviewed; the 246-289 slot-3-10 registers were already in the base set). The underlying methods being on the universal mixin is a separate model-awareness issue, now tracked in #203.

tests/client/test_inverter_commands.py
```diff
@@ -143,3 +162,131 @@ def test_frozen_still_enforced_after_mixin():
...
+def test_single_phase_does_not_inherit_three_phase_commands():
+    assert not issubclass(SinglePhaseInverter, _ThreePhaseCommands)
+
+
+def test_three_phase_write_safe_registers_is_superset_of_base():
+    """Three-phase allowlist must contain everything the base allows plus the 1112/1122/1123 trio."""
+    assert _InverterCommands.WRITE_SAFE_REGISTERS.issubset(_ThreePhaseCommands.WRITE_SAFE_REGISTERS)
+    assert {1112, 1122, 1123}.issubset(_ThreePhaseCommands.WRITE_SAFE_REGISTERS)
+    # AC-limit / pause-mode / EMS registers are still excluded — those mixins haven't landed yet.
+    assert {313, 314, 318, 319, 320, 2040, 2062}.isdisjoint(_ThreePhaseCommands.WRITE_SAFE_REGISTERS)
```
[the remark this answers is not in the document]
dewet22: Done in 107b339 — the superset assertion now checks `{1078, 1112-1116, 1118-1123}` is a subset of the three-phase allowlist.

tests/client/test_inverter_commands.py
```diff
@@ -143,3 +162,131 @@ def test_frozen_still_enforced_after_mixin():
...
+def test_ems_set_ems_charge_target_soc_emits_correct_write():
+    assert _ems().set_ems_charge_target_soc(2, 75) == [
+        WriteHoldingRegisterRequest(RegisterMap.EMS_CHARGE_TARGET_SOC_2, 75),
+    ]
+
+
+def test_ems_set_ems_export_power_limit_validates():
+    """Validation happens in the underlying primitive; mixin just delegates."""
+    with pytest.raises(ValueError, match=r"\[0-65535\]"):
+        _ems().set_ems_export_power_limit(70000)
```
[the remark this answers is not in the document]
dewet22: Done in 107b339 — added `test_ems_target_soc_validation_holds_through_mixin`, asserting the index `[1-3]` and SoC `[0-100]` bounds raise through the mixin for all three target-SoC setters.

givenergy_modbus/client/commands.py
```diff
@@ -827,6 +827,14 @@ def set_system_date_time(self, dt: datetime) -> list[TransparentRequest]:
         """Set the date & time of the inverter."""
         return set_system_date_time(dt)
 
+    def set_export_priority(self, priority: ExportPriority) -> list[TransparentRequest]:
```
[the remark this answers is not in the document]
dewet22: Declining this and the ~21 sibling DeepSource comments on these wrappers: they're intentional instance methods that form the command surface — meant to be called as `inverter.set_x(...)` / `ems.set_x(...)` and composed via mixins onto the model classes, not `@staticmethod`s. Converting them would change the API shape for a negligible per-instance saving. DeepSource has also since been removed from this repo (advisory-only, redundant with ruff/mypy/bandit/CodeQL), so these won't recur. Resolving the set.

dewet22-codex:
I’ve started re-reviewing this PR against the latest changes.

docs/usage.md
```diff
@@ -190,9 +190,36 @@ await client.one_shot_command(inverter.set_enable_discharge(True))
...
+The inverter knows its own `slot_map`, so slot setters don't need it threaded through. The base `_InverterCommands` mixin is composed onto both `SinglePhaseInverter` and `ThreePhaseInverter`; most methods carrying no mixin marker in the tables below apply to both.
 
-Three-phase-only and EMS-only commands (`set_ac_charge`, `set_force_charge`/`_discharge`, `set_battery_*_limit_ac`, `set_battery_pause_mode`, `set_pause_slot_*`, `set_ems_plant`, `set_export_slot_*`) are not yet exposed via the inverter API — call them on `commands.*` directly while their model-vs-firmware applicability is resolved (see [#75](https://github.com/dewet22/givenergy-modbus/issues/75)). They will appear on model-specific mixins in later 2.x minors.
+> **Not yet fully model-aware.** A few inherited methods delegate to primitives with hardcoded register numbers and are *not* safe to treat as universal until per-model command routing lands ([#75](https://github.com/dewet22/givenergy-modbus/issues/75), ultimately [#106](https://github.com/dewet22/givenergy-modbus/issues/106)):
```
dewet22-codex: The caveat still seems too narrow for the claim that most unmarked methods apply to both models. The three-phase LUT explicitly moves `battery_soc_reserve` from HR110 to HR1109 and `charge_target_soc` from HR116 to HR1111; it also says HR1112/1122 are semantically distinct from the single-phase HR96/59 enables. The inherited `set_battery_soc_reserve()`, `set_charge_target()`, `set_enable_charge()`, and `set_enable_discharge()` therefore still emit the single-phase registers on a three-phase instance, but are not called out here. Given these are hardware writes, I would either enumerate all known non-model-aware inherited methods here or avoid presenting the remaining unmarked surface as generally applicable until #203 fixes the routing.

givenergy_modbus/client/commands.py
```diff
@@ -907,3 +917,153 @@ def set_mode_storage(
...
+    """Commands that only apply to three-phase inverters.
+
+    Composed onto `ThreePhaseInverter` alongside `_InverterCommands`. The
+    `WRITE_SAFE_REGISTERS` set unions the base allowlist with the three-phase
+    additions, so any future write-time enforcement that consults the mixin's
+    allowlist (rather than the global PDU-level set) gets the right view for
+    a three-phase target.
+    """
+
+    WRITE_SAFE_REGISTERS: ClassVar[frozenset[int]] = _InverterCommands.WRITE_SAFE_REGISTERS | frozenset(
```
dewet22-codex: Adding the native registers fixes the false-negative side, but the union still retains known single-phase-only addresses such as HR94/95, HR56/57, HR110, HR116, HR96, and HR59. That means the future enforcement described in this docstring would still allow the unsafe inherited three-phase calls documented in #203, including `set_mode_storage()` and the reserve/target/enable helpers. I do not think this set can be described as the right three-phase view while it is a superset of the universal set. I would either make it an exact model allowlist by subtracting the known wrong-register entries, or describe it only as an incomplete declaration that must not be used for enforcement yet.

docs/usage.md
dewet22: Agreed, and broadened in 67858e4. The caveat now covers the full set rather than two methods: `set_charge_target` (HR116 vs three-phase HR1111), `set_battery_soc_reserve` (HR110 vs HR1109), `set_mode_storage` (HR56/57 vs HR1118-1121), and the enable helpers — flagged as single-phase-validated-only, pre-existing, and tracked for the model-aware fix in #203 → #106.

givenergy_modbus/client/commands.py
dewet22: Good call — rather than caveat around a half-correct allowlist, I dropped the per-model `WRITE_SAFE_REGISTERS` from both `_ThreePhaseCommands` and `_EmsCommands` entirely in 67858e4. They were never enforced (only `pdu.write_registers.WRITE_SAFE_REGISTERS` gates writes), and as you note the three-phase union inheriting the single-phase base (HR94/95, 56/57, 110, 116, 96, 59) would have green-lit the inherited wrong-register writes. Write-safety stays PDU-level; the per-model allowlists are deferred to #106 (where they can be correct once commands resolve registers per model), documented there, with the routing bug on #203.

dewet22-codex:
I’ve started one final review against the latest changes.

docs/usage.md
```diff
@@ -203,70 +232,80 @@ for tests and lower-level integration.
...
 individual primitives turn out to be persistent footguns when used without an
 inverter (e.g. routing the wrong `slot_map`), they may be `@deprecated`
-case-by-case as model-specific mixins land in later 2.x minors.
+case-by-case.
+
+### Where each command lives
+
+| Surface | Composed onto | Marker in tables below |
+|---|---|---|
+| `_InverterCommands` | `SinglePhaseInverter`, `ThreePhaseInverter` | _(none — universal)_ |
```
dewet22-codex: This legend still labels every unmarked `_InverterCommands` method as universal, which conflicts with the warning immediately above and makes the affected table rows (`set_charge_target`, `set_enable_charge`, `set_battery_soc_reserve`, and others) read as safe on three-phase again. Since these are hardware writes, I think the legend needs to describe the unmarked surface as inherited or single-phase-validated rather than universal; that keeps the tables aligned with the new caveat without needing to mark every affected row individually.

dewet22: Fixed in b151002. The legend no longer calls the unmarked surface "universal" — it's now labelled "inherited base surface; single-phase-validated (see the warning above)", with an added note that an unmarked row means inherited from `_InverterCommands`, not three-phase-correct. So the table and the warning are consistent now.

dewet22-codex:
I’ve started the final review against the latest changes.

dewet22-codex (approved):
I’ve completed the final review and have no remaining concerns. The revised documentation now clearly distinguishes the inherited command surface from commands validated for three-phase use.

PR #202, opened 2026-06-05
feat: support Python 3.11+ (lower floor from 3.14)

dewet22:
## What

Lowers the supported Python floor from 3.14 to **3.11**. The `>=3.14` pin only ever tracked the Home Assistant integration's pins — nothing in the library genuinely needs 3.14.

The only two blockers:
- 6 unparenthesized `except A, B:` clauses (3.14 syntax) → reparenthesized
- `from warnings import deprecated` (PEP 702, 3.13+) → `from typing_extensions import deprecated` (the backport; re-exports the stdlib one on 3.13+)

`StrEnum` and `X | Y` unions are already 3.11-safe, so nothing else changed.

Also:
- `requires-python >=3.11`; classifiers (3.11–3.14), ruff `target-version = py311`, and the tox `envlist` (now a `py311`–`py314` matrix) updated to match
- `typing-extensions` promoted to a declared dependency (was transitive via pydantic; now imported directly for `deprecated`)

## Why

Widens reach for non-HA consumers (e.g. Raspberry Pi OS Bookworm ships 3.11) without affecting the HA integration, which follows HA core's own pin regardless. 3.11 is the lowest non-near-EOL rung (3.10 reaches EOL Oct 2026; 3.11 Oct 2027).

## Verification

`uv run --group test tox` — full matrix green: **py311, py312, py313, py314** + format + lint + build. 757 tests pass on each.

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved Python 3 compatibility issues across core modules.

* **Chores**
  * Expanded Python version support to include Python 3.11, 3.12, and 3.13.
  * Updated development tooling and configuration targets.


dewet22-codex:

pyproject.toml
```diff
@@ -14,15 +14,19 @@ classifiers = [
...
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
     "Programming Language :: Python :: 3.14",
     "Topic :: Home Automation",
     "Topic :: Scientific/Engineering",
     "Topic :: System :: Networking :: Monitoring",
 ]
-requires-python = ">=3.14"
+requires-python = ">=3.11"
```
dewet22-codex, line 25: I think the `pyupgrade` target needs to move with this floor as well. `prek.toml` still runs pyupgrade with `--py313-plus`, so a later unrelated `prek` run is allowed to introduce rewrites that assume Python 3.13 and can silently break the newly advertised 3.11/3.12 support. I would change that hook to `--py311-plus` in this PR so the automatic fixer enforces the same compatibility contract as `requires-python` and Ruff.

dewet22: Good catch — fixed in 846e5f1. Moved `prek.toml`'s pyupgrade hook to `--py311-plus` so the auto-fixer enforces the same floor as `requires-python` and the ruff target-version. Verified the hook runs clean (no rewrites) at the new target.

Issue #207, opened 2026-06-06
refresh(): self-tuning skip-if-fresh for IR(0,60) (auto-adapt to consumer cadence)


dewet22:
Follow-up to #196. The opt-in `refresh(ir0_max_age=...)` skip-if-fresh shipped there requires the consumer to pick an absolute staleness threshold in seconds. This issue tracks a self-tuning variant that needs no threshold.

### Background

GivEnergy dongles fan out the responses to whoever polls them (the cloud, the app, another client) to every connected TCP client. So the network consumer continuously receives `IR(0,60)` (and the other live blocks) without us asking — confirmed across AIO/AC/EMS/Gen1 captures at a ~10–32s cloud-driven cadence (see `tests/debug/unsolicited_responses.py`). #196 added per-block ingestion timestamps (#65) and uses them to skip re-soliciting `IR(0,60)` when it was committed within `ir0_max_age` seconds.

### The idea

Instead of an absolute threshold, gate on *"did a fan-out frame for this block arrive since my last `refresh()`?"*. If yes, the passive stream is keeping pace with the consumer's own cadence → skip the solicit. If no, solicit. This auto-adapts:

- consumer polling slower than the fan-out (e.g. 60s vs ~12s) → never solicits `IR(0,60)`, zero added load;
- consumer polling faster than the fan-out → always solicits, because the stream genuinely can't meet their freshness need.

It removes the "pick a number" burden entirely.

### Open questions / caveats

- **Flapping** when the consumer interval ≈ the fan-out interval: jitter would cause it to skip some cycles and solicit others. May need a small hysteresis margin, which reintroduces a (smaller) tunable.
- **State**: needs to remember the per-block timestamp observed at the *previous* `refresh()` call, or compare against a refresh-start watermark.
- **Scope**: like #196, prove on `IR(0,60)` before generalising to the other fanned-out blocks.
- Should soak-test the absolute-threshold version (#196) on real hardware first to validate the fan-out assumption in the wild before adding the auto variant.

Not urgent — the opt-in absolute threshold covers the need for now; this is the ergonomic next step once it's soak-tested.

Issue #206, opened 2026-06-06
Reject degenerate all-zero register banks at commit (Pattern B) + surface staleness


dewet22:
## Problem

`Plant._commit_bank` commits an all-zero register bank over good cached data. Confirmed empirically: pushing an all-zero `IR(0,60)` `ReadInputRegistersResponse` through `plant.update()` on a plant seeded from the `hybrid_2_bat_a` fixture zeroes the live bank (`IR42` 3685→0, `IR0` 1→0, `IR52`→0).

Nothing currently catches it:
- `is_suspicious()` (`pdu/read_registers.py:128`) only matches #78 **Pattern A** — 16 *non-zero* constants at fixed offsets; an all-zero bank matches none.
- `is_coherent()` (`model/register.py:405`) is a no-op for IR banks (no serial field in `IR(0,60)`).
- bounds validation doesn't flag zeros; the whole-bank-discard in `_commit_bank` is still a `TODO`.

## This is #78 Pattern B, on the inverter bank

#78 documented two failure modes: **A** = non-zero deterministic garbage, single-frame, recovers on the next read (handled). **B** = "zero for an entire response block", occurring in *windows* (90 windows / 26h; 21 complete 5-field block dropouts; only 6 isolated). #147 tracks Pattern B on the **battery sub-bus**. The #199 `IR42` transient zeros look like the same Pattern B on the **inverter `IR(0,60)` bank** — sustained (≈ the reporter's "~1 min"), not a single-register glitch.

## Why it's been hard to action — and the resolution

An all-zero bank is ambiguous: a genuine dropout vs a legitimately absent/empty device (the `register.py:346` all-zero guard exists for the latter). But `_commit_bank(incoming, committed)` already holds the committed cache, so the disambiguator is contextual:

> previously non-zero, now all-zero ⇒ dropout ⇒ reject (keep last-good); always-zero ⇒ absent ⇒ leave alone.

## Proposed

1. Contextual all-zero-bank rejection in `_commit_bank`, covering inverter + battery banks — unifying the handling for #147 and #199.
2. Because Pattern B is sustained, pair with a staleness/freshness signal (#91 frozen-cache, #65 ingestion timestamps) so consumers can distinguish held-last-good from freshly-confirmed.
3. Regression test via the MockPlant harness (commit good data, then push an all-zero bank; assert the cache is retained and staleness is flagged).

## Open

[...]

Issue #199, opened 2026-06-05
Investigate: Gen2 IR42 (home load) emits transient zeros — library or consumer responsibility to debounce?


dewet22:
## Background

The ACBC field report (#10) observes that on Gen2, the home-load register `IR42`
(`p_load_demand`, `model/inverter.py:668`) intermittently reads `0` for 1–3 polls (~10–30s) even when
real load is non-zero — a dongle/inverter glitch, not a genuine zero. The fork debounces it (holds
the last non-zero value for ~120s) to avoid chart dips and kWh-delta corruption.

## What we do today

The library reports the register faithfully. A transient `0` is an **in-bounds** `uint16`, so the
existing out-of-bounds suppression (`model/register.py:345-359`) does not — and should not — catch
it. We currently pass it straight through.

## GivTCP dev3 cross-check

GivTCP applies **no** debouncing either — `IR42` is a raw `uint16` passthrough with no smoothing
anywhere in its model. So neither reference implementation filters this; debouncing is non-standard.

## The open question

Where should transient-zero handling live?

- **Consumer (e.g. the HA integration):** debouncing for charts/history is presentation policy, and
  both reference libraries leave the register raw — which argues the protocol library should keep
  reporting ground truth and let consumers smooth.
- **Library:** if the glitch is universal on Gen2 and always spurious, a documented opt-in filter or
  a "last-known-good" helper could save every consumer reinventing it.

Recorded here for a decision rather than pre-judged. Leaning consumer, but noting the option.

## Verification path

- Gen2 capture showing the zero-glitch cadence and that surrounding polls are non-zero (author has
  Gen2 AC hardware).

## Scope

Investigation / policy decision. Refs the ACBC field report; tracked from #197.

dewet22:
Field follow-up from @ACBCSoftware via `givenergy-cli capture`: a redacted 120s Gen2 capture showed no transient zeros in that window (stable load), consistent with the zeros being brief (1–3 polls) and tied to inverter state transitions — so a longer overnight capture is more likely to catch one (offered). Their consumer-side fix holds the last-good value until 12 consecutive zeros before accepting a genuine zero, with no false positives — a useful data point for the library-vs-consumer debounce question this issue is weighing.

ACBCSoftware:
Following up directly on the capture I shared (relayed in the earlier comment).

[...]

Issue #196, opened 2026-06-05
Investigate: Gen3/AIO local poll responsiveness may depend on request burst shape, not just device address — follow-up to #105


dewet22:
## Background

#105 ("AIO refresh times out") was resolved largely as an **addressing** problem (the `0x11→0x32`
inversion, #119). A separate fork now reports that even the current published library still does not
get fast local poll responses from Gen3/AIO units, and isolates a different suspected cause.

## Evidence (third-party, unverified field annotations)

`ACBCSoftware/acbc-givenergy-dashboard` @ `95afbc5` was built on the premise that the published
`givenergy-modbus` only sees ~300s cloud-sync broadcasts from Gen3/AIO, while GivTCP's
`givenergy_modbus_async` fork gets fast responses. To find the difference it ships two wire-capture
harnesses that tee the exact TX/RX bytes:

- `library-capture/library_capture.py` — our library, via `Client.capture_frames()`.
- `givtcp-capture/givtcp_capture.py` — GivTCP's fork, via an `asyncio.open_connection` monkeypatch.

Its `gen3-capture --givtcp` mode (`gen3-capture/gen3_capture.py:83-93, 393-523`) mirrors GivTCP's
Gen3 poll as a **5-frame burst** at `0x11`, 0.25s apart:

```
IR(0,60)    live inverter data
IR(180,60)  extended input regs
HR(0,60)    config block 1
HR(60,60)   config block 2   ← the author flags this as never-previously-tried
HR(120,60)  config block 3   ← likewise
```

The hypothesis: the two extra HR frames (or the burst/pacing) are what make the dongle answer
promptly — a request *shape* `refresh_plant()` may not reproduce.

## Why it's distinct from #105

#105's fix corrected *where* we poll (device address). This is about *how* we batch and pace the
requests within a refresh. Both can be true; addressing being right doesn't guarantee the burst
shape is.

## Verification path

- Run `library_capture` and `givtcp_capture` side by side against the same Gen3/AIO unit; diff the
  TX frame set, ordering, and inter-frame timing, and the latency to first matching RX.
- If GivTCP's burst is the unlock, evaluate adapting `refresh_plant()`'s request batching/pacing to
  match (without regressing single-phase, EMS, three-phase paths).
- Needs a capture from a Gen3/AIO reporter (hence `awaiting user data`).

## Scope

Investigation. Refs #105, #119; tracked from #197.

---

## Update: GivTCP dev3 confirms the burst shape from source

[...]

PR #208, opened 2026-06-06
feat: ingestion timestamps (#65), IR(0,60) skip-if-fresh (#196), all-zero bank rejection (#206)

dewet22:
Three related changes building on one another: ingestion timestamps as the foundation, then two consumers of them.

## #65 — per-block ingestion timestamps

`Plant` now records when each register block was last successfully committed, keyed by `(device_address, register_type, base_register)`, in a new `register_block_updated_at` map maintained in `update()`. `_commit_bank` returns a bool so a discarded bank (incoherent serial, all-zero dropout) does **not** record a timestamp — only data that actually landed counts as fresh. A `block_age()` helper exposes seconds-since-last-commit (None if never seen), with optional clock injection for deterministic tests. The field is excluded from `model_dump()` as ephemeral runtime bookkeeping.

The map is maintained continuously by the network consumer, so it captures both solicited responses **and** the dongle's unsolicited fan-out.

## #196 — opt-in IR(0,60) skip-if-fresh

GivEnergy dongles fan out the responses to whoever is polling them (the cloud, the app, another TCP client) to every connected client — so the consumer often already holds a recent `IR(0,60)` without us asking. Confirmed across every device class we have a capture for (AIO, AC, EMS, Gen1 hybrid): the live blocks arrive unsolicited on a ~10–32s cloud-driven cadence.

`refresh(ir0_max_age=...)` opts in to skipping the `IR(0,60)` solicit when its last ingest is within the threshold, sparing the often-flaky dongle a request. Defaults to `None` — always solicit, the historic behaviour. Scoped to `IR(0,60)` only for now. Degrades safely: the fan-out only exists while something else polls the unit, so on a cloud-disconnected dongle the block ages out and we solicit it as normal.

Also adds `tests/debug/unsolicited_responses.py`, the capture-analysis tool used to establish the fan-out behaviour.

## #206 — reject degenerate all-zero register banks (Pattern B)

`_commit_bank` committed an all-zero bank straight over good cached data. `is_coherent()` catches all-zero *battery* banks (serial check) but is a no-op for the serial-less inverter `IR(0,60)` bank, so the #199 inverter case (and #147 on the battery sub-bus) could zero out live values.

A contextual rejection now fires when a bank is entirely zero **and** its registers previously held non-zero data — a block-level dropout — keeping last-good. The "previously non-zero" gate resolves the dropout-vs-absent ambiguity: an always-zero/first-read bank (an absent device's ghost ACK) falls through to existing handling, so device detection is unaffected (verified). Staleness rides on #65 — a rejected bank records no timestamp, so `block_age()` grows.

Logged at **WARNING**: there is no on-wire capture of this event in any fixture (only a synthetic reproduction), so surfacing it turns every deployment into an evidence collector. It's a silent no-op on healthy systems.

## Notes

- `ir0_max_age` (#196) and the rejection (#206) are both **safe by construction** — opt-in / conservative-contextual respectively. Self-tuning skip-if-fresh is tracked in #207.
- 802 tests pass; ruff + mypy clean.

## Summary by CodeRabbit

# Release Notes

* **New Features**
  * Refresh operation can now conditionally skip re-fetching register data if cached data is recent enough.
  * System now tracks when each register block was last updated.

* **Bug Fixes**
  * System now properly detects and rejects all-zero register blocks that follow valid data, preventing misinterpretation of temporary data loss.

* **Documentation**
  * Added debug harness documentation for analyzing unsolicited response patterns.

### DIFF
--- a/givenergy_modbus/model/plant.py
+++ b/givenergy_modbus/model/plant.py
@@ -469,7 +469,7 @@ class Plant(GivEnergyBaseModel):
     # solicited responses and the dongle's unsolicited fan-out. Consumers (and refresh()'s
     # skip-if-fresh path, #196) use block_age() to reason about freshness. Excluded from
     # model_dump(): it's ephemeral runtime bookkeeping, not part of the plant's dumpable state.
-    register_block_updated_at: dict[tuple[int, str, int], datetime] = Field(default_factory=dict, exclude=True)
+    register_block_updated_at: dict[tuple[int, str, int, int], datetime] = Field(default_factory=dict, exclude=True)
 
     def model_post_init(self, __context: Any) -> None:
         """Ensure a default register cache is always present."""
@@ -540,7 +540,7 @@ class Plant(GivEnergyBaseModel):
         if isinstance(pdu, ReadHoldingRegistersResponse):
             incoming = {HR(k): v for k, v in pdu.to_dict().items()}
             if self._commit_bank(device_address, incoming):
-                self._stamp_block(device_address, "HR", pdu.base_register, received_at)
+                self._stamp_block(device_address, "HR", pdu.base_register, pdu.register_count, received_at)
         elif isinstance(pdu, ReadInputRegistersResponse):
             if pdu.is_suspicious():
                 # Pattern A dongle-side substitution from #78 — known fingerprint of 16 fixed
@@ -548,7 +548,7 @@ class Plant(GivEnergyBaseModel):
                 return
             incoming = {IR(k): v for k, v in pdu.to_dict().items()}  # type: ignore[misc]
             if self._commit_bank(device_address, incoming):
-                self._stamp_block(device_address, "IR", pdu.base_register, received_at)
+                self._stamp_block(device_address, "IR", pdu.base_register, pdu.register_count, received_at)
         elif isinstance(pdu, WriteHoldingRegisterResponse):
             if pdu.register == 0:
                 _logger.warning(f"Ignoring, likely corrupt: {pdu}")
@@ -612,24 +612,50 @@ class Plant(GivEnergyBaseModel):
         return True
 
     def _stamp_block(
-        self, device_address: int, reg_type: str, base_register: int, received_at: datetime | None
+        self,
+        device_address: int,
+        reg_type: str,
+        base_register: int,
+        register_count: int,
+        received_at: datetime | None,
     ) -> None:
-        """Record the ingestion time of a committed register block (#65)."""
-        self.register_block_updated_at[(device_address, reg_type, base_register)] = received_at or datetime.now(UTC)
+        """Record the ingestion time of a committed register block (#65).
+
+        The key includes ``register_count`` so that a partial response (e.g. IR(0,1)) does
+        not mark a full 60-register block as fresh — skip-if-fresh (#196) specifically checks
+        for the IR(0,60) key (count=60). See Codex review on PR #208.
+        """
+        ts = received_at or datetime.now(UTC)
+        if ts.tzinfo is None:
+            ts = ts.replace(tzinfo=UTC)
+        self.register_block_updated_at[(device_address, reg_type, base_register, register_count)] = ts
 
     def block_age(
-        self, device_address: int, reg_type: str, base_register: int, *, now: datetime | None = None
+        self,
+        device_address: int,
+        reg_type: str,
+        base_register: int,
+        register_count: int,
+        *,
+        now: datetime | None = None,
     ) -> float | None:
         """Seconds since a register block was last committed, or None if never seen.
 
-        ``reg_type`` is ``"HR"`` or ``"IR"``. Used to reason about freshness — e.g.
-        refresh()'s skip-if-fresh path for the fan-out IR(0,60) block (#196), and the
-        staleness signal that pairs with all-zero-bank rejection (#206).
+        ``reg_type`` is ``"HR"`` or ``"IR"``. ``register_count`` must match the count used
+        when the block was stamped — typically 60 for standard GivEnergy IR/HR blocks. This
+        prevents a partial response (e.g. IR(0,1)) from being mistaken for a full IR(0,60)
+        block by the skip-if-fresh logic in refresh() (#196).
+
+        Used to reason about freshness and staleness in the fan-out skip (#196) and Pattern B
+        all-zero rejection (#206).
         """
-        ts = self.register_block_updated_at.get((device_address, reg_type, base_register))
+        ts = self.register_block_updated_at.get((device_address, reg_type, base_register, register_count))
         if ts is None:
             return None
-        return ((now or datetime.now(UTC)) - ts).total_seconds()
+        effective_now = now or datetime.now(UTC)
+        if effective_now.tzinfo is None:
+            effective_now = effective_now.replace(tzinfo=UTC)
+        return (effective_now - ts).total_seconds()
 
     @property
     def inverter(self) -> SinglePhaseInverter | ThreePhaseInverter:
--- a/givenergy_modbus/model/plant.py
+++ b/givenergy_modbus/model/plant.py
@@ -569,6 +569,27 @@ class Plant(GivEnergyBaseModel):
         so the caller only records an ingestion timestamp (#65) for banks that actually
         landed.
         """
+        cache = self.register_caches[device_address]
+        # Pattern B (#78/#147/#199, tracked in #206): a bank that previously held non-zero data and
+        # now reads entirely zero is a block-level dropout (an empty page served during a
+        # transition), not real data — reject it and keep last-good. Contextual, to resolve the
+        # dropout-vs-absent ambiguity that made #147 hard to action: was-non-zero & now-all-zero =
+        # dropout -> reject; always-zero (absent / first read) -> fall through to the existing
+        # serial-coherence / is_valid() handling that already treats all-zero as device-absence.
+        # Staleness is free: a rejected bank records no #65 timestamp, so block_age() keeps growing.
+        if incoming and all(v == 0 for v in incoming.values()) and any(cache.get(k) for k in incoming):
+            sample = next(iter(incoming))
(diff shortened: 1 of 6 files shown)


dewet22-codex:

givenergy_modbus/model/plant.py
```diff
@@ -526,14 +539,16 @@ def update(self, pdu: ClientIncomingMessage):
...
+                self._stamp_block(device_address, "HR", pdu.base_register, received_at)
         elif isinstance(pdu, ReadInputRegistersResponse):
             if pdu.is_suspicious():
                 # Pattern A dongle-side substitution from #78 — known fingerprint of 16 fixed
                 # constants. is_suspicious() logs at debug when it fires.
                 return
             incoming = {IR(k): v for k, v in pdu.to_dict().items()}  # type: ignore[misc]
-            self._commit_bank(device_address, incoming)
+            if self._commit_bank(device_address, incoming):
+                self._stamp_block(device_address, "IR", pdu.base_register, received_at)
```
dewet22-codex: This stamps freshness by `(device, type, base)` only, regardless of how many registers the response actually contained. A fan-out or lower-level caller reading, for example, IR(0,1) will therefore refresh the same key that `refresh(ir0_max_age=...)` uses for IR(0,60), causing the next full live-block read to be skipped while registers 1-59 may still be stale. Since the optimisation relies on the whole 60-register block being fresh, I think the tracked identity needs to include the response count/coverage, or the skip path needs a timestamp that is only updated by responses covering IR(0,60). A regression test with a short IR0 response would protect this boundary.

tests/debug/unsolicited_responses.py
```diff
@@ -0,0 +1,142 @@
...
+    dev = getattr(pdu, "device_address", None)
+    base = getattr(pdu, "base_register", None)
+    cnt = getattr(pdu, "register_count", None)
+    dev_s = f"0x{dev:02x}" if isinstance(dev, int) else dev
+    fc_s = f"0x{fc:02x}" if isinstance(fc, int) else fc
+    return f"{type(pdu).__name__} dev={dev_s} fc={fc_s} base={base} count={cnt}"
+
+
+def analyse(paths: list[str]) -> None:
+    events = sorted((e for p in paths for e in _frames(p)), key=lambda e: e[0])
```
dewet22-codex, line 92: Combining all files into one event stream allows requests from one capture/session to match responses from another when they share a shape hash. That can classify a genuinely unsolicited response as solicited, and cadence deltas can also span unrelated devices or sessions. Because the CLI explicitly accepts multiple capture paths and this tool is being used as evidence for the optimisation, I think matching/cadence state needs to remain scoped per capture (with results aggregated afterwards), unless the input files are explicitly known contiguous segments of one session.

dewet22-codex:
I’ve started re-reviewing this PR against the latest changes.

givenergy_modbus/model/plant.py
```diff
@@ -547,27 +562,74 @@ def update(self, pdu: ClientIncomingMessage):
...
+        """
+        cache = self.register_caches[device_address]
+        # Pattern B (#78/#147/#199, tracked in #206): a bank that previously held non-zero data and
+        # now reads entirely zero is a block-level dropout (an empty page served during a
+        # transition), not real data — reject it and keep last-good. Contextual, to resolve the
+        # dropout-vs-absent ambiguity that made #147 hard to action: was-non-zero & now-all-zero =
+        # dropout -> reject; always-zero (absent / first read) -> fall through to the existing
+        # serial-coherence / is_valid() handling that already treats all-zero as device-absence.
+        # Staleness is free: a rejected bank records no #65 timestamp, so block_age() keeps growing.
+        if incoming and all(v == 0 for v in incoming.values()) and any(cache.get(k) for k in incoming):
```
dewet22-codex: This rule also applies to arbitrary short reads, where an all-zero transition can be perfectly legitimate. For example, after a lower-level caller reads one live power register as non-zero, a later read of that same single register returning zero will satisfy this condition and be discarded forever, leaving stale non-zero data in the cache. The same applies to small groups of status/power fields that can all naturally settle to zero. I think Pattern B rejection needs to be restricted to the known full-page response shapes (or otherwise require evidence that this is a block dropout), rather than applying to every incoming mapping. A regression test for a one-register non-zero → zero transition would expose the current behaviour.

givenergy_modbus/model/plant.py
```diff
@@ -547,27 +562,53 @@ def update(self, pdu: ClientIncomingMessage):
...
+                # violation. When that happens, also raise this back to WARNING — at that point
+                # it has user-visible consequences (a poll cycle's data is dropped).
         self.register_caches[device_address].update(incoming)
+        return True
+
+    def _stamp_block(
+        self, device_address: int, reg_type: str, base_register: int, received_at: datetime | None
+    ) -> None:
+        """Record the ingestion time of a committed register block (#65)."""
+        self.register_block_updated_at[(device_address, reg_type, base_register)] = received_at or datetime.now(UTC)
```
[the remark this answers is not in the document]
dewet22: Fixed in 8bb0ac0 — normalisation applied at both ends: a naive `received_at` passed to `_stamp_block` is replaced with `ts.replace(tzinfo=UTC)` before storage, and `block_age`'s `effective_now` is normalised the same way. The public interface still accepts naive datetimes but will no longer raise `TypeError` on subtraction.

givenergy_modbus/model/plant.py
```diff
@@ -547,27 +562,53 @@ def update(self, pdu: ClientIncomingMessage):
...
+        """Seconds since a register block was last committed, or None if never seen.
+
+        ``reg_type`` is ``"HR"`` or ``"IR"``. Used to reason about freshness — e.g.
+        refresh()'s skip-if-fresh path for the fan-out IR(0,60) block (#196), and the
+        staleness signal that pairs with all-zero-bank rejection (#206).
+        """
+        ts = self.register_block_updated_at.get((device_address, reg_type, base_register))
+        if ts is None:
+            return None
+        return ((now or datetime.now(UTC)) - ts).total_seconds()
```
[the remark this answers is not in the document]
dewet22: Fixed alongside the `_stamp_block` change in 8bb0ac0 — `block_age` now normalises `now` to UTC if it arrives timezone-naive.

tests/debug/unsolicited_responses.py
```diff
@@ -0,0 +1,142 @@
...
+        i, n = 0, len(buf)
+        while i + 6 <= n:
+            if buf[i : i + 4] != _MARKER:
+                i += 1
+                continue
+            flen = 6 + int.from_bytes(buf[i + 4 : i + 6], "big")
+            if flen < 18 or i + flen > n:
+                break
+            yield ts, parts[1], buf[i : i + flen]
+            i += flen
```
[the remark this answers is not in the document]
dewet22: Declining for now — this is a debug-only investigative script and capture files from the test fixtures are in the 10–60 minute range (under 1 MB). The line-by-line approach is correct advice for production code; for a one-off offline analysis tool the current read_text path is fine. Happy to revisit if the tool sees use against multi-GB captures.

givenergy_modbus/model/plant.py
dewet22: Fixed in 8bb0ac0. The stamp key is now a 4-tuple `(device_address, reg_type, base_register, register_count)`, and `block_age()` gains a required `register_count` positional arg. The `refresh()` skip path explicitly queries `block_age(inverter, "IR", 0, 60)` so only a stamp placed by a full 60-register response satisfies the freshness check. All callers in `update()` pass `pdu.register_count`. The regression test for a short-IR0 response was added as part of the Pattern B guard fix (separate commit).

tests/debug/unsolicited_responses.py
dewet22: Acknowledged, but intentional for now. The tool is designed to be pointed at a series of known-contiguous segments from one session (the CLI captures one session split across files by time). The help text documents the usage as `CAPTURE.log [CAPTURE2.log ...]` implying a single-session set. Cross-session mixing is a misuse that the user controls. Separating per-file state would be the right move if the tool is extended to handle multi-session analysis — I'll leave that as a follow-up note in the header comment rather than adding the complexity now.

givenergy_modbus/model/plant.py
dewet22: Fixed in 18bcd7a. `_commit_bank` now accepts an explicit `register_count` (passed from `update()` via `pdu.register_count`), and the Pattern B guard is gated on `register_count >= 60`. A short read (single register, or any count below 60) bypasses Pattern B and commits normally. Regression test `test_commit_allows_short_read_zero_transition` covers the non-zero → zero single-register case.

tests/debug/README.md
```diff
@@ -60,3 +60,44 @@ One JSON object per OOB event, one event per line:
...
+
+### Usage
+
+```bash
+uv run python tests/debug/unsolicited_responses.py CAPTURE.log [CAPTURE2.log ...]
+```
+
+### Interpreting the output
+
+```
```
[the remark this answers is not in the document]
dewet22: Fixed in 18bcd7a — the fenced block at line 87 now opens with ```text.

tests/debug/unsolicited_responses.py
```diff
@@ -0,0 +1,142 @@
...
+        if isinstance(pdu, HeartbeatRequest):
+            hb_in += 1
+        elif isinstance(pdu, HeartbeatResponse):
+            hb_out += 1
+        elif isinstance(pdu, TransparentRequest):
+            requests += 1
+            try:
+                pending[pdu.expected_response().shape_hash()].append(ts)
+            except Exception:  # noqa: BLE001 — malformed request; skip
+                pass
```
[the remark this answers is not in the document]
dewet22: Fixed in 8bb0ac0 — added `# nosec B110` alongside the existing `# noqa: BLE001`. The exception is intentional (malformed request frame in a debug tool; silently skipping is correct).

PR #211, opened 2026-06-07
feat(model): per-inverter battery & HV-stack ownership (#106 Phase 2)

dewet22:
## Summary

- **`Inverter.batteries` / `.hv_stacks` populated**: `from_direct` and `merge` now accept `batteries` / `hv_stacks` keyword args; `Plant.inverters` injects the decoded sub-devices, keeping `devices.py` import-clean of concrete types. Blinded (EMS-rollup) inverters honestly carry `[]`.
- **`Plant.devices` nests sub-devices under INVERTER rows**: flat `BATTERY` / `HV_STACK` rows are gone; both LV batteries and HV stacks now live on `inverter_row.device.batteries` / `.hv_stacks`. Meters remain flat (Phase 1 limitation — no `device_address` on `Meter`).
- **Orphan guard**: when a gateway plant suppresses the (spurious) inverter row, any batteries / stacks that would have nested under it are emitted as flat rows rather than dropped.
- **Model layer stays additive**: `Plant.batteries`, `Plant.hv_stacks`, `Plant.number_batteries`, every other legacy accessor — all untouched.

## Why reconciliation by serial is deferred to Phase 3

The original spec framed Phase 2 as breaking (behind a capability flag, with serial reconciliation). Reconciliation only bites when a plant has multiple direct inverters or a merged direct+rollup view — which requires the multi-Client transport abstraction (#75 / Phase 3). On today's single-Client plants the one direct inverter trivially owns every battery in the cache, so this increment ships the honest additive slice and defers reconciliation to Phase 3.

## Consumer note (`Plant.devices` shape change vs 2.1.5)

The `Plant.devices` row shape is breaking versus Phase 1 (2.1.5), but that surface had no production consumer yet — this PR targets `v2.2`. Consumers (Home Assistant) should enumerate batteries via `inverter_row.device.batteries`, not by filtering for flat `BATTERY` rows.

## Test plan

- [ ] `tests/model/test_devices.py` — 4 new `Inverter` facade unit tests (injected batteries & HV stacks on `from_direct`/`merge`; blinded → `[]`)
- [ ] `tests/model/test_plant_devices.py` — fully rewritten to the nested contract (LV batteries, HV stacks, EMS blinded inverters, gateway orphan guard, meters flat, back-compat)
- [ ] 819 tests passing; `ruff check --fix` + `ruff format` clean; `tox` (py311–py314 + lint + build) exit 0

### DIFF
--- a/docs/plant-device-graph.md
+++ b/docs/plant-device-graph.md
@@ -83,12 +83,42 @@ where meaningful, and the already-decoded typed model.
 Consumers (Home Assistant) can now enumerate typed devices to name and scope
 entities per device-type instead of assuming a single inverter.
 
-### Phase 2 — per-inverter battery ownership
+> **Superseded by Phase 2:** Phase 1 emitted batteries and HV stacks as
+> top-level `BATTERY` / `HV_STACK` rows. Phase 2 nests them under their owning
+> inverter instead (see below). The `DeviceType.BATTERY` / `HV_STACK` members
+> remain the vocabulary; `Plant.devices` now surfaces them via
+> `inverter_row.device.batteries` / `.hv_stacks`.
 
-Move batteries under their owning inverter (`Inverter.batteries`) with
-reconciliation by serial. **Breaking** for consumers that enumerate a flat
-`Plant.batteries` at setup time, so it ships behind a capability flag with a
-consumer migration. Depends on serial reconciliation maturing.
+### Phase 2 — per-inverter battery & HV-stack ownership ✅ (this increment)
+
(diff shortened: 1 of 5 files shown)


dewet22-codex:

dewet22-codex (approved):
I’ve reviewed this and have no concerns to add. The ownership model is explicit about the current single-direct-inverter constraint, and the blinded-inverter and gateway-orphan cases are handled cleanly. The remaining red checks are Codecov coverage rather than a behavioural issue.

PR #214, opened 2026-06-07
feat(model): RegisterCache.redact_serials() — share-safe cache export (#212)

### LINKED ISSUE #212: feat: public RegisterCache serial-redaction helper (for export share-safety)
## Background

`givenergy-cli export` writes a full register dump to JSON; these files get shared on issues/forums. Unlike `capture` (which redacts on the wire via `FrameRedactor`), `export` currently leaks serials encoded in raw register-cache values — battery `IR(110–114)`, the EMS rollup slots, gateway AIO blocks, and the `HR` battery/inverter groups.

The CLI agent (givenergy-cli#47) has asked for a supported way to get a **redacted copy of a `RegisterCache`** that:
- decodes the same suffix-zeroed placeholders the wire capture produces (`CE2231G000`-style, parity with `FrameRedactor`);
- leaves non-serial registers untouched and is idempotent;
- covers every serial group the library knows (`_SERIAL_GROUPS` in `client.py`);
- unrecognised serial shapes pass through unchanged.

## Reference implementation

The algorithm already lives in `FrameRedactor._redact_frame` (`client.py:199–216`) + `_SERIAL_GROUPS` (`client.py:48–97`). The gap is that it operates on PDU byte frames, not on a `RegisterCache` directly.

## Proposed surface

A method or free function:

```python
# Option A — method on RegisterCache
cache.redact_serials() -> RegisterCache

# Option B — free function
from givenergy_modbus.model.register_cache import redact_serials
redact_serials(cache: RegisterCache) -> RegisterCache
```

The CLI export path would call this on each cache before serialising.

## Out of scope here

Top-level string serials (`inverter_serial_number`, `data_adapter_serial_number`) are already handled CLI-side via `Converter.redact_serial`.

Relates to givenergy-cli#47.

dewet22:
## Summary

- Adds `RegisterCache.redact_serials() -> RegisterCache` — returns a redacted copy with all known serial-number registers zeroed out.
- Produces the same `AAYYWWA000`-style placeholders as `FrameRedactor`, so a redacted export is indistinguishable from a redacted wire capture.
- Driven by the model LUTs (every `Converter.serial` field, across inverter/battery/EMS/gateway) — no register addresses hardcoded; automatically covers new serial fields added in future.
- Groups absent from the cache are skipped; unrecognised serial shapes pass through unchanged; original is never mutated.

Unblocks [givenergy-cli#47](https://github.com/dewet22/givenergy-cli/issues/47) share-safe `export` command.

Closes #212.

## Test plan

- [ ] 7 new unit tests in `test_register_cache.py`: battery IR group redacted, inverter HR group redacted, idempotent, unrecognised shape passthrough, absent group not injected, original not mutated, empty cache.
- [ ] 827 tests pass; `tox` (py311–py314 + lint + build) exit 0.

### DIFF
--- a/givenergy_modbus/model/register_cache.py
+++ b/givenergy_modbus/model/register_cache.py
@@ -11,6 +11,40 @@ if TYPE_CHECKING:
 
 _logger = logging.getLogger(__name__)
 
+_SERIAL_GROUPS: "list[tuple[str, int, int]] | None" = None
+
+
+def _get_serial_groups() -> "list[tuple[str, int, int]]":
+    """Return (reg_type, base, count) for every C.serial register group (built once)."""
+    global _SERIAL_GROUPS
+    if _SERIAL_GROUPS is not None:
+        return _SERIAL_GROUPS
+    from givenergy_modbus.model import battery, ems, gateway, inverter
+    from givenergy_modbus.model.register import Converter
+
+    seen: set[tuple[str, int, int]] = set()
+    groups: list[tuple[str, int, int]] = []
+    for module in (inverter, battery, ems, gateway):
+        for attr in dir(module):
+            cls = getattr(module, attr)
+            if not isinstance(cls, type):
+                continue
+            lut = getattr(cls, "REGISTER_LUT", None)
+            if not lut:
+                continue
+            for _field, defn in lut.items():
+                pre_conv = defn.pre_conv[0] if isinstance(defn.pre_conv, tuple) else defn.pre_conv
+                if pre_conv is Converter.serial and defn.registers:
+                    reg_type = type(defn.registers[0]).__name__
+                    base = defn.registers[0]._idx
+                    count = len(defn.registers)
+                    key = (reg_type, base, count)
+                    if key not in seen:
+                        seen.add(key)
+                        groups.append(key)
+    _SERIAL_GROUPS = groups
+    return _SERIAL_GROUPS
+
 
 class RegisterCache(defaultdict[Register, int]):
     """Holds a cache of Registers populated after querying a device."""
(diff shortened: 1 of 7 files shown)


dewet22-codex:

givenergy_modbus/model/register_cache.py
```diff
@@ -11,6 +11,40 @@
...
+    """Return (reg_type, base, count) for every C.serial register group (built once)."""
+    global _SERIAL_GROUPS
+    if _SERIAL_GROUPS is not None:
+        return _SERIAL_GROUPS
+    from givenergy_modbus.model import battery, ems, gateway, inverter
+    from givenergy_modbus.model.register import Converter
+
+    seen: set[tuple[str, int, int]] = set()
+    groups: list[tuple[str, int, int]] = []
+    for module in (inverter, battery, ems, gateway):
```
dewet22-codex, line 42: This discovery misses the known HV BMU serials. `Bmu.from_register_cache()` decodes each serial manually from `IR(114 + 120*bmu_index)` through the next four registers, so those fields have no `REGISTER_LUT`/`Converter.serial` entry and are never added here. A share-safe export of an HV stack cache will therefore retain each BMU’s full serial number. I think these dynamic groups need to be included explicitly (or represented through shared serial metadata) before this can provide the advertised safe export.

givenergy_modbus/model/register_cache.py
```diff
@@ -81,6 +115,39 @@ def to_datetime(self, y: Register, m: Register, d: Register, h: Register, min: R
...
+        _reg_cls: dict[str, type[Register]] = {"HR": HR, "IR": IR}
+        result = RegisterCache(dict(self))
+        for reg_type, base, count in _get_serial_groups():
+            reg_cls = _reg_cls.get(reg_type)
+            if reg_cls is None:
+                continue
+            regs = [reg_cls(base + i) for i in range(count)]
+            if not all(r in self for r in regs):
+                continue
+            raw = b"".join(self[r].to_bytes(2, "big") for r in regs)
```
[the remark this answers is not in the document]
dewet22: Applied — now uses `isinstance(self.get(r), int)` and masks with `& 0xFFFF` before `to_bytes()`.

givenergy_modbus/model/register_cache.py
dewet22: Fixed — added explicit groups for `IR(114 + 120*i)` up to `_MAX_BMUS_PER_BCU = 8`. Absent modules are harmlessly skipped by the present-registers check.

dewet22-codex:

dewet22-codex (approved):
I’ve re-reviewed the latest changes and have no concerns to add. The BMU serial groups now close the previously identified privacy gap, and the defensive handling keeps malformed cache values from breaking redaction. All checks are green.

Issue #192, opened 2026-06-05
Expose per-module AIO battery data (serial, 24 cell voltages, temperatures) as enumerable objects in Plant


dewet22:
## What the integration needs (API boundary)

A way to enumerate the modules within each AIO battery pack from the `Plant` model, where each module exposes: its serial, 24 cell voltages, per-cell/group temperatures, and a reference back to its parent pack. Field names are the library's call — the integration just needs to iterate modules per pack and read those values, to create per-module Home Assistant devices (4 modules × 24 cells per AIO).

Context: the data is on the wire (BMUs at `0x50`–`0x53`; cell voltages around 3.2–3.3 V) and the BMU register definitions exist (`hv_bcu.py:89-158`), but aren't surfaced to consumers as enumerable per-module objects under each pack.

## Must not regress

- The existing pack-level `plant.batteries` list and its fields.

## Deferred / out of scope

- HA device/entity creation (givenergy-hass side, parked until this lands).
- Serial canonicalisation (separate issue).

## Evidence

- dewet22/givenergy-hass#95 (GivTCP per-module screenshots)
- `aio1.txt` capture

dewet22:
A concrete **decode template** that may be useful for shaping the per-module API, with an important
caveat about *which* battery bus it targets.

`ACBCSoftware/acbc-givenergy-dashboard` @ `95afbc5`, `dashboard_server.py:1972-2036`
(`_read_battery_module`) reads `IR(60,60)` from a battery module and decodes: 24 cell voltages (mV),
group temperatures (IR76-79) + MOSFET/PCB temp, max/min cell temp, cycle count, SOC, BMS firmware,
and a warning word (0 = healthy). That's close to the per-module field set this issue wants to
enumerate.

Caveat: it reads at the **Gen2 LV** convention (device-excluded CRC, `IR(60,…)` block), **not** the
HV BMU `0x50-0x53` register file (`hv_bcu.py:89-158`) that the AIO per-module data actually lives on.
So it's a **shape reference** for "what per-module fields to surface and how to scale them," not a
drop-in for the HV map. Noted in the umbrella cross-reference #197.

dewet22:
A GivTCP dev3 cross-check refines my earlier note about the ACBC `_read_battery_module` template.

GivTCP decodes the **LV** battery as **16 cells** (IR60–75, `model/battery.py`) and reserves the
[...]

PR #218, opened 2026-06-09
feat(model)!: remove unreliable first_battery_serial_number (#191)

### LINKED ISSUE #191: Confirm AIO battery module serials (HX-prefixed) are the module's own hardware serial
## What the integration needs (API boundary)

givenergy-hass surfaces `battery.serial_number` verbatim as the Home Assistant device identifier. It needs the decoded module serial to be the canonical, stable serial for that module.

Context: on a Gen-1 Gateway + single-phase AIO ([dewet22/givenergy-hass#95](https://github.com/dewet22/givenergy-hass/issues/95)) the AIO unit serial decodes with a `CH` prefix, but the four battery modules (BMUs at `0x50`–`0x53`) decode with an `HX` prefix, in both the frame header and the `IR(114..118)` serial field. Decoding looks faithful (big-endian, `register.py:62-83` / `hv_bcu.py:142-149`). These are most likely the modules' own hardware serials, distinct from the unit serial — not a byte-swap.

Ask: confirm via GivTCP cross-reference that the `HX…`-prefixed value is the module's genuine serial (or identify the canonical one GE exposes). The likely outcome is to document/keep it, not change it.

## Must not regress

- Existing `battery.serial_number` values for non-AIO packs (used downstream as device identifiers — any change is breaking for consumers).
- The faithful big-endian decode.

## Deferred / out of scope

- Per-module device modelling (separate issue).
- Any HA-side device-identifier migration (handled in givenergy-hass if the serial ever changes).

## Evidence

- dewet22/givenergy-hass#95
- `aio1.txt` capture (linked in #95)

dewet22:
## What

Removes `first_battery_serial_number` (HR 8–12) from `SinglePhaseInverter`.

## Why

A GivTCP-heritage field that earns its keep nowhere:

- **Unused** — not in `Plant.devices`, the `Inverter`/`UnifiedInverter` facade, any computed field, the cli, or docs; only the LUT def + test fixtures.
- **Unverifiable** — GivEnergy is bust, so nothing confirms HR(8–12) is ever a genuine battery serial. The only non-AIO value in the tree was a synthetic test fixture.
- **Harmful in its one known use** — on residential AIO firmware it holds the inverter's own serial with the first register word byte-swapped (`CH…` → `HC…`), not a battery serial. The Home Assistant integration was creating a ghost `HC…` device from it (dewet22/givenergy-hass#95). GivTCP ignores it entirely.
- **Superseded** — the genuine per-module serials (`HX…`, #192) come from the battery-module registers, not this field.

Removed outright — no deprecation alias.

## Redaction preserved

HR(8–12) stays in the capture-redaction set via an explicit serial group in `_build_serial_register_groups`, mirroring the BMU-serial precedent — the byte-swapped serial is recoverable to the real one, so it mustn't start leaking in shared captures just because the field is gone. Guarded by `test_frame_redactor_redacts_hr8_serial_register`.

## Breaking

`first_battery_serial_number` no longer appears in `SinglePhaseInverter.model_dump()`. It was never reliable.

Closes #191

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced serial number redaction mechanism to reliably protect device serials in legacy register storage locations, ensuring consistent data security across all supported device generations.

* **Tests**
  * Added new regression tests and updated existing test cases to comprehensively verify serial redaction functionality works correctly across all device types and configurations.

### DIFF
--- a/givenergy_modbus/client/client.py
+++ b/givenergy_modbus/client/client.py
@@ -90,6 +90,13 @@ def _build_serial_register_groups() -> "list[tuple[str, int, int]]":
                     if key not in seen:
                         seen.add(key)
                         groups.append(key)
+    # Legacy first_battery_serial_number registers (HR 8-12) — removed from the LUT
+    # (#191: GivTCP-heritage, unused, unverifiable), but still redacted: AIO firmware
+    # stores the unit serial here byte-swapped (CH… → HC…), recoverable to the real
+    # serial, so it must not leak in shared captures. Appended unconditionally — no LUT
+    # Def carries it any more, and a duplicate (if a future field re-used HR 8) would
+    # only redact it twice, which is idempotent.
+    groups.append(("HR", 8, 5))
     return groups
 
 
--- a/givenergy_modbus/model/inverter.py
+++ b/givenergy_modbus/model/inverter.py
@@ -411,7 +411,10 @@ class SinglePhaseInverterRegisterGetter(RegisterGetter):
         "num_phases": Def((C.duint8, 1), None, HR(3)),
         # HR(4-6) unused
         "enable_ammeter": Def(C.bool, None, HR(7)),
-        "first_battery_serial_number": Def(C.serial, None, HR(8), HR(9), HR(10), HR(11), HR(12)),
+        # HR(8-12) was first_battery_serial_number — removed (#191): GivTCP-heritage,
+        # unused, unverifiable; on AIO firmware it held the unit serial byte-swapped,
+        # not a battery serial. Still redacted in captures via an explicit serial group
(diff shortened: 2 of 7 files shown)


dewet22-codex:

dewet22-codex (approved):
Reviewed the field removal and preserved HR(8–12) capture redaction. I have no concerns to add.

dewet22-codex:

dewet22-codex (approved):
Re-reviewed after the rebase. The cache-level `redact_serials()` path now preserves HR(8–12) redaction consistently with captured frames, and I have no remaining concerns.

PR #219, opened 2026-06-09
feat(model): per-module AIO battery data as enumerable devices (#192)

### LINKED ISSUE #192: Expose per-module AIO battery data (serial, 24 cell voltages, temperatures) as enumerable objects in Plant

dewet22:
## What

Exposes per-module AIO battery data as enumerable devices (#192), closing the gap AberDino raised in dewet22/givenergy-hass#95: the integration only surfaced a pack-level battery for All-in-One units, not the per-module sub-devices GivTCP shows.

## Background

An AIO stores each removable battery module at its **own device address** (`0x50–0x53`), each a plain `IR(60–119)` block — 24 cell voltages, temperatures, and the module's own `HX`-prefixed serial. This is a *separate-address* layout, distinct from the HV `Bmu` stride model (modules offset within one BCU cache), which is left untouched. Confirmed against the real `tests/fixtures/captures/aio_a/` fixture (4 modules).

## Changes

- **`AioBatteryModule`** (`model/aio_battery.py`) — decoded from each module's own cache. The per-cell/temperature/serial decode is extracted from `Bmu` into a shared `hv_bcu.decode_cells_temps_serial` helper (AIO passes `base=0`); `Bmu` behaviour is unchanged.
- **Detection + refresh** — `PlantCapabilities.aio_battery_module_addresses`; `detect()` records `0x50+i` for an AIO BCU's modules; `refresh()` polls each so the caches populate on live hardware.
- **Enumeration** — `Plant.aio_battery_modules`; new `DeviceType.BATTERY_MODULE` rows in `Plant.devices`, nested under the AIO inverter and injected on its facade (`inverter.battery_modules`), keyed by module serial.
- **Redaction** — module serials (`IR 114–118`) are already covered by the existing serial group; regression test added.

## Scope / safety

Additive — no existing accessor, and neither the HV `Bmu`/`HvStack` path nor non-AIO plants change. Single-BCU AIO; multi-BCU module addressing is noted as a follow-up.

Full suite + tox py311–py314 green.

Closes #192

### DIFF
--- a/givenergy_modbus/client/client.py
+++ b/givenergy_modbus/client/client.py
@@ -465,21 +465,42 @@ class Client:
                 num_modules = bcu_cache.get(IR(64)) or 0
                 caps.bcu_stacks.append((i, num_modules))
 
+    #: Maximum number of battery modules on a single-BCU AIO (addresses 0x50–0x53).
+    _AIO_MAX_MODULES = 4
+
     async def _detect_aio_battery_modules(
-        self, caps: PlantCapabilities, probe_timeout: float, probe_retries: int
+        self,
+        caps: PlantCapabilities,
+        prior: PlantCapabilities | None,
+        probe_timeout: float,
+        probe_retries: int,
     ) -> None:
         """Populate caps.aio_battery_module_addresses for an All-in-One (#192).
 
-        The AIO's first BCU reports its module count (IR 64); each module answers at its
-        own device address 0x50 + index with a plain IR(60-119) block. Probe each, and
-        record those that respond with a present module (a valid serial). Single-BCU AIO
-        only — multi-BCU module addressing is a future extension.
+        Hinted mode (prior is not None): probe only the previously-seen module addresses
+        so detect() stays a confirmation pass rather than a fresh sweep — consistent with
+        the meter and battery hinted contracts.
+
+        Cold mode: derive candidates from the BCU-reported module count (IR 64), bounded
+        to _AIO_MAX_MODULES (0x50–0x53) to guard against stale or corrupt register values.
+
+        Single-BCU AIO only — multi-BCU module addressing is a future extension.
         """
-        if not caps.bcu_stacks:
-            return
-        _offset, num_modules = caps.bcu_stacks[0]
-        for i in range(num_modules):
-            addr = 0x50 + i
+        if prior is not None:
+            candidates: list[int] = list(prior.aio_battery_module_addresses)
+        else:
+            if not caps.bcu_stacks:
+                return
+            _offset, num_modules = caps.bcu_stacks[0]
+            if num_modules > self._AIO_MAX_MODULES:
+                _logger.warning(
+                    "detect: BCU reports %d modules but AIO maximum is %d — clamping",
+                    num_modules,
+                    self._AIO_MAX_MODULES,
+                )
+                num_modules = self._AIO_MAX_MODULES
+            candidates = [0x50 + i for i in range(num_modules)]
+        for addr in candidates:
             if not await self._probe(
                 ReadInputRegistersRequest(base_register=60, register_count=60, device_address=addr),
                 timeout=probe_timeout,
@@ -655,7 +676,7 @@ class Client:
         # battery module at its own device address (0x50+), distinct from the bcu_stacks
         # stride layout, so its per-module cell/temperature/serial data is reachable.
         if caps.device_type is Model.ALL_IN_ONE and caps.bcu_stacks:
-            await self._detect_aio_battery_modules(caps, probe_timeout, probe_retries)
+            await self._detect_aio_battery_modules(caps, prior, probe_timeout, probe_retries)
             _logger.info(
                 "detect: aio_battery_modules=[%s]",
                 ", ".join(f"0x{a:02x}" for a in caps.aio_battery_module_addresses),
--- a/givenergy_modbus/client/client.py
+++ b/givenergy_modbus/client/client.py
@@ -487,8 +487,12 @@ class Client:
             ):
                 continue
             cache = self.plant.register_caches.get(addr)
-            if cache is None or not AioBatteryModule.from_register_cache(cache, addr).is_valid():
-                _logger.debug("detect: AIO module probe at 0x%02x responded but is_valid()=False — skipping", addr)
+            try:
+                if cache is None or not AioBatteryModule.from_register_cache(cache, addr).is_valid():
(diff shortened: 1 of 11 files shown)


givenergy_modbus/client/client.py
```diff
@@ -464,6 +465,33 @@ async def _detect_bcu_stacks(
...
+            if not await self._probe(
+                ReadInputRegistersRequest(base_register=60, register_count=60, device_address=addr),
+                timeout=probe_timeout,
+                retries=probe_retries,
+            ):
+                continue
+            cache = self.plant.register_caches.get(addr)
+            if cache is None or not AioBatteryModule.from_register_cache(cache, addr).is_valid():
+                _logger.debug("detect: AIO module probe at 0x%02x responded but is_valid()=False — skipping", addr)
+                continue
```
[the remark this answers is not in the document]
dewet22: Done — wrapped in try/except in 41d1caa; decode failures log at DEBUG and skip the address rather than crashing detect().

givenergy_modbus/model/plant.py
```diff
@@ -750,6 +761,22 @@ def hv_stacks(self) -> list[HvStack]:
...
+        each carrying 24 cell voltages, temperatures, and the module's own serial. Empty for
+        non-AIO plants and until the module caches have been polled.
+        """
+        if not self.capabilities or not self.capabilities.aio_battery_module_addresses:
+            return []
+        return [
+            AioBatteryModule.from_register_cache(self.register_caches[addr], addr)
+            for addr in self.capabilities.aio_battery_module_addresses
+            if addr in self.register_caches
+        ]
```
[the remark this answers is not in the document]
dewet22: Done — switched to a loop with try/except in 41d1caa; decode failures log at ERROR and the remaining modules are still returned.

givenergy_modbus/model/plant.py
```diff
@@ -901,6 +936,19 @@ def devices(self) -> list[PlantDevice]:
...
+        # AIO per-module battery sub-devices (#192) — enumerable BATTERY_MODULE rows,
+        # owned by the inverter (also injected on its ``device.battery_modules``), keyed
+        # by the module's own HX-prefixed serial. GivTCP surfaces these as separate
+        # per-module devices; mirror that so consumers can name one HA device per module.
+        for module in self.aio_battery_modules:
+            rows.append(
+                PlantDevice(
+                    device_type=DeviceType.BATTERY_MODULE,
+                    device=module,
+                    serial_number=module.serial_number if module.is_valid() else None,  # type: ignore[attr-defined]
```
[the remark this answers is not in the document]
dewet22: Good catch — switched to _validated_serial(module) in 41d1caa; removes the type: ignore and keeps it consistent with the other device rows.

dewet22-codex:

givenergy_modbus/client/client.py
```diff
@@ -619,6 +651,16 @@ async def detect(
                 ", ".join(f"0x{0x70 + o:02x} (x{n})" for o, n in caps.bcu_stacks),
             )
 
+        # Step 2b — AIO per-module battery probing (#192). The All-in-One exposes each
+        # battery module at its own device address (0x50+), distinct from the bcu_stacks
+        # stride layout, so its per-module cell/temperature/serial data is reachable.
+        if caps.device_type is Model.ALL_IN_ONE and caps.bcu_stacks:
+            await self._detect_aio_battery_modules(caps, probe_timeout, probe_retries)
```
dewet22-codex: This makes hinted detection probe every count-derived module address, including addresses absent from `prior.aio_battery_module_addresses`. That breaks `detect()`'s documented confirm-only contract that a prior restricts the sweep to previously seen addresses. It also means a persisted AIO capability from before this field existed will discover modules and raise `PlantTopologyMismatch`, rather than performing a fast confirmation. I think hinted mode needs to probe only the prior module addresses, while the count-derived 0x50+ sweep remains cold-detect behaviour.

givenergy_modbus/client/client.py
```diff
@@ -464,6 +465,37 @@ async def _detect_bcu_stacks(
...
+
+        The AIO's first BCU reports its module count (IR 64); each module answers at its
+        own device address 0x50 + index with a plain IR(60-119) block. Probe each, and
+        record those that respond with a present module (a valid serial). Single-BCU AIO
+        only — multi-BCU module addressing is a future extension.
+        """
+        if not caps.bcu_stacks:
+            return
+        _offset, num_modules = caps.bcu_stacks[0]
+        for i in range(num_modules):
```
dewet22-codex: This is the first place the BCU-reported module count directly expands into network device addresses, but the count is used without validating the documented AIO range of 0x50–0x53. A corrupt or stale IR(64) value above four would make detection probe arbitrary addresses from 0x54 onwards; a large value can also make discovery unexpectedly long or eventually produce an invalid device address. I think the count needs to be bounded or rejected before this loop, with the four-address limit made explicit.

givenergy_modbus/client/client.py
dewet22: Fixed in 2b4a34e — _detect_aio_battery_modules now accepts prior; hinted mode sweeps only prior.aio_battery_module_addresses, consistent with the meter/battery hinted contract.

givenergy_modbus/client/client.py
dewet22: Fixed in 2b4a34e — cold mode now clamps num_modules to _AIO_MAX_MODULES (4) and logs a warning if the BCU-reported value exceeds the AIO range.

dewet22-codex:

dewet22-codex (approved):
The earlier findings are addressed, and I found no further concerns on re-review.

Issue #91, opened 2026-05-27
Detect "frozen BMS cache" during firmware updates and surface it as a stale-data signal


dewet22:
## Background

What this library exposes for an LV battery at `addr=0x32..0x37` is the inverter's cached copy of the BMS register page, not a live read of the battery — the BMS sits one RS485 hop below the inverter and the dongle re-exposes the inverter's cache via TCP. The [open-giv/bms-analysis](https://github.com/open-giv/bms-analysis) repository documents the BMS-side protocol in detail; for our purposes, the relevant invariant is that when a BMS goes quiet on RS485 (most commonly: bootloader mode during a firmware update), the inverter freezes its cache rather than blanking it.

From the TCP side this is indistinguishable from "live, but unchanged" without comparing multiple consecutive reads. Consumers (Home Assistant, dashboards) keep happily plotting the frozen values as if they were current.

## Observed instance

Captured this morning during a battery 1 firmware update on my own plant — see local `frames-0830.log` and the analysis notes from 2026-05-27. After 08:47 BST, battery 1's IR(0..59) response came back byte-for-byte identical for the remaining ~1 hour of the capture while battery 2 kept updating normally. The frozen field that's normally a temperature stuck at `00010008` and never moved. The matching SOC trace in Home Assistant showed a flat line through the same window, then a transient SOC=0 spike when the cache eventually unfroze.

## Proposed direction

A heuristic check at the `Plant._commit_bank` layer, something like:

- per `(device_address, register_block)` pair, hold the last N committed banks
- if the latest N are byte-identical for a register page that contains *any* field expected to vary (cell voltages, temperatures, counters), flag the bank as `stale`
- expose a per-device freshness predicate that integrations can poll to decide whether to mark a battery `unavailable`

The "expected to vary" set is the tricky bit — we'd want it narrow enough to avoid false positives during genuinely-idle periods (battery offline-but-healthy) and wide enough to catch real freezes. Probably a small whitelist of always-changing fields per device type, derived empirically.

## Caveats

- I've only seen this once, so the heuristic threshold (`N consecutive identical banks`) needs more data to tune. A 30-second cache freeze is normal during certain transitions; a 60-minute one is clearly broken.
- Not all consumers will want this — some prefer stale data to no data. The right surface is probably a flag/timestamp on `RegisterCache` rather than an automatic invalidation.
[...]

Issue #65, opened 2026-05-15
Track ingestion timestamps on RegisterCache to support freshness queries


dewet22:
Right now `RegisterCache` records register *values* but not *when* they arrived, which means downstream code can't reason about staleness. Passive mode is the most obviously affected — without timestamps it has no freshness signal at all — but the same primitive comes up in three places we've been working around:

- **Passive mode**, where there's no concept of "I just refreshed" because we don't initiate reads. The only signal of how current the data is is the timestamp on the last response we observed for that bank.
- **Lazy / observational detection** (the natural next step after #56). Building topology up from ingested responses needs a "when did I last see anything from device X" signal, which falls out for free once we record timestamps.
- **In-bounds-but-spurious data** (#57). Discriminating "this register changed because reality changed" from "this register changed because we got junk" gets a lot easier when you can rate-limit, compare to last known good, or weight newer reads against older ones — all of which need timestamps to make sense.

### Design

Two parallel concepts kept deliberately separate:

- **Bank timestamps.** Keyed by `(device_address, register_type, base_register)` with `base_register % 60 == 0`. Updated *atomically* when a response covers exactly that 60-register block. This is the freshness signal for register data, and the only thing consulted when reasoning about derived values built from multiple registers in the same bank (e.g. computed inverter power from voltage × current).
- **Per-device last-seen timestamps.** Updated on *any* valid response from the device, regardless of size. This answers "has this device gone silent?" and feeds the topology-from-traffic story for passive mode and lazy detection. Separate from data freshness on purpose — they answer different questions and conflating them is what made this tricky to reason about in earlier drafts.

The atomicity property is the one I'd specifically like to preserve: registers in the same bank are sampled by the firmware at the same instant, so derived values across them are temporally coherent. Per-register timestamps would technically be more granular but would lose that property.

### Ingest policy

| Response shape | Bank timestamp | Last-seen timestamp | Cache values |
|---|---|---|---|
| Full 60-aligned bank | Updated | Updated | Stored |
| Anything else (probe, partial bank, irregular sniff) | — | Updated | Dropped (debug-level log) |

Dropping non-conforming responses from the data cache is the part I'm least confident about, but the hypothesis is testable: if the debug-level drop count turns out to be meaningful in real deployments — particularly in passive mode where we'll see whatever the data adapter does — we'll know and can revisit. If it stays at noise levels, the alignment assumption was sound.

[...]

Issue #57, opened 2026-05-13
Bounds enforcement: flip detection-only validation to discard out-of-bounds register banks


dewet22:
## Background

Register bounds checking was added as part of the fork-merge work (`RegisterDefinition` `min`/`max` parameters, `RegisterGetter.validate_bank()`, `Plant._commit_bank()` calling it). The current implementation is deliberately detection-only: violations are logged at ERROR but the bank is still committed. This was intentional — run in observation mode first, validate the bounds are sensibly calibrated, then enforce.

## What needs to change

Three one-liner flips, in order:

**1. `RegisterGetter.get()`** — return `None` for out-of-bounds values instead of passing them through:
```python
# current
_logger.error(...); return val
# after
_logger.error(...); return None
```

**2. `RegisterGetter.validate_bank()`** — once `get()` returns `None` for violations, simplify the direct bounds re-check:
```python
# current: checks bounds directly because get() passes through bad values
# after: if getter.get(name) is None: ...
```

**3. `Plant._commit_bank()`** — discard the bank on any violation:
```python
# add `return` after the violations log
```

Each is marked with a `# TODO(enforcement)` comment in the code.

## When to do this

I'd hold off until the sweeping fork-merge changes have had some time in the wild and there's confidence that:

- The bounds are correctly calibrated (not too tight for any edge-case hardware variants).
- No legitimate readings are being flagged as violations in production logs.

Since GivEnergy is no longer producing firmware updates, the bounds are unlikely to need revision — but it's worth checking the ERROR logs from a few real deployments before flipping, to be sure nothing is being incorrectly flagged.

## Notes

- The double-buffer design (pending/committed cache split, so committed is never transiently overwritten by a bad frame) is a further step beyond this and is also deferred. The enforcement flip alone gives most of the protection.
- Detailed design is recorded in `fork-merge-plan.md` under "Cache integrity: double-buffering and bounds validation".

britkat1980:
I have found that handling min/max bounds can be tricky. given the huge variants of devices and power capabilities a true max value for something like inverter power varies hugely between a Gen1 Hybrid (3600w) and a 3PH HV (15000w).
I'm not sure if it makes sense to have some bounds which are device dependent, or if that's too much?
In GivTCP I implement different bounds for three phase devices as a minimum.

dewet22:
[...]

PR #220, opened 2026-06-09
feat(model): content-staleness primitive for frozen-BMS detection (#91)

dewet22:
## What

Lands the **content-staleness primitive** on `Plant`: per-register-block tracking of how long the block's content has been byte-identical, exposed read-only as `content_unchanged_seconds()`. This is the duration substrate for #91 (frozen-BMS-cache detection) — it reports a raw duration and makes **no freeze verdict** (see "Why no verdict yet").

## Background

When an LV battery's BMS enters firmware-update bootloader mode it goes quiet on RS485; the inverter freezes its cached register page and serves it byte-identical indefinitely. From the TCP side that's indistinguishable from "live but unchanged" without comparing consecutive reads over time. A ~1-hour freeze was captured on my own plant during a battery firmware update (2026-05-27).

## Changes

- `Plant._block_unchanged_since` (private) — per-`(device, reg_type, base, count)` `(content_hash, unchanged_since)`, where `unchanged_since` is the ingestion time of the **first** commit in the current byte-identical run. O(1) per block; survives unchanged runs of any length.
- `Plant._track_content_change()` — called from `update()` whenever `_commit_bank()` returns True, mirroring `_stamp_block()` (#65). Equal hash keeps the timestamp anchored; a changed/first hash resets it. Discarded banks update nothing.
- `Plant.content_unchanged_seconds(...)` — read-only accessor returning the duration (or `None` if never seen). Parallels `block_age()`.

## Why no verdict yet

The first cut exposed `is_battery_frozen()` with a consecutive-equality test. Codex review flagged it fires on healthy hardware, and replaying the committed corpus confirmed it:

| Capture | Battery | Samples | Max byte-identical streak |
|---|---|---|---|
| hybrid 10-min | 0x32 | 96 | **26** |
| hybrid 10-min | 0x33 | 93 | **23** |
| hybrid 60-min | 0x32 | 346 | 15 |
| AC 30-min | 0x32 | 64 | 7–8 |

The dongle fans out the same cached frame to every connected TCP client (so consecutive `update()` calls aren't independent polls), and battery telemetry genuinely doesn't change every poll. Distinguishing a bootloader freeze from a live-but-static BMS needs a **duration** threshold validated against more than the one freeze capture I have (the real incident never changed for a full hour; healthy 0x32 changed 14× in 10 min). `content_unchanged_seconds()` provides the measured duration that detector will threshold on; the threshold + policy is the deferred #91 work.

A second Codex pass caught that the original retained primitive (a bounded hash deque) discarded the timing a duration detector needs — and evicted the run's origin once a run exceeded the deque size. The `unchanged_since` tracker here fixes that: it records when the current content first appeared and survives arbitrarily long runs.

## Tests

Seven unit tests pinning the duration semantics: never-seen → None, accumulation while identical, reset on change, a 20-commit run reporting its true 600 s origin (the case the old deque couldn't represent), per-block keying, discarded banks leaving the tracker untouched, naive-`now` normalisation.

Full suite + tox py311–py314 + lint + build all green.

Refs #91 (stays open — the freeze *verdict* is deferred pending a validated threshold).

### DIFF
--- a/givenergy_modbus/model/plant.py
+++ b/givenergy_modbus/model/plant.py
@@ -1,5 +1,6 @@
 import logging
 import warnings
+from collections import deque
 from datetime import UTC, datetime
 from typing import Any, ClassVar
 
@@ -34,6 +35,11 @@ from givenergy_modbus.pdu import (
 )
 
 _logger = logging.getLogger(__name__)
+
+# Size of the rolling hash deque used for frozen-BMS-cache detection (#91).
+# Storing 10 hashes per block allows is_block_frozen() to be called with n ≤ 10
+# while keeping the per-block memory cost constant and small.
+_FROZEN_HASH_DEQUE_SIZE = 10
 
 # Models whose battery architecture is HV (BCU/BMU stacks rather than LV packs).
 # Coarse families "4" (HYBRID_3PH), "6" (AC_3PH) and "8" (ALL_IN_ONE and variants)
@@ -487,6 +493,13 @@ class Plant(GivEnergyBaseModel):
     # model_dump(): it's ephemeral runtime bookkeeping, not part of the plant's dumpable state.
     register_block_updated_at: dict[tuple[int, str, int, int], datetime] = Field(default_factory=dict, exclude=True)
 
+    # Rolling content-hash deques for frozen-BMS-cache detection (#91).  Keyed identically to
+    # register_block_updated_at; each value is a bounded deque of the last
+    # _FROZEN_HASH_DEQUE_SIZE content hashes of the committed bank.  Private (not part of the
+    # model's serialised state) — ephemeral, resets on reconnect (acceptable: a fresh connection
+    # gets a fresh history, and a reconnected battery starts a new freeze-detection cycle).
+    _block_content_hashes: dict = PrivateAttr(default_factory=dict)
+
     # Direct-inverter register caches injected by add_direct_source() for multi-Client
     # reconciliation (#106 Phase 3). Stored separately from register_caches to avoid the
     # Modbus address collision (both EMS controller and direct inverter live at 0x11).
@@ -563,6 +576,7 @@ class Plant(GivEnergyBaseModel):
             incoming = {HR(k): v for k, v in pdu.to_dict().items()}
             if self._commit_bank(device_address, incoming, pdu.register_count):
                 self._stamp_block(device_address, "HR", pdu.base_register, pdu.register_count, received_at)
+                self._append_block_hash(device_address, "HR", pdu.base_register, pdu.register_count, incoming)
         elif isinstance(pdu, ReadInputRegistersResponse):
             if pdu.is_suspicious():
                 # Pattern A dongle-side substitution from #78 — known fingerprint of 16 fixed
@@ -571,6 +585,7 @@ class Plant(GivEnergyBaseModel):
             incoming = {IR(k): v for k, v in pdu.to_dict().items()}  # type: ignore[misc]
             if self._commit_bank(device_address, incoming, pdu.register_count):
                 self._stamp_block(device_address, "IR", pdu.base_register, pdu.register_count, received_at)
+                self._append_block_hash(device_address, "IR", pdu.base_register, pdu.register_count, incoming)
         elif isinstance(pdu, WriteHoldingRegisterResponse):
             if pdu.register == 0:
                 _logger.warning(f"Ignoring, likely corrupt: {pdu}")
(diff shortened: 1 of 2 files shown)


dewet22-codex:

givenergy_modbus/model/plant.py
```diff
@@ -692,6 +707,63 @@ def block_age(
...
+
+        n=5 at a 30 s poll interval ≈ 2.5 min before the signal fires — long enough to avoid
+        false positives during normal idle periods, short enough to catch a firmware-update
+        freeze within the first few minutes.
+        """
+        dq = self._block_content_hashes.get((device_address, reg_type, base_register, register_count))
+        if dq is None or len(dq) < n:
+            return False
+        recent = list(dq)[-n:]
+        return all(h == recent[0] for h in recent)
```
dewet22-codex: This signal fires on healthy batteries in the committed capture corpus. I replayed the real IR(60,60) responses and found byte-identical streaks of 23 and 26 consecutive samples in `hybrid_gen1_arm449_0x11_poll_10min.log`, plus streaks of 15 in the 60-minute hybrid capture and 7–8 in the EMS AC captures. With the default `n=5`, those ordinary batteries would return `True` and a downstream consumer following the proposed usage could mark them unavailable. The LV block has no heartbeat/counter field, so consecutive content equality alone cannot distinguish bootloader freeze from a live but unchanged BMS; I think this needs an additional discriminating signal or a threshold derived and validated against the healthy captures before exposing it as `is_battery_frozen()`.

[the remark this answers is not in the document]
dewet22: Done in 6248b81 — is_block_frozen now raises ValueError for n outside 1.._FROZEN_HASH_DEQUE_SIZE, so an unsatisfiable n fails loudly rather than silently never firing.

tests/model/test_plant.py
```diff
@@ -1956,3 +1956,85 @@ def test_plant_capabilities_is_ac_coupled():
...
+def test_hash_deque_capped_at_max_size(plant: Plant):
+    """The internal deque never grows beyond _FROZEN_HASH_DEQUE_SIZE."""
+    from givenergy_modbus.model.plant import _FROZEN_HASH_DEQUE_SIZE
+
+    bank = {110: 0x4358, 111: 0x3232, 112: 0x3331, 113: 0x4734, 114: 0x3832}
+    for i in range(_FROZEN_HASH_DEQUE_SIZE * 2):
+        _feed_bank(plant, {**bank, 60: i})  # vary one register so hashes differ
+    key = (0x32, "IR", 60, 60)
+    dq = plant._block_content_hashes[key]
+    assert len(dq) == _FROZEN_HASH_DEQUE_SIZE
```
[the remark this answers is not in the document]
dewet22: Added test_is_block_frozen_invalid_n_raises in 6248b81 — covers n=0, n>_FROZEN_HASH_DEQUE_SIZE, and the guard through the is_battery_frozen alias.

givenergy_modbus/model/plant.py
dewet22: Confirmed — I replayed the committed corpus and reproduced exactly this. On `hybrid_gen1_arm449_0x11_poll_10min.log` the actively-polled batteries 0x32/0x33 hold byte-identical IR(60,60) content for max streaks of 26 and 23 samples; 15 in the 60-min hybrid, 7-8 in the EMS AC captures. Dongle fan-out re-serving the same cached frame plus genuinely-static telemetry, as you say, and the LV block has no heartbeat field to discriminate on.

I've pulled the public `is_block_frozen` / `is_battery_frozen` predicates in 43e068c and kept only the primitive (`_block_content_hashes` + `_append_block_hash`). That's the right substrate for an eventual duration-based detector validated against more than the single freeze capture I have, and #57's bounds-discard calibration reuses the same store. #91 stays open with these findings recorded. Thanks for catching it before it shipped.

dewet22-codex:

givenergy_modbus/model/plant.py
```diff
@@ -692,6 +709,35 @@ def block_age(
...
+        and battery telemetry genuinely doesn't change every poll). Naive consecutive-equality
+        therefore can't distinguish a bootloader-mode freeze from a live-but-static BMS without a
+        discriminating signal (or a duration threshold validated against more than one freeze
+        capture). The deque is retained because it is the right substrate for that future work,
+        and #57's bounds-discard calibration ("discarded bank matches last accepted vs genuinely
+        new") consumes the same store. See #91.
+        """
+        key = (device_address, reg_type, base_register, register_count)
+        dq = self._block_content_hashes.setdefault(key, deque(maxlen=_FROZEN_HASH_DEQUE_SIZE))
+        dq.append(hash(frozenset(committed.items())))
```
dewet22-codex: The revised analysis identifies duration as the likely discriminator, but this retained primitive discards all timing information. `register_block_updated_at` only holds the latest commit time and is overwritten by every fan-out response, while this bounded deque stores ten bare hashes; once those ten fill, there is no way to tell whether an unchanged run began seconds or an hour ago. That means it is not yet the substrate for the proposed duration-based detector. I think each retained hash needs its commit timestamp, or the plant needs to track an `unchanged_since` value alongside the current hash, before landing this as #91 groundwork.

dewet22: Good catch — you're right that the bare-hash deque retained none of the timing the duration detector needs, and that a bounded deque evicts the run's origin once the run exceeds its size. Replaced it in c6b8ede with an `unchanged_since` tracker: per block I now store `(content_hash, unchanged_since)` where `unchanged_since` is the ingestion time of the first commit in the current byte-identical run. Equal hash keeps the timestamp anchored, a changed/first hash resets it — O(1) and it survives runs of any length. Surfaced read-only as `content_unchanged_seconds()` (a raw duration, no verdict). Added a test feeding a 20-commit unchanged run that reports its true 600 s origin, which the old deque couldn't represent.

PR #223, opened 2026-06-10
docs: security policy + committed audit (2026-06-10)

dewet22:
## What

Stands up a security-reporting channel and anchors the 2026-06-10 security review in the repo.

- **`SECURITY.md`** — private vulnerability reporting via GitHub security advisories, supported-versions table mirroring the branch layout (2.2.x pre-release / 2.1.x stable / 2.0.x support), and the Modbus-TCP threat model (unauthenticated protocol; write-path integrity; share-safe cache exports).
- **`SECURITY-AUDIT-2026-06-10.md`** — the full review verbatim. No critical remote vulnerability; High→Low hardening findings with a built-in 5-batch remediation sequence and a "defences that work" section.

## Why on `main`

`SECURITY.md` is only surfaced by GitHub from the **default branch**, and the audit reflects `main`'s state (reviewed at `936523e` = 2.1.6). After merge I'll cherry-pick both files to `v2.2` so the active dev line carries them too.

## Follow-on (not this PR)

- Enable private vulnerability reporting in repo settings (makes the "Report a vulnerability" button live).
- Open the public **security hardening** tracking issue — a checklist mirroring the audit's sequencing — as the anchor every remediation PR references.
- The fixes themselves land as separate PRs per the sequencing; the write-path batch routes through the register-safety reviewer.

Docs/governance only — no code touched.

### DIFF
--- a/SECURITY-AUDIT-2026-06-10.md
+++ b/SECURITY-AUDIT-2026-06-10.md
@@ -1,0 +1,259 @@
+# Security Audit — givenergy-modbus
+
+**Date:** 2026-06-10
+**Scope:** full codebase at commit `936523e` (chore: release 2.1.6)
+**Method:** four parallel security-focused reviews (network input parsing; client & write path;
+model layer & redaction; scripts/CI/supply chain), with the highest-impact claims re-verified
+directly against source before inclusion.
+
+**Threat model:** Modbus TCP has no authentication or TLS. The remote peer (inverter, or any
+device/attacker on the LAN segment) can send arbitrary bytes. The library controls real
+power-grid hardware, so write-path integrity matters as much as parse robustness. Cache
+exports are intended to be share-safe (`redact_serials()`, #212/#214).
+
+**Overall assessment:** No critical remote vulnerability found. No eval/pickle/dynamic-dispatch
+sinks. `WRITE_SAFE_REGISTERS` is enforced at encode time with no bypass found. Frame sizes are
+capped, fixtures are clean, publishing uses OIDC. The findings below are ranked by priority for
+remediation planning.
+
+Severity scale: **High** = fix soon, real attacker value or undercuts a stated guarantee.
+**Medium** = should fix, requires more attacker capability or has bounded impact.
+**Low** = hygiene/hardening.
+
+---
+
+## High priority
+
+### H1 — CRC mismatches on register responses accepted, logged only at DEBUG
+
+- **Where:** `givenergy_modbus/pdu/read_registers.py:105-122` (`_validate_check_code`)
+- **Status:** verified in source
+- **Detail:** The CRC16/Modbus over the response payload is recomputed and compared to the wire
+  `check` field; on mismatch it logs at `_logger.debug()` and the data is accepted. The docstring
+  documents this as deliberate ("incoming inverter frames are the source of truth") — reasonable
+  for tolerating firmware quirks, but DEBUG level gives operators zero visibility if responses
+  are being tampered with. The CRC is the *only* integrity signal available on this protocol.
+- **Attack scenario:** An on-path attacker substitutes register values in a
+  `ReadHoldingRegistersResponse` (e.g. charge-enable, SOC reserve). The forged values populate
+  the plant cache silently; downstream consumers (e.g. Home Assistant) act on them.
+- **Fix plan:**
+  - [ ] Minimum: raise the mismatch log to `WARNING`, with a message noting possible tampering.
+  - [ ] Better: add a configurable strict mode that raises `InvalidPduState` on mismatch
+        (lenient default preserves current firmware tolerance).
+
+### H2 — `redact_serials()` gaps: meter serial, Plant header fields, untested AIO byte-swap path
+
+- **Where:** `givenergy_modbus/model/register_cache.py:25-72,143-178`;
+  `givenergy_modbus/model/meter.py:89`; `givenergy_modbus/model/plant.py` (serial fields,
+  populated in `Plant.update()`)
+- **Status:** verified in source
+- **Detail:** Three gaps in the share-safe-export guarantee from #212/#214:
+  1. **Meter product serial not redacted.** `MeterProductRegisterGetter` declares
+     `"serial_number": Def(C.string, None, MR(60), MR(61))`. `_get_serial_groups()` walks
+     `(inverter, battery, ems, gateway)` — no `meter` — and filters on
+     `pre_conv is Converter.serial`, which `C.string` fails. `redact_serials()`'s `_reg_cls`
+     maps only `"HR"`/`"IR"`, so `"MR"` groups are skipped regardless. A redacted export
+     containing MR(60–61) leaks the meter identifier untouched.
+  2. **Plant-level serial fields out of scope.** `Plant.inverter_serial_number` and
+     `Plant.data_adapter_serial_number` are populated from PDU headers and live outside any
+     `RegisterCache`. Redacting every cache then sharing a `Plant` dump still leaks both.
+  3. **AIO byte-swapped HR(8–12) redaction is untested.** This group was explicitly added
+     because AIO firmware stores the unit serial byte-swapped (recoverable). If the
+     byte-swapped string fails both patterns in `Converter.redact_serial()`, the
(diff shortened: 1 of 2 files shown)


dewet22-codex:

SECURITY.md
```diff
@@ -0,0 +1,58 @@
...
+| 2.0.x | Support branch — security/critical fixes only |
+| < 2.0 | Unsupported |
+
+## Scope and threat model
+
+This library speaks Modbus TCP to GivEnergy inverters. A few properties shape
+what counts as a vulnerability:
+
+- **The protocol has no authentication or transport encryption.** Any device or
+  process on the same LAN segment can send arbitrary bytes to a listening client,
```
dewet22-codex: The client does not listen for inbound LAN connections; `Client.connect()` opens an outbound TCP stream to the configured inverter and reads from that established peer. An arbitrary same-LAN process cannot simply send bytes to it unless it controls the configured peer, can become on-path/spoof the connection, or the application deliberately connects to it. I think this should describe hostile input as coming from a compromised/malicious configured peer or an on-path attacker, otherwise the policy materially overstates the network exposure.

SECURITY-AUDIT-2026-06-10.md
```diff
@@ -0,0 +1,259 @@
...
+- **Detail:** The CRC16/Modbus over the response payload is recomputed and compared to the wire
+  `check` field; on mismatch it logs at `_logger.debug()` and the data is accepted. The docstring
+  documents this as deliberate ("incoming inverter frames are the source of truth") — reasonable
+  for tolerating firmware quirks, but DEBUG level gives operators zero visibility if responses
+  are being tampered with. The CRC is the *only* integrity signal available on this protocol.
+- **Attack scenario:** An on-path attacker substitutes register values in a
+  `ReadHoldingRegistersResponse` (e.g. charge-enable, SOC reserve). The forged values populate
+  the plant cache silently; downstream consumers (e.g. Home Assistant) act on them.
+- **Fix plan:**
+  - [ ] Minimum: raise the mismatch log to `WARNING`, with a message noting possible tampering.
```
dewet22-codex: CRC is an accidental-corruption check, not an integrity signal against the on-path attacker used in this scenario: after substituting register values, that attacker can recompute the unauthenticated CRC just as the legitimate peer does. Raising mismatches to WARNING or rejecting them in strict mode is useful robustness hardening, but it does not detect a competent on-path attacker and therefore does not support a High-severity tampering claim. I think this finding needs to be reframed around corrupted/malformed frames rather than possible tampering, with its severity adjusted accordingly.

SECURITY-AUDIT-2026-06-10.md
```diff
@@ -0,0 +1,259 @@
...
+        instead of a hardcoded module list, so future device types can't silently miss
+        coverage (same failure mode as the meter gap).
+
+### H3 — `/tmp/givenergy-coordination` inbox is a prompt-injection vector (operational, not library)
+
+- **Where:** `AGENTS.md` coordination-inbox protocol; `.claude/settings.json` Stop hook
+  (`check-inbox.sh`); live dir `/tmp/givenergy-coordination`
+- **Status:** verified protocol + dir permissions; `check-inbox.sh` contents not inspectable
+  from this worktree (lives under `$CLAUDE_CONFIG_DIR`)
+- **Detail:** `/tmp` is world-writable (sticky bit only prevents deletion of others' files).
```
dewet22-codex: This does not match the verified directory permissions: `/tmp/givenergy-coordination` is currently `drwxr-xr-x` (0755), owned by the user, so another local user cannot create or replace files inside it despite `/tmp` itself being world-writable. A same-UID process can write there, but it can also write to the proposed `~/.local/share/...` location and much of the user's configuration, so moving the directory or changing it to 0700 does not mitigate that attacker model. I don't think this supports a High prompt-injection finding as written; it needs either evidence that the directory is sometimes created writable by other users, or reframing as a broader same-account trust issue with a different mitigation.

SECURITY-AUDIT-2026-06-10.md
```diff
@@ -0,0 +1,259 @@
...
+  1. **Meter product serial not redacted.** `MeterProductRegisterGetter` declares
+     `"serial_number": Def(C.string, None, MR(60), MR(61))`. `_get_serial_groups()` walks
+     `(inverter, battery, ems, gateway)` — no `meter` — and filters on
+     `pre_conv is Converter.serial`, which `C.string` fails. `redact_serials()`'s `_reg_cls`
+     maps only `"HR"`/`"IR"`, so `"MR"` groups are skipped regardless. A redacted export
+     containing MR(60–61) leaks the meter identifier untouched.
+  2. **Plant-level serial fields out of scope.** `Plant.inverter_serial_number` and
+     `Plant.data_adapter_serial_number` are populated from PDU headers and live outside any
+     `RegisterCache`. Redacting every cache then sharing a `Plant` dump still leaks both.
+  3. **AIO byte-swapped HR(8–12) redaction is untested.** This group was explicitly added
```
dewet22-codex: I don't think this supports a High-priority redaction gap as written. At the audited commit, tests/client/test_capture.py::test_frame_redactor_redacts_hr8_serial_register feeds the AIO-style byte-swapped HC2114G047 through HR(8-12) and asserts HC2114G000, while tests/model/test_register_cache.py::test_redact_serials_inverter_hr_group exercises the same explicit HR(8-12) group through RegisterCache.redact_serials(). A combined RegisterCache + byte-swapped case would improve coverage, but both the value shape and the relevant path are already covered, and HC2114G047 matches the same serial pattern as the existing RegisterCache fixture. I would narrow this to that specific missing combination rather than reporting a verified leak risk.

SECURITY.md
dewet22: Right — verified there's no listening socket (asyncio.open_connection only). Reworded in 1046172: hostile input now means a compromised/malicious configured peer or an on-path attacker on that connection.

SECURITY-AUDIT-2026-06-10.md
dewet22: Agreed — an on-path attacker recomputes the unauthenticated CRC, so it only catches corruption/naive splicing. H1 downgraded to Medium in 1046172 with the scenario reframed around corrupted/malformed frames; the WARNING/strict-mode fix plan stands as robustness hardening.

SECURITY-AUDIT-2026-06-10.md
dewet22: Confirmed 0755 user-owned — premise corrected and H3 downgraded to Medium in 1046172. Reframed around the fixed-path pre-creation race + same-UID trust (which relocation can't fix); the fix plan now leads with creation-time ownership checks and treating inbox content as untrusted.

SECURITY-AUDIT-2026-06-10.md
dewet22: Verified both tests — the FrameRedactor case does prove the byte-swapped shape redacts (HC2114G047 → HC2114G000). Narrowed in 094f51d to the missing combination (byte-swapped value through RegisterCache.redact_serials()) rather than a leak risk; H2 stays High on the meter-serial and Plant-header gaps.

dewet22-codex:

SECURITY-AUDIT-2026-06-10.md
```diff
@@ -0,0 +1,282 @@
...
+capped, fixtures are clean, publishing uses OIDC. The findings below are ranked by priority for
+remediation planning.
+
+Severity scale: **High** = fix soon, real attacker value or undercuts a stated guarantee.
+**Medium** = should fix, requires more attacker capability or has bounded impact.
+**Low** = hygiene/hardening.
+
+---
+
+## High priority
```
dewet22-codex, line 29: The amendments now explicitly downgrade H1 and H3 to Medium, but both findings still sit under High priority. That leaves the document ranking contradictory: a reader scanning the priority sections, or copying them into the proposed tracking issue, will still treat them as High. I think H1 and H3 need moving into the Medium section (keeping their IDs stable is fine), or this heading needs restructuring so only H2 remains classified as High.

dewet22: Done in 04946c4 — H1 and H3 moved bodily into the Medium section (IDs kept, banners note the move); the High section now contains only H2, and the header amendment note states that explicitly.

Issue #224, opened 2026-06-10
Security hardening backlog — audit 2026-06-10


dewet22:
Umbrella for remediating the findings in [`SECURITY-AUDIT-2026-06-10.md`](https://github.com/dewet22/givenergy-modbus/blob/main/SECURITY-AUDIT-2026-06-10.md) (committed in #223). Full detail, evidence and fix plans live in the audit; this issue tracks the work. Each batch below is intended as one PR referencing this issue.

Headline: no critical remote vulnerability. After pre-publication review, **H2 (redaction gaps) is the sole High finding**; H1/H3 were downgraded to Medium with premises corrected. The audit also records a substantial "defences that work" list.

## Batch 1 — quick wins (single PR)

- [x] **H1 (min):** CRC-mismatch log raised to WARNING (corruption/malformed-frame visibility) — #225
- [x] **M2:** device serials repr-escaped in `null.py` + heartbeat `__str__` — #225. (`is_valid_serial` already used `%r`; left as-is — its serial is unrecognised-by-pattern, so redaction is a no-op that would lose the diagnostic.)
- [x] **M3:** `hdr_len > 300` cap applied in `FrameRedactor` (recovers instead of stalling/leaking the real frame) — #225
- [x] **L3:** addressed as a comment, not a call — the error-path future is already resolved, so `cancel()` is a no-op (the audit itself rated it symmetry-only) — #225
- [x] **L4:** `frame_sent` timeout now backlog-aware (maxsize × inter-frame delay) with identity-conditional stale-future cleanup — #225
- [x] **L5:** `RegisterCache.json()` docstring warns it emits unredacted serials, points to `redact_serials().json()` — #225

## Batch 2 — redaction integrity (H2, restores the #212/#214 guarantee)

- [x] Redact the meter product serial — explicit `("MR", 60, 2)` group + `MR` mapping; blanked (full or partial), since MR is a distinct namespace — #228
- [x] `Plant.redact()` — redacts caches + the header serials (fail-closed via `redact_serial_strict`) — #228
- [x] Byte-swapped AIO serial regression through `RegisterCache.redact_serials()` HR(8-12) — #228
- [x] Auto-discovering serial groups from all `RegisterGetter` subclasses — split out to #235 (defence-in-depth, not a 2.2.0 blocker)

## Batch 3 — write-path assurance (route through register-safety review)

[...]

PR #225, opened 2026-06-10
fix: security-audit batch 1 quick wins (#224)

dewet22:
First remediation PR from the [security audit](https://github.com/dewet22/givenergy-modbus/blob/main/SECURITY-AUDIT-2026-06-10.md) — Batch 1 "quick wins" from the tracking issue #224. Six low-risk hardening fixes; none touch the write path, so no register-safety review needed. TDD throughout.

## Fixes

| Item | Change | Test |
|---|---|---|
| **H1** | CRC integrity-check failures on register responses → **WARNING** (was DEBUG). Still non-fatal — frames remain the source of truth, and the CRC is unauthenticated so this is corruption visibility, not tamper detection. | `test_crc_mismatch_logs_at_warning` |
| **M2** | repr-escape device serials in log/str output (`null.py`, heartbeat `__str__`) so a latin-1 control char can't forge/split log lines. | `test_null_response_warning_escapes_serial`, `test_heartbeat_str_escapes_serial` |
| **M3** | `FrameRedactor` caps the MBAP length at 300 (mirrors `framer.py`) and recovers — previously an oversized length stalled it and, on flush, **leaked the unredacted real frame** behind it. | `test_frame_redactor_recovers_from_oversized_length_field` |
| **L4** | `frame_sent` safety-net timeout is a fixed 5 s, not `qsize()+1` sampled after `put()` (a drained queue gave a 1 s window that could expire under load). | covered by existing client tests |
| **L5** | `RegisterCache.json()` docstring warns it emits unredacted serials → `redact_serials().json()`. | doc-only |

## Two reasoned deviations from the audit's literal wording

- **`is_valid_serial` (`register.py`) left unchanged.** It already uses `%r` (injection-safe). The audit's "stop logging full serials" is a near-no-op here: the warning fires *because* the serial matches no known pattern, so `redact_serial()` returns it unchanged and the unredacted shape is the whole diagnostic.
- **L3 is a comment, not a `cancel()` call.** In the error-retry path the future is already resolved (`.result()` was just called), so `response_future.cancel()` would be a no-op — the audit itself notes "old future is already done; practical risk is low". Added a comment explaining why, rather than dead code. Also dropped the `client.py:1070` `{message}` → `{message!r}` idea: PDUs define no `__repr__`, so `!r` would emit the useless default object repr, and the `__str__` fixes already make that path serial-safe.

Closes the Batch-1 checkboxes in #224 (the `is_suspicious()` DEBUG-dump sub-item is deferred there — DEBUG-level numeric data, no injection vector).

Full suite + tox py311–py314 + lint + build green.

### DIFF
--- a/givenergy_modbus/client/client.py
+++ b/givenergy_modbus/client/client.py
@@ -166,6 +166,16 @@ class FrameRedactor:
             if len(self._buf) < 6:
                 break  # not enough bytes for the MBAP length field yet
             hdr_len = int.from_bytes(self._buf[4:6], "big")
+            if hdr_len > 300:
+                # A real frame's MBAP length never exceeds ~300 (60-register cap). A larger
+                # value means this marker is a false positive (random bytes that happen to
+                # match) — emit it intact as garbage and resume scanning, rather than buffering
+                # up to ~64 KB for a frame that will never complete. Mirrors framer.py's guard.
+                skip = len(_FRAME_MARKER)
+                garbage, self._buf = self._buf[:skip], self._buf[skip:]
+                _logger.debug("FrameRedactor: false marker (len=0x%04x), %db emitted intact", hdr_len, len(garbage))
+                out += garbage
+                continue
             frame_len = 6 + hdr_len
             if len(self._buf) < frame_len:
                 break  # partial frame — wait for more data
@@ -1195,9 +1205,11 @@ class Client:
                 await asyncio.wait_for(self.tx_queue.put((raw_frame, frame_sent, response_future)), timeout=5.0)
             except TimeoutError as exc:
                 raise TimeoutError("TX queue full — producer task has likely died") from exc
-            await asyncio.wait_for(
-                frame_sent, timeout=self.tx_queue.qsize() + 1
-            )  # this should only happen if the producer task is stuck
+            # Fixed safety-net timeout for the producer to dequeue and send this frame. The
+            # previous `qsize() + 1` was sampled *after* put() returned, so a just-drained queue
+            # yielded a 1 s window that could expire under transient load even with a healthy
+            # producer. A timeout here only fires if the producer task is genuinely stuck.
+            await asyncio.wait_for(frame_sent, timeout=5.0)
             try:
                 await asyncio.wait_for(response_future, timeout=timeout)
             except TimeoutError:
@@ -1218,6 +1230,9 @@ class Client:
             if response.error:
                 _logger.error(f"Received error response, retrying: {response}")
                 tries += 1
+                # Unlike the timeout path above, no response_future.cancel() is needed here:
+                # the future is already resolved (we just called .result()), so cancel() would
+                # be a no-op, and the next attempt overwrites expected_responses[hash] anyway.
                 if tries <= retries and retry_delay > 0:
                     await asyncio.sleep(retry_delay)
                 continue
--- a/givenergy_modbus/client/client.py
+++ b/givenergy_modbus/client/client.py
@@ -106,6 +106,11 @@ _SERIAL_GROUPS: "list[tuple[str, int, int]]" = _build_serial_register_groups()
 # MBAP start marker used by the framer to locate frames within a byte stream.
 _FRAME_MARKER = bytes.fromhex("59590001")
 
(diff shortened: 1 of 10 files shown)


dewet22-codex:

givenergy_modbus/client/client.py
```diff
@@ -1195,9 +1205,11 @@ async def send_request_and_await_response(
...
             except TimeoutError as exc:
                 raise TimeoutError("TX queue full — producer task has likely died") from exc
-            await asyncio.wait_for(
-                frame_sent, timeout=self.tx_queue.qsize() + 1
-            )  # this should only happen if the producer task is stuck
+            # Fixed safety-net timeout for the producer to dequeue and send this frame. The
+            # previous `qsize() + 1` was sampled *after* put() returned, so a just-drained queue
+            # yielded a 1 s window that could expire under transient load even with a healthy
+            # producer. A timeout here only fires if the producer task is genuinely stuck.
+            await asyncio.wait_for(frame_sent, timeout=5.0)
```
dewet22-codex: This can still time out with a healthy producer when the bounded queue is busy. If put() waits for a slot in a full 20-item queue, it then appends this frame at the tail behind roughly 20 messages; the producer deliberately sleeps up to tx_message_wait + tx_jitter (0.35 s by default) after each send, so reaching this frame can legitimately take about 7 seconds plus writer.drain() time. The fixed 5-second window therefore turns ordinary maximum backlog into a request failure. I think the safety-net needs to retain a backlog-derived allowance captured before/at enqueue, or derive a conservative bound from maxsize and the configured inter-frame delay, while still enforcing a sensible minimum.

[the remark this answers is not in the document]
dewet22: Good catch — fixed in dd3721c. On the stuck-producer timeout I now cancel response_future, pop it from expected_responses, and raise a descriptive "Producer task is stuck" error, so no stale future is left registered.

dewet22: You're right — a flat 5s regresses a healthy-but-backlogged producer (20 × ~0.35s ≈ 7s). Reworked in dd3721c to derive the bound from tx_queue.maxsize × (tx_message_wait + tx_jitter) × 1.5 for drain/scheduling headroom, floored at 5s. Conservative and deterministic — no qsize() race.

dewet22-codex:

givenergy_modbus/client/client.py
```diff
@@ -1195,9 +1210,25 @@ async def send_request_and_await_response(
...
+                _FRAME_SENT_MIN_TIMEOUT,
+                self.tx_queue.maxsize * (self.tx_message_wait + self.tx_jitter) * 1.5,
+            )
+            try:
+                await asyncio.wait_for(frame_sent, timeout=frame_sent_timeout)
+            except TimeoutError as exc:
+                # Producer is genuinely stuck. Drop the orphaned future so a late send can't
+                # resolve a stale request, and surface a clear error.
+                response_future.cancel()
+                self.expected_responses.pop(expected_shape_hash, None)
```
dewet22-codex: This unconditional pop can remove a newer caller’s future. Same-shaped requests deliberately replace one another: call B cancels call A’s response_future and installs its own under expected_shape_hash, but if A is still waiting on frame_sent and that wait times out, A reaches this line and deletes B’s mapping. That leaves B unable to receive its response. I think cleanup needs to be identity-conditional, for example only popping when self.expected_responses.get(expected_shape_hash) is response_future (and the queue-put timeout path has the same stale-future cleanup concern).

dewet22: Good catch — fixed in c0892a8. Added a _discard() helper that cancels the future and only deletes the mapping when self.expected_responses.get(expected_shape_hash) is still that future, so A's timeout can't evict B's newer entry. Routed both the put()-timeout and frame_sent-timeout paths through it, and added a regression test (A blocked, B replaces the mapping, A times out → B's future survives).

dewet22-codex:
I’ve started the final re-review of this PR.

dewet22-codex (approved):
I’ve completed the final re-review and have no concerns to add. The backlog-aware timeout and identity-conditional cleanup address the review findings cleanly.

Issue #119, opened 2026-05-29
Inverter addressing is inverted — 0x11 is canonical, not 0x32; remove the 0x11→0x32 rewrite and make addressing model-aware


dewet22:
## Summary

`plant.update()` rewrites every response arriving at `0x11`/`0x00` to be stored under `0x32`, and the inverter/EMS accessors read `register_caches[caps.inverter_address]` with `inverter_address` defaulting to `0x32`. Investigating #105 (AIO can't connect) showed this whole arrangement is inverted relative to the actual protocol, and that it's been masking a latent bug in our EMS support. This issue records the findings and proposes the redesign.

## Evidence

Passive captures (the dongle relaying the GE cloud's polling — all `rx`, zero requests) from three independent installs. Device address → register windows observed:

| Install | `0x11` | `0x31` | `0x32` | peripherals |
|---|---|---|---|---|
| HYBRID_GEN1 | identity only | inverter banks | inverter banks + battery #1 (IR60) | `0x30`, `0x33–0x37` |
| EMS | **everything incl. `IR/HR(2040)` rollup** | — | **absent** | `0x01,0x03,0x07,0x08` meters |
| ALL_IN_ONE | **everything** | — | **absent** | `0x50–0x53,0x70,0xA0` HV BCU |

`0x11`/`0x31`/`0x32` carry the **same** inverter (identical serial, DTC, firmware) on different bus addresses, each exposing a different register window. `0x00` never appears in any capture.

GivTCP's reference map confirms it: inverter address = `0x11` by default, `0x31` only for `AC`/`HYBRID_GEN1`; `0x32` is **battery #1's** address, never the inverter's.

## What this means

- `0x32` is not "the inverter." Our library overloaded it as such; it works for older hybrids only because they mirror the inverter banks onto `0x32` (alongside battery #1's IR block). EMS/AIO expose nothing at `0x32`.
- The `0x11→0x32` rewrite is three tangled things: a dead `0x00` branch (never seen on the wire), a conflation of inverter (`0x11`) with battery #1 (`0x32`), and a passive-harvest workaround from before `0x11` was understood.

## Latent bug it has been masking

[...]

PR #226, opened 2026-06-10
fix(model): only adopt inverter serial from the inverter's own PDU (hass#95)

dewet22:
## Problem

On an AIO, the inverter HA device **merges into a battery module** downstream — the inverter's own serial disappears. Reported by the hass agent validating #192 on AberDino's real 4-module AIO (givenergy-hass#95); blocks hass v1.2.0 stable.

## Root cause

`Plant.update()` set `inverter_serial_number` / `data_adapter_serial_number` from **every** `TransparentResponse`, last-write-wins. The `0x50–0x53` AIO battery-module responses carry their *own* `HX…` serial in the envelope `inverter_serial_number` field, so a poll cycle ending on a module overwrote the real inverter serial.

Confirmed against the `aio_a` fixture — the per-device envelope serials are:

| device | envelope `inverter_serial_number` |
|---|---|
| `0x11` inverter | `CH2414G000` ✅ the real one |
| `0x50–0x53` modules | `HX0000G000` |
| `0x70`/`0xA0` BCU/BMS | `HC0000G000` |

The fixture only *looks* correct because its last frame happens to be the `0x11` response — pure ordering luck. Live hardware ends on a module. The same latent bug affects HV BCU (`0x70`) envelopes on non-AIO HV systems, which is why this lands on `main` rather than v2.2-only.

## Fix

Gate the envelope-serial adoption on the inverter address — `capabilities.inverter_address` when known, else the canonical `{0x11, 0x31}` before `detect()` has run. Modules, batteries, BCU/BMS and meters are all excluded. The model field `plant.inverter.serial_number` (HR13-17 decode) was always correct; this makes the envelope field match it regardless of frame order.

## Tests (TDD, synthetic — the fixture is ordering-masked so a golden replay can't catch it)

- inverter PDU sets the serial, then a `0x50` module PDU with an `HX…` envelope must **not** clobber it
- bootstrap without capabilities: `0x11` sets it, a `0x32` battery PDU does not
- with `capabilities.inverter_address = 0x31`: only `0x31` sets it

Full suite (826) + tox py311–314 + lint + build green; zero fallout. `aio_a` replay still resolves `CH2414G000`.

Fixes givenergy-hass#95 (library side).

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where inverter and data adapter serial numbers could be incorrectly overwritten by other connected devices (such as batteries or stack modules) within the larger system. Serial number assignments are now properly validated and gated by device address, ensuring only the true primary inverter can update these identifiers.

---

### Updates after review

- **Codex/Gemini**: gate **only** `inverter_serial_number` (on the canonical `{0x11, 0x31}` set, which sidesteps a stale pre-#119 `0x32` capability). `data_adapter_serial_number` is the **dongle** serial — identical across every response address (verified) — so it's adopted unconditionally; gating it had regressed peripheral-only streams.

### DIFF
--- a/givenergy_modbus/model/plant.py
+++ b/givenergy_modbus/model/plant.py
@@ -534,8 +534,17 @@ class Plant(GivEnergyBaseModel):
             _logger.debug(f"First time encountering device address 0x{device_address:02x}")
             self.register_caches[device_address] = RegisterCache()
 
-        self.inverter_serial_number = pdu.inverter_serial_number
-        self.data_adapter_serial_number = pdu.data_adapter_serial_number
+        # Only the inverter's own responses carry the plant's identity. Battery (0x32-0x37),
+        # BCU/BMS (0x70+/0xA0) and AIO battery-module (0x50-0x53) responses put their *own*
+        # serial in the envelope inverter_serial_number field, so adopting it from every PDU
+        # lets whichever device is polled last clobber the real inverter serial — which merged
+        # the AIO inverter HA device into a battery module downstream (givenergy-hass#95).
+        # Gate on the inverter address (capabilities, or the canonical 0x11/0x31 before detect).
+        inv_addr = self.capabilities.inverter_address if self.capabilities is not None else None
+        is_inverter_pdu = device_address == inv_addr if inv_addr is not None else device_address in (0x11, 0x31)
+        if is_inverter_pdu:
+            self.inverter_serial_number = pdu.inverter_serial_number
+            self.data_adapter_serial_number = pdu.data_adapter_serial_number
 
         if isinstance(pdu, ReadHoldingRegistersResponse):
             incoming = {HR(k): v for k, v in pdu.to_dict().items()}
--- a/givenergy_modbus/model/plant.py
+++ b/givenergy_modbus/model/plant.py
@@ -461,7 +461,16 @@ class Plant(GivEnergyBaseModel):
 
     register_caches: dict[int, RegisterCache] = {}
     capabilities: PlantCapabilities | None = None
-    inverter_serial_number: str = ""
+    # DEPRECATED (#227): the envelope-derived inverter serial duplicates the
+    # register-authoritative ``Plant.inverter.serial_number`` (HR13-17). Retained for
+    # backward compatibility; consumers should migrate to ``inverter.serial_number``. A
+    # runtime DeprecationWarning + removal will follow once consumers have migrated.
+    inverter_serial_number: str = Field(
+        default="",
+        description="DEPRECATED (#227): prefer Plant.inverter.serial_number (register-authoritative).",
+    )
+    # NOT deprecated: the TCP dongle's serial has no register source — the frame envelope is the
+    # only place it appears — so this remains the canonical accessor for the dongle identity.
     data_adapter_serial_number: str = ""
     # Ingestion timestamps per committed register block, keyed by
     # (device_address, register_type_name, base_register) → last successful commit time (#65).
--- a/givenergy_modbus/model/plant.py
+++ b/givenergy_modbus/model/plant.py
@@ -534,17 +534,21 @@ class Plant(GivEnergyBaseModel):
             _logger.debug(f"First time encountering device address 0x{device_address:02x}")
             self.register_caches[device_address] = RegisterCache()
 
-        # Only the inverter's own responses carry the plant's identity. Battery (0x32-0x37),
-        # BCU/BMS (0x70+/0xA0) and AIO battery-module (0x50-0x53) responses put their *own*
-        # serial in the envelope inverter_serial_number field, so adopting it from every PDU
-        # lets whichever device is polled last clobber the real inverter serial — which merged
-        # the AIO inverter HA device into a battery module downstream (givenergy-hass#95).
-        # Gate on the inverter address (capabilities, or the canonical 0x11/0x31 before detect).
-        inv_addr = self.capabilities.inverter_address if self.capabilities is not None else None
-        is_inverter_pdu = device_address == inv_addr if inv_addr is not None else device_address in (0x11, 0x31)
-        if is_inverter_pdu:
+        # The TCP dongle's serial is identical on every response regardless of the addressed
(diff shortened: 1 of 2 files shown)


dewet22-codex:

givenergy_modbus/model/plant.py
```diff
@@ -534,8 +534,17 @@ def update(self, pdu: ClientIncomingMessage, *, received_at: datetime | None = N
...
+        # BCU/BMS (0x70+/0xA0) and AIO battery-module (0x50-0x53) responses put their *own*
+        # serial in the envelope inverter_serial_number field, so adopting it from every PDU
+        # lets whichever device is polled last clobber the real inverter serial — which merged
+        # the AIO inverter HA device into a battery module downstream (givenergy-hass#95).
+        # Gate on the inverter address (capabilities, or the canonical 0x11/0x31 before detect).
+        inv_addr = self.capabilities.inverter_address if self.capabilities is not None else None
+        is_inverter_pdu = device_address == inv_addr if inv_addr is not None else device_address in (0x11, 0x31)
+        if is_inverter_pdu:
+            self.inverter_serial_number = pdu.inverter_serial_number
+            self.data_adapter_serial_number = pdu.data_adapter_serial_number
```
dewet22-codex: I think only inverter_serial_number should be behind this gate. data_adapter_serial_number identifies the TCP dongle, not the addressed downstream device, and the committed AIO capture confirms that every response address (meters, 0x11, 0x50-0x53, 0x70 and 0xA0) carries the same WJ2414G000 adapter serial while only inverter_serial_number changes between CH/HX/HC. Gating both fields regresses Plant.update() for peripheral-only captures/streams: a valid battery, meter or BCU response used to populate the adapter identity, but now leaves it empty until an inverter-address response happens to arrive. I would continue adopting pdu.data_adapter_serial_number from every accepted TransparentResponse and gate only pdu.inverter_serial_number; the synthetic module test should use the real distinction as well.

dewet22: Good catch — fixed in d9c81c4. Confirmed the dongle serial is identical (WJ2414G000) across every response address in the AIO capture, so data_adapter is now adopted unconditionally and only inverter_serial_number is gated. Added a peripheral-only-stream test asserting a battery PDU still populates the dongle serial but not the inverter serial.

givenergy_modbus/model/plant.py
```diff
@@ -534,8 +534,17 @@ def update(self, pdu: ClientIncomingMessage, *, received_at: datetime | None = N
...
-        self.inverter_serial_number = pdu.inverter_serial_number
-        self.data_adapter_serial_number = pdu.data_adapter_serial_number
+        # Only the inverter's own responses carry the plant's identity. Battery (0x32-0x37),
+        # BCU/BMS (0x70+/0xA0) and AIO battery-module (0x50-0x53) responses put their *own*
+        # serial in the envelope inverter_serial_number field, so adopting it from every PDU
+        # lets whichever device is polled last clobber the real inverter serial — which merged
+        # the AIO inverter HA device into a battery module downstream (givenergy-hass#95).
+        # Gate on the inverter address (capabilities, or the canonical 0x11/0x31 before detect).
+        inv_addr = self.capabilities.inverter_address if self.capabilities is not None else None
+        is_inverter_pdu = device_address == inv_addr if inv_addr is not None else device_address in (0x11, 0x31)
```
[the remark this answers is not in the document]
dewet22: Addressed in d9c81c4 — gated on the canonical {0x11, 0x31} set rather than capabilities.inverter_address (which inverter_address_for only ever yields anyway), so a stale pre-#119 0x32 can never qualify. Added a regression test with inverter_address=0x32 asserting a battery PDU still can't set the inverter serial.

dewet22-codex:

givenergy_modbus/model/plant.py
```diff
@@ -461,7 +461,16 @@ class Plant(GivEnergyBaseModel):
...
     register_caches: dict[int, RegisterCache] = {}
     capabilities: PlantCapabilities | None = None
-    inverter_serial_number: str = ""
+    # DEPRECATED (#227): the envelope-derived inverter serial duplicates the
+    # register-authoritative ``Plant.inverter.serial_number`` (HR13-17). Retained for
+    # backward compatibility; consumers should migrate to ``inverter.serial_number``. A
+    # runtime DeprecationWarning + removal will follow once consumers have migrated.
+    inverter_serial_number: str = Field(
+        default="",
+        description="DEPRECATED (#227): prefer Plant.inverter.serial_number (register-authoritative).",
```
dewet22-codex: I do not think Plant.inverter.serial_number is an equivalent replacement yet. Client.detect() always reads identity at 0x11, so this envelope field is populated immediately; for AC/HYBRID_GEN1, capabilities then point Plant.inverter at 0x31, whose cache is deliberately empty until the first load_config()/refresh() (the Plant.inverter docstring explicitly handles that window). A consumer following this deprecation therefore changes from a valid serial after detect() to an empty value. Bare/offline Plants have a similar mismatch because Plant.inverter still falls back to the legacy 0x32 cache. I would defer the deprecation until the replacement accessor preserves the envelope identity across those lifecycle states, or document the narrower conditions under which it is actually authoritative rather than directing all consumers to migrate now.

dewet22-codex:

givenergy_modbus/model/plant.py
dewet22: Verified and agreed — reverted the deprecation in 31db117. Confirmed Plant.inverter reads 0x31 for AC/HYBRID_GEN1 (empty until first refresh) and the 0x32 battery cache on a bare plant, so it's not lifecycle-equivalent. The envelope field (now correct) is the earliest-available identity, so it stays; #227 is reframed to 'unify into one lifecycle-robust accessor first, then deprecate'. The hass migration steer has been withdrawn.

dewet22-codex (approved):
I’ve completed the re-review and have no concerns to add. The inverter serial is now gated correctly, the dongle serial remains available from peripheral responses, and the lifecycle-aware field documentation is accurate.

PR #228, opened 2026-06-10
fix(model): close redact_serials() gaps — meter serial + Plant header (#224 H2)

dewet22:
Batch 2 of the security audit (#224) — **H2 redaction integrity, the sole remaining High**. Restores the #212/#214 share-safe-export guarantee.

## Fixes

| Item | Change | Test |
|---|---|---|
| **Meter serial leak** | `MR(60-61)` was never redacted — the LUT walk skips the meter module, its Def is `C.string`, and `redact_serials()` only mapped HR/IR. Added an explicit `("MR", 60, 2)` group + mapped `MR`. It's a 2-register identifier that doesn't match the GE 10-char pattern, so it's **blanked** rather than pattern-redacted. | `test_redact_serials_meter_mr_group` |
| **Plant header serials** | `inverter_serial_number` / `data_adapter_serial_number` live on the `Plant` (from the PDU envelope), outside any cache — so redacting caches alone still leaked them. New **`Plant.redact()`** returns a copy with caches redacted *and* both header serials redacted; original untouched. | `test_plant_redact_clears_header_serials_and_redacts_caches` |
| **Byte-swap coverage** | Regression test for a byte-swapped AIO serial (`HC2114G047`) through `RegisterCache.redact_serials()` HR(8-12) — the FrameRedactor path was already covered. | `test_redact_serials_byte_swapped_aio_serial_hr8` |

## One design call worth flagging

The audit assumed the meter serial could be `redact_serial`-pattern-redacted, but it's only **4 chars** (`MR(60-61)`) and matches neither GE pattern (`[A-Z]{2}\d{4}[A-Z]\d{3}` / EMS). So I **zero it** instead — for a share-safe export, blanking a short meter identifier is the conservative choice.

## Deferred follow-ups (noted, tracked against #224)

- **FrameRedactor capture path** doesn't redact `MR` — its `reg_type` is HR/IR only (it types frames by function code), a separate surface from the `redact_serials()` export this PR targets. The audit's H2 fix plan is scoped to `redact_serials()`.
- **Auto-discovery**: both serial-group builders still hardcode `(inverter, battery, ems, gateway)`; walking all `RegisterGetter` subclasses would prevent a future getter being silently missed (as meter was). The audit lists this as "Consider".

Full suite + tox py311–314 + lint + build green.

### DIFF
--- a/givenergy_modbus/model/plant.py
+++ b/givenergy_modbus/model/plant.py
@@ -519,11 +519,15 @@ class Plant(GivEnergyBaseModel):
         """
         from givenergy_modbus.model.register import Converter
 
+        # Fail closed: redact_serial_strict blanks any unrecognised identifier rather than leaking
+        # it verbatim (redact_serial is fail-open). register_block_updated_at is copied so the
+        # redacted snapshot stays independent of later updates to the original.
         return self.model_copy(
             update={
                 "register_caches": {addr: cache.redact_serials() for addr, cache in self.register_caches.items()},
-                "inverter_serial_number": Converter.redact_serial(self.inverter_serial_number) or "",
-                "data_adapter_serial_number": Converter.redact_serial(self.data_adapter_serial_number) or "",
+                "inverter_serial_number": Converter.redact_serial_strict(self.inverter_serial_number),
+                "data_adapter_serial_number": Converter.redact_serial_strict(self.data_adapter_serial_number),
+                "register_block_updated_at": dict(self.register_block_updated_at),
             }
         )
 
--- a/givenergy_modbus/model/plant.py
+++ b/givenergy_modbus/model/plant.py
@@ -509,6 +509,24 @@ class Plant(GivEnergyBaseModel):
             return BcuRegisterGetter
         return None
 
+    def redact(self) -> "Plant":
+        """Return a share-safe copy: every register cache redacted and the header serials cleared.
+
+        ``redact_serials()`` only covers the register caches; ``inverter_serial_number`` and
+        ``data_adapter_serial_number`` live on the Plant itself (populated from the PDU envelope),
+        so a dumped Plant still leaks both unless they're redacted here too. The original is left
+        untouched (#212/#214 share-safe-export guarantee).
+        """
+        from givenergy_modbus.model.register import Converter
+
+        return self.model_copy(
+            update={
+                "register_caches": {addr: cache.redact_serials() for addr, cache in self.register_caches.items()},
+                "inverter_serial_number": Converter.redact_serial(self.inverter_serial_number) or "",
+                "data_adapter_serial_number": Converter.redact_serial(self.data_adapter_serial_number) or "",
+            }
+        )
+
     def update(self, pdu: ClientIncomingMessage, *, received_at: datetime | None = None):
         """Update the Plant state from a PDU message.
 
--- a/givenergy_modbus/model/register_cache.py
+++ b/givenergy_modbus/model/register_cache.py
@@ -67,6 +67,11 @@ def _get_serial_groups() -> "list[tuple[str, int, int]]":
     # (CH… → HC…), recoverable to the real serial, so it must not leak in a shared
     # export. Appended explicitly, like the BMU serials, since no LUT Def carries it.
     groups.append(("HR", 8, 5))
+    # Meter product serial (MR 60-61, FC 0x16). The MeterProductRegisterGetter walk above
+    # doesn't reach it (meter isn't a walked module and its Def is C.string), and it's a short
+    # 2-register identifier that doesn't match the GE serial pattern — so redact_serials() blanks
+    # it rather than pattern-redacting (audit H2). Appended explicitly here.
+    groups.append(("MR", 60, 2))
     _SERIAL_GROUPS = groups
     return _SERIAL_GROUPS
 
@@ -163,7 +168,7 @@ class RegisterCache(defaultdict[Register, int]):
         """
         from givenergy_modbus.model.register import Converter
 
-        _reg_cls: dict[str, type[Register]] = {"HR": HR, "IR": IR}
+        _reg_cls: dict[str, type[Register]] = {"HR": HR, "IR": IR, "MR": MR}
         result = RegisterCache(dict(self))
         for reg_type, base, count in _get_serial_groups():
             reg_cls = _reg_cls.get(reg_type)
@@ -171,6 +176,13 @@ class RegisterCache(defaultdict[Register, int]):
                 continue
             regs = [reg_cls(base + i) for i in range(count)]
             if not all(isinstance(self.get(r), int) for r in regs):
+                continue
+            # Meter identifiers (MR) are short, non-GE-pattern values that redact_serial can't
+            # match — blank them entirely so a shared export doesn't leak the meter identity.
+            # (Zeroing an already-zero group is a no-op, so this stays idempotent.)
+            if reg_type == "MR":
+                for reg in regs:
+                    result[reg] = 0
                 continue
             raw = b"".join((self[r] & 0xFFFF).to_bytes(2, "big") for r in regs)
             serial_str = raw.decode("latin1").replace("\x00", "").upper()
--- a/givenergy_modbus/model/register_cache.py
+++ b/givenergy_modbus/model/register_cache.py
@@ -68,10 +68,13 @@ def _get_serial_groups() -> "list[tuple[str, int, int]]":
     # export. Appended explicitly, like the BMU serials, since no LUT Def carries it.
     groups.append(("HR", 8, 5))
     # Meter product serial (MR 60-61, FC 0x16). The MeterProductRegisterGetter walk above
-    # doesn't reach it (meter isn't a walked module and its Def is C.string), and it's a short
-    # 2-register identifier that doesn't match the GE serial pattern — so redact_serials() blanks
-    # it rather than pattern-redacting (audit H2). Appended explicitly here.
-    groups.append(("MR", 60, 2))
+    # doesn't reach it (meter isn't a walked module and its Def is C.string), so add it
+    # explicitly. It's a short non-GE-pattern identifier; redact_serials() blanks it via the
+    # fail-closed strict redaction (audit H2). Guarded against a future auto-discovery duplicate.
+    mr_key = ("MR", 60, 2)
+    if mr_key not in seen:
+        seen.add(mr_key)
+        groups.append(mr_key)
     _SERIAL_GROUPS = groups
     return _SERIAL_GROUPS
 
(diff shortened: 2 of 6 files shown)


dewet22-codex:

givenergy_modbus/model/plant.py
```diff
@@ -509,6 +509,24 @@ def _getter_for_device_address(self, device_address: int) -> type[RegisterGetter
...
+        ``data_adapter_serial_number`` live on the Plant itself (populated from the PDU envelope),
+        so a dumped Plant still leaks both unless they're redacted here too. The original is left
+        untouched (#212/#214 share-safe-export guarantee).
+        """
+        from givenergy_modbus.model.register import Converter
+
+        return self.model_copy(
+            update={
+                "register_caches": {addr: cache.redact_serials() for addr, cache in self.register_caches.items()},
+                "inverter_serial_number": Converter.redact_serial(self.inverter_serial_number) or "",
```
dewet22-codex: This still leaks an unrecognised header serial unchanged: `Converter.redact_serial()` deliberately returns its input when it does not match either known GE pattern. I can reproduce `Plant(inverter_serial_number="UNKNOWN123", data_adapter_serial_number="ODD-SERIAL").redact()` retaining both values. Since this new method is the share-safe boundary, I think it needs to fail closed here (blank an unchanged/non-matching identifier), with a regression test for that case. The same guarantee is worth considering for cache serial groups too, since `RegisterCache.redact_serials()` currently leaves unrecognised identifier-shaped fields unchanged.

givenergy_modbus/model/plant.py
```diff
@@ -509,6 +509,24 @@ def _getter_for_device_address(self, device_address: int) -> type[RegisterGetter
...
+        ``redact_serials()`` only covers the register caches; ``inverter_serial_number`` and
+        ``data_adapter_serial_number`` live on the Plant itself (populated from the PDU envelope),
+        so a dumped Plant still leaks both unless they're redacted here too. The original is left
+        untouched (#212/#214 share-safe-export guarantee).
+        """
+        from givenergy_modbus.model.register import Converter
+
+        return self.model_copy(
+            update={
+                "register_caches": {addr: cache.redact_serials() for addr, cache in self.register_caches.items()},
```
dewet22-codex, line 527: There is a second fail-open path in the new share-safe helper: `RegisterCache.redact_serials()` explicitly leaves a known serial register group unchanged when its decoded value does not match a recognised pattern. That means a plant containing a new/vendor-specific serial format will still dump that identifier verbatim through `Plant.redact()`. I would treat known identifier locations as sensitive regardless of format for this API, either by making this call use a strict/fail-closed mode or by changing the cache redaction contract, and cover an unrecognised value in a `C.serial` group.

givenergy_modbus/model/plant.py
```diff
@@ -509,6 +509,24 @@ def _getter_for_device_address(self, device_address: int) -> type[RegisterGetter
...
+                "inverter_serial_number": Converter.redact_serial(self.inverter_serial_number) or "",
+                "data_adapter_serial_number": Converter.redact_serial(self.data_adapter_serial_number) or "",
+            }
+        )
```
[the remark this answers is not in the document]
dewet22: Done in 107d825 — register_block_updated_at is now copied (dict(self.register_block_updated_at)) so the redacted snapshot stays independent of later updates.

givenergy_modbus/model/register_cache.py
```diff
@@ -67,6 +67,11 @@ def _get_serial_groups() -> "list[tuple[str, int, int]]":
     # (CH… → HC…), recoverable to the real serial, so it must not leak in a shared
     # export. Appended explicitly, like the BMU serials, since no LUT Def carries it.
     groups.append(("HR", 8, 5))
+    # Meter product serial (MR 60-61, FC 0x16). The MeterProductRegisterGetter walk above
+    # doesn't reach it (meter isn't a walked module and its Def is C.string), and it's a short
+    # 2-register identifier that doesn't match the GE serial pattern — so redact_serials() blanks
+    # it rather than pattern-redacting (audit H2). Appended explicitly here.
+    groups.append(("MR", 60, 2))
```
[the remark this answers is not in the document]
dewet22: Done in 107d825 — the explicit ('MR', 60, 2) group is now guarded with the seen set, so a future auto-discovery walk reaching the meter getter won't produce a duplicate.

givenergy_modbus/model/plant.py
dewet22: Good catch — fixed in 107d825. Added Converter.redact_serial_strict() (fail-closed: recognised serials are date-redacted, anything else is blanked) and Plant.redact() now uses it, so UNKNOWN123/ODD-SERIAL are blanked. Regression test added.

givenergy_modbus/model/plant.py
dewet22: Agreed — fixed the same fail-open path in RegisterCache.redact_serials() in 107d825: it now uses redact_serial_strict, so a C.serial group holding an unrecognised format is blanked rather than passed through. Updated the old passthrough test to assert blanking + covers an unrecognised value.

dewet22-codex:

givenergy_modbus/model/register_cache.py
```diff
@@ -174,8 +182,10 @@ def redact_serials(self) -> "RegisterCache":
                 continue
             raw = b"".join((self[r] & 0xFFFF).to_bytes(2, "big") for r in regs)
             serial_str = raw.decode("latin1").replace("\x00", "").upper()
-            redacted = Converter.redact_serial(serial_str)
-            if redacted is None or redacted == serial_str:
+            # Fail closed: a recognised serial is date-redacted, anything else (vendor/meter/
+            # unknown format) is blanked — a known serial location must never leak verbatim.
+            redacted = Converter.redact_serial_strict(serial_str)
```
dewet22-codex: The strict path still fails open for a partial serial group: the completeness check immediately above skips the group unchanged if a cache contains only some registers from a known identifier location, and `Plant.redact()` then exports those values as share-safe. Partial caches are a normal shape here, so even a fragment such as the first two registers of HR(13-17) can disclose part of the serial. I think that branch needs to zero whichever registers from the known group are present (without injecting absent registers), while preserving the current absent-group behaviour; a partial-group regression would pin that down. The method docstring also still says partial/unrecognised groups are left unchanged, which no longer matches the intended security contract.

dewet22: Good catch — fixed in cecbfa2. A partial group now has its present registers blanked (no absent registers injected). Done in two passes since serial groups overlap (LV battery IR110-114 and BMU#0 IR114-118 share IR114): full groups redact first and record ownership, then partial fragments are blanked only if not already owned by a full group — so the partial BMU group can't clobber the redacted battery register. Docstring updated; partial-group regression test added.

dewet22-codex:
I’ve started the final re-review of this PR.

givenergy_modbus/model/register_cache.py
```diff
@@ -150,36 +158,53 @@ def redact_serials(self) -> "RegisterCache":
...
+                partial_fragments.extend(present)  # handled in the second pass
                 continue
             raw = b"".join((self[r] & 0xFFFF).to_bytes(2, "big") for r in regs)
             serial_str = raw.decode("latin1").replace("\x00", "").upper()
-            redacted = Converter.redact_serial(serial_str)
-            if redacted is None or redacted == serial_str:
+            covered.update(regs)
+            # Fail closed: a recognised serial is date-redacted, anything else (vendor/meter/
+            # unknown format) is blanked — a known serial location must never leak verbatim.
+            redacted = Converter.redact_serial_strict(serial_str)
```
dewet22-codex: The overlap protection only covers partial groups; a fully-present but irrelevant overlapping group can still overwrite a valid redaction. This occurs in a normal LV battery cache because refresh reads IR(60-119): IR(110-114) contains the battery serial, while the globally-applied BMU group IR(114-118) is also fully present. After the battery serial becomes `CE2231A000`, the non-BMU data at 114-118 is treated as an unrecognised full group and blanked, overwriting shared IR(114); I reproduced the final battery placeholder as `CE2231A0` rather than `CE2231A000`. That conflicts with the documented date-preserving placeholder and can break downstream decoding. I think recognised full groups need priority, with unrecognised full groups blanking only registers not owned by a recognised group, analogous to the second pass for partial fragments. A regression using a complete IR(60-119) battery bank would cover the real cache shape.

dewet22: Verified your CE2231A0 repro, and it pointed at something deeper: the cache redaction can't safely fail closed here. Serial groups are applied to a RegisterCache with no device-type context and they overlap — the globally-applied BMU groups (IR(114-118), …) are real serials only on HV BMU stacks, but on an LV battery IR(114) is the battery serial's last register and **IR(115) is `usb_device_inserted`** (manufacturer-documented, battery.py:76). So blanking an "unrecognised full group" destroys legitimate battery data and corrupts the overlapping serial — and the priority-based variant would still blank IR(115-118).

So in acf7608 I've reverted the cache-group fail-closed: redact_serials() fails **open** for HR/IR (date-redact only pattern-matching serials, leave everything else), which never destroys data it can't positively identify as a serial. The fail-closed guarantee is kept where it's unambiguous and can't overlap non-serial data: the inverter/dongle header serials (Plant.redact → redact_serial_strict) and the meter MR identifier (distinct register namespace, blanked). Added a regression with a full LV battery bank (serial + usb_device_inserted) proving the serial is redacted and IR(115) preserved.

The residual you flagged — a genuinely unrecognised *serial* in an HR/IR location passing through — is real but unfixable at this layer without device context (it's indistinguishable from non-serial data in an overlapping group); the right home for that is the eventual device-type-aware redaction, not blanket blanking.

dewet22-codex:
I’ve started another re-review of this PR.

givenergy_modbus/model/register_cache.py
```diff
@@ -172,11 +186,17 @@ def redact_serials(self) -> "RegisterCache":
             regs = [reg_cls(base + i) for i in range(count)]
             if not all(isinstance(self.get(r), int) for r in regs):
```
dewet22-codex: This completeness check also applies to MR, so a partial meter identifier still passes through unchanged even though the new contract says MR is the unambiguous fail-closed namespace. I reproduced `RegisterCache({MR(60): 0x4142}).redact_serials()` retaining the `AB` fragment. Unlike HR/IR, MR cannot overlap those ordinary-data registers, so I think the MR branch needs to run before the completeness check and zero whichever MR registers are present without injecting absent ones. A partial-MR regression would close the remaining meter leak while preserving the necessary HR/IR fail-open behaviour.

dewet22: Agreed and fixed in 687f0b5 — moved the MR branch ahead of the HR/IR completeness check so a partial meter identifier is blanked too (MR(60): 0x4142 → MR(60) zeroed, MR(61) not injected). This is safe precisely because MR is a distinct namespace that can't overlap ordinary data, unlike the HR/IR fail-open case. Partial-MR regression added.

PR #229, opened 2026-06-10
fix(commands): security-audit batch 3 — write-path assurance (#224)

dewet22:
Batch 3 of the security audit (#224) — write-path assurance. Narrow by design: nothing changes *what* can be written, only local input validation and a drift-pinning test. **Register-safety review: passed** (its one suggestion — `1 if enabled else 0` over `int(enabled)` — adopted; details below).

## Changes

| Item | Change | Test |
|---|---|---|
| **L1** | `WriteHoldingRegisterRequest` rejects `bool` values — bool ⊂ int, so `True`/`False` previously slipped the guard and silently wrote 0/1; a bool reaching a numeric register (e.g. ACTIVE_POWER_RATE) is a caller bug that now fails loudly. The nine boolean command helpers pass `1 if enabled else 0`. | `test_write_request_rejects_bool_value` |
| **L2** | New test pinning `_InverterCommands.WRITE_SAFE_REGISTERS ⊆` the PDU-level enforced set. The mixin set is **kept**, not deleted — it's documented per-model groundwork (deferred to #106/#203) and exactly the shape Pillar A's model-aware write policy builds on; zero refs in givenergy-hass/cli (verified). | `test_mixin_write_safe_registers_is_subset_of_pdu_set` |
| **M1** | **Dropped — documented in the audit** (inline amendment, same style as H1/H3). Write-response correlation is by shape hash (no transaction ID in the GE framing) and the dongle fans out responses to *all* connected TCP clients (#196) — so another actor's write echo for the same register resolves our future. A value mismatch usually means "caught someone else's echo", not "device clamped our write", and a match can mask a real clamp: the signal is unattributable, and a WARNING would misfire routinely during normal cloud+app+hass concurrency. Read-back-after-write is the sound consumer pattern. | n/a (analysis) |

## Register-safety notes

- Wire bytes are **identical** for all in-contract calls (`struct.pack('!H', True) == pack('!H', 1)` — verified in review).
- `1 if enabled else 0` (rather than `int(enabled)`) preserves the previous truthy-clamping semantics (`set_enable_eps(2)` still writes 1) without widening accepted types (no silent `int(1.7)` truncation).
- Decoded responses always carry plain ints from the codec, so the shared `__init__` guard is response-safe.
- IntEnum command values (`ExportPriority`, `BatteryPauseMode`) are unaffected (`isinstance(IntEnum, bool)` is False).

## Consumer note

External callers constructing `WriteHoldingRegisterRequest(reg, True)` **directly** now get `ValueError` (fail-closed; flagged in the Changelog trailer). The public command helpers keep accepting `bool` parameters unchanged.

Full suite (838) + tox py311–314 + lint + build green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced input validation to reject invalid data types (such as booleans and non-numeric values) in numeric parameters, preventing type-related errors and improving robustness.

* **Tests**
  * Added comprehensive test coverage for input validation to ensure only valid numeric types are accepted for command parameters.

* **Documentation**
  * Updated security audit documentation to reflect refined findings.

### DIFF
--- a/givenergy_modbus/client/commands.py
+++ b/givenergy_modbus/client/commands.py
@@ -308,6 +308,10 @@ def set_export_priority(priority: ExportPriority) -> list[TransparentRequest]:
     Determines where surplus energy goes: battery first, grid first, or load first.
     Confirmed writable on Model.AC via direct portal observations (hass#52).
     """
+    # bool subclasses int, so ExportPriority(True) would resolve to GRID_FIRST and pass as an
+    # IntEnum — silently selecting a write mode. Reject it before the enum conversion (audit L1).
+    if isinstance(priority, bool):
+        raise ValueError(f"Export priority must be an ExportPriority, not bool (got {priority!r})")
     try:
         priority = ExportPriority(priority)
     except ValueError as e:
--- a/givenergy_modbus/client/commands.py
+++ b/givenergy_modbus/client/commands.py
@@ -113,15 +113,19 @@ class RegisterMap:
 
 
 def _as_int(val: int, name: str) -> int:
-    """Coerce a numeric command argument to int, rejecting bool (audit L1).
+    """Validate a numeric command argument as an integral number (audit L1).
 
-    bool subclasses int, so a plain ``int(val)`` coercion would silently turn True/False
-    into 1/0 *before* the PDU-level bool guard can see it — e.g. ``set_active_power_rate(True)``
-    writing 1. Numeric helpers route their caller input through this instead.
+    Fails loud on the silent-coercion caller-error class: bool subclasses int (True would
+    write 1 or select slot 1's registers), a non-integral float would truncate (2.9 selecting
+    slot 2), and a string is type confusion. Integral floats (50.0) are unambiguous and
+    accepted. Numeric helpers route caller input through this instead of bare ``int(val)``.
     """
-    if isinstance(val, bool):
-        raise ValueError(f"{name} must be an int, not bool")
-    return int(val)
+    if isinstance(val, bool) or not isinstance(val, int | float):
+        raise ValueError(f"{name} must be a number, not {type(val).__name__}")
+    i = int(val)
+    if i != val:
+        raise ValueError(f"{name} must be integral, got {val!r}")
+    return i
 
 
 @deprecated("Call Client.detect() then load_config()/refresh() instead — see PlantNotDetected.")
@@ -359,6 +363,7 @@ def set_pause_slot(slot: TimeSlot | None) -> list[TransparentRequest]:
 
 def set_smart_load_slot_start(idx: int, t: dt_time | None) -> list[TransparentRequest]:
     """Set the start time of Smart Load slot *idx* (1-based, 1–10), or clear it if t is None."""
+    idx = _as_int(idx, "idx")
     if not (1 <= idx <= 10):
         raise ValueError(f"Smart Load slot index must be 1–10 (got {idx})")
     return _set_slot_endpoint(RegisterMap.SMART_LOAD_SLOT_1_START + (idx - 1) * 2, t)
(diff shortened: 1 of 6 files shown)


dewet22-codex:

givenergy_modbus/pdu/write_registers.py
```diff
@@ -195,7 +195,10 @@ def __init__(self, register: int, value: int, *args, **kwargs):
         if not isinstance(register, int):
             raise ValueError(f"Register type {type(register)} is unacceptable")
         self.register = register
-        if not isinstance(value, int):
+        # bool subclasses int, so it would pass the isinstance check below and silently write
+        # 0/1 — a bool reaching a numeric register (e.g. ACTIVE_POWER_RATE) is a caller bug.
+        # Boolean command helpers pass int(enabled) explicitly (audit L1).
+        if isinstance(value, bool) or not isinstance(value, int):
```
dewet22-codex, line 201: This guard can still be bypassed through the public numeric command helpers because several call `int(...)` before constructing the request. I reproduced `set_active_power_rate(True)`, `set_battery_charge_limit(True)`, `set_battery_discharge_limit(True)`, and both AC-limit helpers all producing a write of `1`; the constructor only sees the coerced plain int. That leaves the specific ACTIVE_POWER_RATE caller-error scenario described here silently accepted through the primary public interface. I think those numeric helpers need to reject `bool` before coercion as part of this hardening, with a representative regression test.

dewet22: Good catch — fixed in 9e6f51e. Added _as_int(val, name) (raises ValueError on bool, else int(val)) and routed all ten coercion sites through it (the five you named plus the SoC reserves, EMS target-SoC and export-power-limit helpers). Parametrised regression test covers the nine public helpers with True and False. Register-safety re-review passed — bounds checks and plain-int written values unchanged, and no bare coercion site remains.

dewet22-codex:

givenergy_modbus/client/commands.py
```diff
@@ -112,6 +112,18 @@ class RegisterMap:
     EMS_EXPORT_POWER_LIMIT = 2071
 
 
+def _as_int(val: int, name: str) -> int:
```
dewet22-codex, line 115: The bool-as-int write-path issue still applies to numeric selector arguments, which choose the register rather than its value. For example, `set_ems_charge_target_soc(True, 50)`, `set_export_slot_start(True, ...)`, and `set_smart_load_slot_start(True, ...)` all pass their `1 <= idx` checks and silently target slot 1. That is the same caller-error class, and on a write API it can select the wrong register. I think the shared index validators (`_resolve_slot_registers`, `_export_slot_registers`, the EMS target helpers, and Smart Load helpers) should reject bool too, ideally by routing through this helper or a sibling validator, with a representative regression.

dewet22: Agreed and fixed in ea6a2b9 — idx is routed through _as_int at all seven chokepoints (_resolve_slot_registers, _export_slot_registers, both Smart Load entry points, and the three EMS target-SoC helpers), with a parametrised regression over 8 representative selectors. The register-safety re-review traced every public idx-taking helper to a guarded resolver. Also tightened _as_int to reject non-integral floats (2.9 would have truncated to slot 2 — same class) and strings, per the safety review's observation; integral floats stay accepted.

dewet22-codex:
I’ve started the final review of this PR.

givenergy_modbus/client/commands.py
```diff
@@ -112,6 +112,22 @@ class RegisterMap:
...
```
dewet22-codex, line 115: One numeric conversion path still launders bool before the PDU guard: `set_export_priority(True)` calls `ExportPriority(True)`, which resolves to `ExportPriority.GRID_FIRST` (`1`) and is then accepted as an `IntEnum`. That silently changes a write mode rather than failing loudly, the same caller-error class this helper now closes for values and selectors. I think `set_export_priority` needs an explicit bool/type guard before the enum conversion, with a regression test.

dewet22: Good catch — fixed in 86e947e. set_export_priority now rejects bool before the ExportPriority() conversion, so set_export_priority(True) raises instead of silently resolving to GRID_FIRST. (set_battery_pause_mode passes the raw value, so the PDU-level bool guard already rejects it — a test documents that.) Register-safety review passed; no register/value/address change.

dewet22-codex:
I’ve started the final sign-off review of this PR.

dewet22-codex (approved):
I’ve reviewed the final changes and have no concerns to add.

Issue #88, opened 2026-05-26
Harden framer against malformed-frame DoS cases


dewet22:
## Summary

Security review found two denial-of-service resilience issues in the custom framer/decoder path. Both are reachable from bytes received over the Modbus TCP connection.

### 1. Malformed short frames can crash the network consumer

`Framer.decode()` currently catches only `InvalidPduState` and `InvalidFrame` around `self.pdu_class.decode_bytes(frame)`. A syntactically plausible frame with a valid MBAP marker/header but too-short transparent payload can raise lower-level decoder exceptions such as `struct.error`. Those exceptions escape the async generator and can terminate `Client._task_network_consumer()`.

Relevant paths:

- `givenergy_modbus/framer.py`: decode exception handling around `decode_bytes()`
- `givenergy_modbus/client/client.py`: `async for message in self.framer.decode(frame)` in `_task_network_consumer()`

Local reproduction during review:

```python
import asyncio
from givenergy_modbus.framer import ClientFramer

async def main():
    f = ClientFramer()
    async for msg in f.decode(bytes.fromhex("5959000100020102") + b"x" * 10):
        print(msg)

asyncio.run(main())
```

Observed result: uncaught decoder error (`struct.error` / unpack buffer error), rather than a yielded `InvalidFrame`-style exception.

Suggested fix: make `BasePDU.decode_bytes()` or `Framer.decode()` wrap expected low-level decode failures (`struct.error`, `IndexError`, maybe `UnicodeDecodeError` if relevant) as `InvalidFrame`, so malformed frames are logged/ignored without killing the consumer.

### 2. Headerless garbage grows the framer buffer without bound

When `HEADER_START_MARKER` is not found, `Framer.decode()` logs and breaks while retaining the entire accumulated buffer. A faulty or malicious peer can continuously send bytes that never include the marker, causing unbounded memory growth.

Relevant path:

- `givenergy_modbus/framer.py`: `frame_start_offset < 0` branch in `Framer.decode()`

Local reproduction during review:

```python
import asyncio
from givenergy_modbus.framer import ClientFramer

async def main():
    f = ClientFramer()
    for _ in range(1000):
        async for _ in f.decode(b"x" * 300):
            pass
    print(len(f._buffer))

asyncio.run(main())
```

Observed result: `300000` bytes retained.

[...]

PR #230, opened 2026-06-10
fix: security-audit batch 4 — adversarial-input & decode-boundary hardening (#224)

dewet22:
Fourth remediation batch from the security audit (#224), following #225, #228, #229. This is the robustness pass: nothing changes what can be written or which registers exist — it makes the parse/convert layer degrade gracefully instead of crashing a consumer fed a tampered cache JSON or a malformed wire frame.

### M4 — adversarial register values must not crash consumers
- `Converter.datetime` / `Converter.timeslot` wrap their construction in `try/except ValueError → None`, matching the existing enum/lookup posture (a device-supplied `month=0` / `hour=24` or an out-of-range timeslot degrades to unset rather than raising deep in a consumer).
- `register_object_hook` coerces values with `int(v)` and skips (with a warning) on failure, so a string value in a tampered cache JSON can't propagate until it blows up downstream.

### L6 — decode-boundary tidy-ups
- **`base.decode_bytes`**: restored the broad `except Exception → InvalidFrame` wrap so the decode boundary is complete for *every* caller (not just the framer's outer catch). `InvalidPduState` still propagates first.
- **`NullResponse`**: a short null frame (< 126b) now raises `InvalidFrame` instead of dying in `struct.error` mid-loop.
- **`Framer._buffer`**: moved off the class into `__init__` (per-instance state, no shared-default reliance on `bytes` immutability).
- **`is_lan_config`**: additionally requires the 6 preceding pad bytes to be zero, so a crafted padding field can't false-positive the LAN-config discriminator and drop a valid response (silent timeout). Verified against the real LAN-config fixture.
- **`set_system_date_time`**: rejects pre-2000 years with a clear `ValueError` instead of a confusing encode-time `InvalidPduState`.

### H1 (better) — opt-in strict CRC mode
`ReadRegistersResponse.strict_crc` (`ClassVar`, default `False`) makes a CRC mismatch raise `InvalidPduState`. The lenient default keeps the existing WARNING-and-accept behaviour shipped in #225 — firmware-quirk tolerance is preserved unless a consumer opts in.

### Verification
- TDD throughout; `tox` green across py311–py314 (lint, format, build).
- Register-safety review passed on the `commands.py` change (the date guard adds no register/bounds semantics).
- The framer's #88 defence-in-depth broad catch is retained and now covered by a dedicated test (the L6 wrap moved its trigger into `decode_bytes`).

Part of #224. Cherry-pick to `v2.2` to follow.

### DIFF
--- a/givenergy_modbus/model/register_cache.py
+++ b/givenergy_modbus/model/register_cache.py
@@ -114,11 +114,14 @@ class RegisterCache(defaultdict[Register, int]):
                     _logger.warning("Skipping unrecognised register key %r", k)
                     continue
                 try:
-                    ret[lookup[reg](int(idx))] = v
-                except (KeyError, ValueError):
-                    # KeyError: unknown register prefix (e.g. a future namespace
-                    # we don't know about yet). ValueError: idx wasn't an int.
-                    # Either way, skip the entry rather than aborting the load.
+                    ret[lookup[reg](int(idx))] = int(v)
+                except (KeyError, ValueError, TypeError):
+                    # KeyError: unknown register prefix (e.g. a future namespace we don't
+                    # know about yet). ValueError: idx wasn't an int, or the value wasn't
+                    # coercible (a string in a tampered cache JSON). TypeError: value was a
+                    # non-scalar (list/dict). Skip the entry rather than aborting the load,
+                    # and rather than storing a value that crashes a consumer later (M4).
+                    _logger.warning("Skipping unloadable register entry %r=%r", k, v)
                     continue
             return ret
 
--- a/givenergy_modbus/model/register_cache.py
+++ b/givenergy_modbus/model/register_cache.py
@@ -115,12 +115,17 @@ class RegisterCache(defaultdict[Register, int]):
                     continue
                 try:
                     register = lookup[reg](int(idx))
-                    value = int(v)
-                    if value != v or not (0 <= value <= 0xFFFF):
-                        # Fail closed: a register is an unsigned 16-bit word. A fractional
-                        # number (silently truncated by int()) or an out-of-range value would
-                        # later raise OverflowError in to_bytes() deep in a consumer (M4).
-                        raise ValueError(f"register value {v!r} is not an unsigned 16-bit int")
+                    if v is None:
+                        # None is the codebase's legitimate "unset" sentinel (e.g. a missing
+                        # slot endpoint); preserve it so it round-trips through JSON.
+                        value = None
+                    else:
+                        value = int(v)
+                        if value != v or not (0 <= value <= 0xFFFF):
+                            # Fail closed: a register is an unsigned 16-bit word. A fractional
+                            # number (silently truncated by int()) or an out-of-range value
+                            # would later raise OverflowError in to_bytes() in a consumer (M4).
+                            raise ValueError(f"register value {v!r} is not an unsigned 16-bit int")
                     ret[register] = value
                 except (KeyError, ValueError, TypeError):
                     # KeyError: unknown register prefix (e.g. a future namespace we don't know
--- a/givenergy_modbus/model/register_cache.py
+++ b/givenergy_modbus/model/register_cache.py
@@ -114,13 +114,20 @@ class RegisterCache(defaultdict[Register, int]):
                     _logger.warning("Skipping unrecognised register key %r", k)
                     continue
                 try:
-                    ret[lookup[reg](int(idx))] = int(v)
+                    register = lookup[reg](int(idx))
+                    value = int(v)
+                    if value != v or not (0 <= value <= 0xFFFF):
+                        # Fail closed: a register is an unsigned 16-bit word. A fractional
+                        # number (silently truncated by int()) or an out-of-range value would
+                        # later raise OverflowError in to_bytes() deep in a consumer (M4).
+                        raise ValueError(f"register value {v!r} is not an unsigned 16-bit int")
+                    ret[register] = value
                 except (KeyError, ValueError, TypeError):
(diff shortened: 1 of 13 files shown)


dewet22-codex:

givenergy_modbus/pdu/base.py
```diff
@@ -79,8 +79,11 @@ def decode_bytes(cls, data: bytes) -> "BasePDU":
             pdu.ensure_valid_state()
         except InvalidPduState:
             raise
-        # except Exception as e:
-        #     raise InvalidFrame(str(e), data)
+        except Exception as e:
```
dewet22-codex, line 88: The broad catch still starts after the MBAP/header fields have been decoded. A truncated direct call such as `ClientIncomingMessage.decode_bytes(b"")` therefore raises the raw `struct.error` from `decode_16bit_uint()` before reaching this block, so the decode boundary is not yet complete for every caller as described. I think the header parsing and decoder lookup need to live inside the same boundary too, with a regression using a frame shorter than the header.

givenergy_modbus/model/register_cache.py
```diff
@@ -114,11 +114,14 @@ def register_object_hook(object_dict: dict[str, int]) -> dict[Register, int]:
                     _logger.warning("Skipping unrecognised register key %r", k)
                     continue
                 try:
-                    ret[lookup[reg](int(idx))] = v
-                except (KeyError, ValueError):
-                    # KeyError: unknown register prefix (e.g. a future namespace
-                    # we don't know about yet). ValueError: idx wasn't an int.
-                    # Either way, skip the entry rather than aborting the load.
+                    ret[lookup[reg](int(idx))] = int(v)
```
dewet22-codex: This still stores adversarial integers outside the unsigned 16-bit register domain. For example, `RegisterCache.from_json('{"HR(1)": -1}')` succeeds, then the ordinary `cache.to_string(HR(1))` path raises `OverflowError` at `to_bytes(2, ...)`; values above `0xffff` behave similarly. Since this change is intended to prevent tampered cache values from crashing consumers later, I think the coerced value also needs a `0 <= value <= 0xffff` check before it is stored. Rejecting fractional JSON numbers rather than silently truncating them would keep the same fail-closed posture.

givenergy_modbus/pdu/base.py
dewet22: Fixed in 2d47890 — the MBAP/header parsing now lives inside the same try, so a sub-header truncation surfaces as InvalidFrame too (the explicit header InvalidFrame checks and InvalidPduState are re-raised verbatim). Added a regression covering b"", b"\x59" and a 3-byte frame.

givenergy_modbus/model/register_cache.py
dewet22: Fixed in 2d47890 — the coerced value is now range-checked (0..0xffff) and fractional numbers are rejected rather than truncated, so a negative / >0xffff / fractional value is skipped instead of OverflowError-ing later in to_bytes(). Fail-closed, matching the rest of M4.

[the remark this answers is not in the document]
dewet22: Good catch — fixed in bd4e17b. None is now special-cased before the int() coercion so it round-trips through JSON unchanged (other non-integer / out-of-range values are still rejected). Added a None round-trip regression test.

givenergy_modbus/pdu/base.py
```diff
@@ -79,8 +79,11 @@ def decode_bytes(cls, data: bytes) -> "BasePDU":
             pdu.ensure_valid_state()
         except InvalidPduState:
             raise
```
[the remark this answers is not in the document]
dewet22: Already covered — the catch re-raises InvalidFrame alongside InvalidPduState (`except (InvalidFrame, InvalidPduState): raise`) before the broad branch, so an InvalidFrame from a short null frame propagates unwrapped without double-nesting.

dewet22-codex:

givenergy_modbus/model/register_cache.py
```diff
@@ -114,11 +114,26 @@ def register_object_hook(object_dict: dict[str, int]) -> dict[Register, int]:
...
-                    # KeyError: unknown register prefix (e.g. a future namespace
-                    # we don't know about yet). ValueError: idx wasn't an int.
-                    # Either way, skip the entry rather than aborting the load.
+                    register = lookup[reg](int(idx))
+                    if v is None:
+                        # None is the codebase's legitimate "unset" sentinel (e.g. a missing
+                        # slot endpoint); preserve it so it round-trips through JSON.
+                        value = None
+                    else:
+                        value = int(v)
```
dewet22-codex, line 127: One adversarial JSON value still escapes this boundary: Python’s `json.loads` accepts the non-standard constants `Infinity` and `-Infinity` by default, and `int(float("inf"))` raises `OverflowError`, which is not in the catch below. As a result, `RegisterCache.from_json('{"HR(1)": Infinity}')` still crashes rather than skipping the tampered entry. I think `OverflowError` needs to be handled here too, with a regression. While touching this check, an explicit bool rejection would also avoid silently accepting JSON `true` as register value `1` (`1 == True` bypasses `value != v`).

dewet22: Fixed in b6fb883 — added OverflowError to the skip set (so a JSON Infinity is dropped rather than crashing int(float("inf"))) and added an explicit bool rejection before the range check, so JSON true/false is skipped instead of being stored as 1/0. Regressions cover Infinity/-Infinity/NaN and true/false.

dewet22-codex:

dewet22-codex (approved):
I’ve reviewed the latest changes and have no concerns to add.

PR #231, opened 2026-06-10
docs(model): confirm IR(42) p_load_demand includes the EPS branch (IR31)

dewet22:
## What

Confirms, with a reproducible analysis, that IR(42) `p_load_demand` (busbar house load) **includes** the EPS branch IR(31) `p_backup` — it is not additional. Pins the finding on the IR(42) Def docstring and adds the analysis harness.

Prompted by a givenergy-cli question (their power-balance views risked double-counting EPS); this confirms the inclusion semantics they had already adopted.

## Method

At the AC busbar, `(IR24 inverter terminal − IR30 grid CT)` is the total load drawn off the busbar. If IR(42) already accounts for the EPS-protected load, that equals IR(42) and the residual `r = (IR24 − IR30) − IR42 ≈ 0`; if EPS were a separate output, `r` would track IR(31).

Across **259 EPS-active samples** on two independent units:

| Unit | n | median \|r\| | median \|r − p_backup\| |
|---|---|---|---|
| HYBRID_GEN1 (0x31) | 166 | 2 W | 287 W |
| AIO (0x11) | 93 | 39 W | 303 W |

`r` stays at zero; subtracting IR(31) throws it off by ~287 W. So EPS is inside IR(42), confirming the existing docstring's "independently sensed" note while explaining that the 68% residual is small noise, not the EPS branch.

## Caveat

HYBRID + AIO only — the EMS/AC captures never have EPS active, so this says nothing about those topologies (not expected to differ, but unproven).

## Reproduce

```
uv run python tests/debug/eps_in_load_demand.py --capture tests/fixtures/captures/hybrid_2_bat_a tests/fixtures/captures/aio_a
```

### DIFF
--- a/givenergy_modbus/model/inverter.py
+++ b/givenergy_modbus/model/inverter.py
@@ -667,7 +667,11 @@ class SinglePhaseInverterRegisterGetter(RegisterGetter):
         "fault_code": Def(C.uint32, (C.hex, 8), IR(39), IR(40)),
         "t_inverter_heatsink": Def(C.deci, None, IR(41), min=-40.0, max=100.0),
         # House load / consumption at the busbar, independently sensed — empirically NOT
-        # a derived IR(24)−IR(30) identity (residual non-zero in 68% of samples).
+        # a derived IR(24)−IR(30) identity (residual non-zero in 68% of samples, though
+        # small and centred on zero). That residual is NOT the EPS branch: across 259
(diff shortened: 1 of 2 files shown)


dewet22-codex:

dewet22-codex (approved):
I’ve reviewed this and have no concerns to add.

PR #232, opened 2026-06-10
docs: shared power-flow state classification spec

dewet22:
## What

A draft spec (`docs/flow-state-spec.md`) for a shared power-flow classification primitive, so the givenergy-cli TUI and the givenergy-hass dashboard stop diverging on "what is the system doing now". **Spec only — no implementation yet.**

## Why a spec, not code

The cli proposed a simple `FlowState` enum + `classify()`. Evaluating hass's actual glance card showed that won't serve both:

- hass composes five *independent* flow booleans into ~13 sentences (it can say "exporting AND charging"), which a single mutually-exclusive enum can't represent;
- hass runs Schmitt-trigger hysteresis (200/80 W);
- hass is **JavaScript** — it can't import a Python lib regardless.

So the canonical primitive is a **stateless decomposition superset** (independent booleans + per-edge magnitudes), from which the cli collapses to its 8-state headline and hass composes its sentences. The Python module becomes the cli's import plus the reference/oracle; hass mirrors the rules in JS. Hysteresis stays frontend-side.

## Status

Draft for sign-off — open questions (units, edge set, hysteresis placement) are at the foot of the doc. Coordination notes sent to the cli and hass agents. Implementation deferred until the boundaries are agreed.

### DIFF
--- a/docs/flow-state-spec.md
+++ b/docs/flow-state-spec.md
@@ -1,0 +1,187 @@
+# Power-flow state — shared classification spec
+
+Status: **cli + hass signed off; ready to implement**. No implementation yet — this
+pins the semantics so the eventual `givenergy_modbus.model.flows` primitive, the cli
+TUI, and the hass dashboard all classify instantaneous power flow the same way. Both
+frontends have confirmed the decisions below; the enum member names settle the agreed
+naming. hass will also call `headline(decompose(...))` coordinator-side to back a
+translated flow-state enum sensor, so the headline member names are user-facing there.
+
+## Why this exists
+
+Two frontends already classify "what is the system doing right now" from the same
+three power readings, and they do it differently:
+
+- **givenergy-cli** (`givenergy_cli/glance.py::flow_status`): one mutually-exclusive
+  headline — 8 states, grid-over-battery precedence, stateless, 50 W idle threshold.
+- **givenergy-hass** (`givenergy-glance` custom element, `feat/dashboard-glance-mode`):
+  five *independent* flow booleans composed into ~13 natural-language sentences, with
+  stateful Schmitt-trigger hysteresis (200 W on / 80 W off). It is JavaScript and
+  cannot import a Python library.
+
+Because hass is JS, the convergence is a **shared specification with a Python
+reference implementation** (used directly by the cli; mirrored in JS by hass), not a
+shared import. A single `FlowState` enum would serve only the cli — it cannot express
+hass's simultaneous "exporting *and* charging" composites. So the canonical primitive
+is the **stateless decomposition** below; each frontend collapses it to its own
+headline, and hysteresis stays a frontend concern.
+
+## Inputs and sign conventions
+
+All three readings come straight off the inverter registers (single-phase shown;
+three-phase exposes the same concepts):
+
+| Input | Register | Sign convention |
+|---|---|---|
+| `pv` | IR(18)+IR(20) `p_pv1`+`p_pv2` | ≥ 0 (generation) |
+| `grid` | IR(30) `p_grid_out` (external CT) | **+ = export, − = import** |
+| `battery` | IR(52) `p_battery` | **+ = discharge, − = charge** |
+
+`load` (IR(42) `p_load_demand`) is not an input to the decomposition — it is derived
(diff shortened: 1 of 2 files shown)


dewet22-codex:

docs/flow-state-spec.md
```diff
@@ -0,0 +1,147 @@
...
+solar_left     = max(0, solar_gen   - solar_to_batt)
+solar_to_grid  = min(solar_left, grid_export)
+batt_to_grid   = max(0, grid_export - solar_to_grid)
+solar_to_home  = max(0, solar_left  - solar_to_grid)
+grid_to_home   = max(0, grid_import - grid_to_batt)
+batt_to_home   = max(0, batt_discharge - batt_to_grid)
+home           = solar_to_home + grid_to_home + batt_to_home   # derived load
+```
+
+This output is **stateless and unfiltered** — no hysteresis, no idle smoothing beyond
```
dewet22-codex, line 89: These residual allocations are not capped by the source magnitudes, so the canonical decomposition can invent power whenever the independently sensed registers do not balance exactly. For example, `pv=0`, `grid=1.0` export and `battery=0.2` discharge produces `batt_to_grid=1.0`, five times the available battery power; similarly, charging beyond `solar_gen + grid_import` makes `grid_to_batt` exceed the measured import. Sensor/node drift is expected here, so this is not merely an impossible-input case. Before this becomes the shared oracle, I think each edge needs to be capped by its remaining source and the spec needs an explicit policy for any unmatched residual.

docs/flow-state-spec.md
```diff
@@ -0,0 +1,147 @@
...
+| Condition | State |
+|---|---|
+| `exporting` and `solar_on` | EXPORTING (solar ahead) |
+| `exporting` and not `solar_on` | EXPORTING (battery-sourced) |
+| `importing` | IMPORTING |
+| `discharging` | DISCHARGING |
+| `charging` and `solar_on` | CHARGING_FROM_PV |
+| `charging` | CHARGING |
+| `solar_on` | PV_COVERING_HOUSE |
+| otherwise | IDLE |
```
dewet22-codex, line 113: This does not actually split exporting by source: any `solar_on` value above idle selects the solar-ahead state, even when the export is entirely or overwhelmingly battery-sourced. For example, `pv=0.06`, `grid=2.0`, `battery=2.0` becomes `EXPORTING` rather than `EXPORTING_BATTERY`. Since the planned enum explicitly distinguishes the source, I think these conditions should use the decomposed export edges (for example `solar_to_grid > idle` versus `batt_to_grid > idle`) and define precedence when both contribute.

docs/flow-state-spec.md
dewet22: Good catch — these are independent sensors, so drift mustn't be allowed to invent power. The decomposition now caps every edge by `min(remaining source, remaining sink)` and surfaces the unmatched slack as explicit `residual_charge` / `residual_export` rather than folding it into an edge. Your `pv=0, grid=1.0, battery=0.2` case now yields `batt_to_grid=0.2` with `residual_export=0.8`. Pushed in a3f6aa4.

docs/flow-state-spec.md
dewet22: Agreed — the split now keys off the dominant export edge (`solar_to_grid` vs `batt_to_grid`), ties to solar, with the same for charging. The states are also renamed symmetrically (`EXPORTING_SOLAR` / `EXPORTING_BATTERY`). Your `pv=0.06, grid=2.0, battery=2.0` case now resolves to `EXPORTING_BATTERY`. Pushed in a3f6aa4.

docs/flow-state-spec.md
```diff
@@ -0,0 +1,147 @@
...
+    IDLE = "idle"
+
+@dataclass(frozen=True)
+class FlowDecomposition:
+    solar_on: bool; exporting: bool; importing: bool
+    charging: bool; discharging: bool
+    solar_to_home: float; solar_to_grid: float; solar_to_batt: float
+    grid_to_home: float; grid_to_batt: float
+    batt_to_home: float; batt_to_grid: float
+    home: float
```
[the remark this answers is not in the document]
dewet22: Done — one field per line now, plus the two new residual fields. Thanks.

PR #234, opened 2026-06-10
feat(plant): add unified inverter_serial accessor (#227)

### LINKED ISSUE #227: Unify inverter-serial access (then deprecate the envelope field)
## Background

`Plant.inverter_serial_number` (PDU-envelope) and `Plant.inverter.serial_number` (registers, HR13-17) both expose the inverter serial. They look redundant, but they are **not equivalent across the plant lifecycle**, so neither is a clean drop-in for the other:

- **`inverter_serial_number`** is populated immediately at `detect()` (which reads identity at `0x11`) and — after the clobber fix in #226 — stays correct. Earliest-available, lifecycle-robust.
- **`inverter.serial_number`** reads `capabilities.inverter_address`. For AC/HYBRID_GEN1 that's `0x31`, whose cache is **empty until the first `refresh()`/`load_config()`** (the `Plant.inverter` docstring handles that window); on a bare/offline plant it reads the `0x32` *battery* cache. So it returns empty / wrong-source in those windows.

(`data_adapter_serial_number` is out of scope — the TCP dongle has no register source, the envelope is its only home.)

## Why this was reopened from "deprecate the envelope field"

The original plan was to deprecate `inverter_serial_number` in favour of `inverter.serial_number`. Codex review on #226 showed that would regress consumers from a valid serial to empty in the detect→refresh window. So the envelope field is **not** deprecation-worthy as things stand.

## Actual task: a single authoritative accessor first

Make one accessor that preserves the **earliest-available** inverter identity across all lifecycle states — e.g. `inverter.serial_number` (or a new `Plant.inverter_serial`) falling back to the `0x11` cache when `capabilities.inverter_address` (0x31) is not yet populated, and not reading the `0x32` battery cache on a bare plant. Once a single accessor is authoritative across the lifecycle, *then* the envelope field can be deprecated and removed.

Until then, consumers should keep using `Plant.inverter_serial_number` (now correct).

dewet22:
Closes #227.

`Plant.inverter_serial_number` (PDU envelope) and `Plant.inverter.serial_number` (registers HR13-17) both expose the inverter serial, but neither is correct across the whole plant lifecycle:
- the **envelope** is empty until `detect()` (then stays correct after the #226 clobber fix);
- the **register** accessor reads `capabilities.inverter_address` — `0x31` for AC/HYBRID_GEN1, which is **empty in the detect→first-refresh window** — and reads the **`0x32` battery cache** on a bare plant.

### What this adds
A single authoritative accessor, **`Plant.inverter_serial`**, resolving the earliest-available identity in order:
1. HR13-17 in the capability-selected inverter cache (`0x31`/`0x11`);
2. HR13-17 in the **`0x11` cache** — `detect()`'s `HR(0,60)` identity read lands here for *every* model, so this covers the detect→refresh window before `0x31` is populated;
3. the **`inverter_serial_number` envelope field** — populated at `detect()` and the only home on persisted/bare state.

It **never reads the `0x32` battery cache**, so a bare or pre-detect plant can't surface a battery pack's serial as the inverter's. It reads via `.get()` so it never mutates the (defaultdict) caches. A stale pre-#119 `inverter_address=0x32` capability is ignored (only `0x11`/`0x31` qualify, matching the hass#95 gate).

This is the prerequisite the issue asks for: once consumers migrate to `Plant.inverter_serial`, the envelope field can be deprecated (tracked separately — not in this PR).

### Verification
- TDD: 5 new tests covering each tier (capability cache, 0x11 fallback in the detect window, never-0x32, envelope fallback, register-over-envelope precedence); the 4 existing `inverter_serial_number` envelope-gating tests still pass.
- `tox` green py311–314 (incl. mypy). Read-only accessor — no register-write code, no register-safety gate.

### DIFF
--- a/givenergy_modbus/model/plant.py
+++ b/givenergy_modbus/model/plant.py
@@ -20,7 +20,7 @@ from givenergy_modbus.model.inverter import (
 )
 from givenergy_modbus.model.inverter_threephase import ThreePhaseInverter, select_inverter
 from givenergy_modbus.model.meter import Meter, MeterRegisterGetter
-from givenergy_modbus.model.register import HR, IR, Converter, RegisterGetter
+from givenergy_modbus.model.register import HR, IR, Converter, RegisterGetter, is_valid_serial
 from givenergy_modbus.model.register_cache import RegisterCache
 from givenergy_modbus.pdu import (
     ClientIncomingMessage,
@@ -748,6 +748,10 @@ class Plant(GivEnergyBaseModel):
         3. the ``inverter_serial_number`` envelope field — populated at ``detect()`` and the only
            home on a persisted/bare plant carrying no register caches.
 
+        A register block is only accepted if it decodes to a valid serial (``is_valid_serial`` —
+        the same coherence gate ``_commit_bank`` ingestion uses), so a malformed/partial block in a
+        restored or tampered cache falls through to the envelope rather than outranking it.
+
         Deliberately never reads the 0x32 battery cache, so a bare or pre-detect plant can't
         surface a battery pack's serial as the inverter's. Reads via ``.get()`` so it never
         mutates the (defaultdict) caches. Once consumers move to this accessor, the envelope
@@ -766,7 +770,12 @@ class Plant(GivEnergyBaseModel):
             if any(v is None for v in raw):
                 continue  # fail closed on a partially-present serial block
             serial = Converter.serial(*cast("list[int]", raw))
-            if serial:
+            # Coherence gate (same as _commit_bank ingestion): a complete block can still
+            # decode to garbage — an interior zero register strips to a short string
+            # (SA12\x00\x00G047 -> "SA12G047"), and a persisted/tampered cache can hold spaces
+            # or other malformed data. is_valid_serial() requires a clean 10-char serial, so
+            # such a block falls through to a known-good envelope serial rather than outranking it.
+            if serial and is_valid_serial(serial):
                 return serial
         return self.inverter_serial_number
 
--- a/givenergy_modbus/model/plant.py
+++ b/givenergy_modbus/model/plant.py
@@ -1,7 +1,7 @@
 import logging
 import warnings
 from datetime import UTC, datetime
-from typing import Any, ClassVar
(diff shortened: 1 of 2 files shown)


dewet22-codex:

givenergy_modbus/model/plant.py
```diff
@@ -735,6 +735,41 @@ def inverter(self) -> SinglePhaseInverter | ThreePhaseInverter:
...
+            addresses.insert(0, self.capabilities.inverter_address)
+        for addr in dict.fromkeys(addresses):  # de-dup, preserve order
+            cache = self.register_caches.get(addr)
+            if cache is None:
+                continue
+            raw = [cache.get(HR(n)) for n in range(13, 18)]  # .get(): never mutate the defaultdict
+            if any(v is None for v in raw):
+                continue  # fail closed on a partially-present serial block
+            serial = Converter.serial(*cast("list[int]", raw))
+            if serial:
```
dewet22-codex: This only rejects a missing register or a completely empty decode. A complete five-register block containing one zero register decodes to a shorter but truthy string (for example, `SA12\x00\x00G047` becomes `SA12G047`), and a persisted or tampered cache can likewise contain ten spaces or other malformed data; either then outranks a known-good envelope serial. Normal `_commit_bank()` ingestion already uses `is_valid_serial()` as the coherence gate, but this accessor is also intended to be robust over restored plant state. I think applying that same validity check here would make the documented fail-closed fallback hold for malformed complete blocks as well, with a regression test proving it falls through to the envelope.

dewet22: Good catch — fixed in 0e644eb. The decode is now gated on is_valid_serial() (the same coherence check _commit_bank uses), so a complete-but-malformed block (an interior zero stripping to a short string, or spaces/garbage from a restored/tampered cache) fails the 10-char validity gate and falls through to the envelope serial instead of outranking it. Added a regression for the interior-zero case.

dewet22-codex:

dewet22-codex (approved):
I’ve re-reviewed the updated implementation and have no further concerns. The serial validity gate and regression test address the fallback issue cleanly.

PR #236, opened 2026-06-10
feat(release): hand-authored release-notes file leads the generated section

dewet22:
Targets `v2.2` directly — this is release-line tooling plus the v2.2.0 narrative, not 2.1-line material.

### The gap
`release.py generate` only collects commits since the last tag, so a final cut after a long rc line ships a near-empty CHANGELOG section and GitHub Release body — the 2.1.0 release description was a single bullet while all the substance sat buried in rc/beta sections.

### The mechanism
When `docs/release-notes/v<version>.md` exists, `generate` now prepends its content to the rendered body — in both the real write and `--preview`. The narrative therefore lands in the CHANGELOG section *and* the GitHub Release description at cut time, while the generator remains the single writer of CHANGELOG.md. Absent file → behaviour unchanged. Covered by three new tests (notes lead the body; preview includes them; no-file path identical to before).

### The payload
`docs/release-notes/v2.2.0.md` — the 2.2.0 narrative: graph-shaped Plant (#106), device-identity fixes (#227, hass#95), per-module AIO batteries (#192), share-safe redaction (#212), the full Mythos/Fable security audit (#223/#224), and migration notes from 2.1 (including the `first_battery_serial_number` removal, #191). Consumer blurbs from the cli and hass agents follow as a separate commit once their replies arrive — their go/no-go on rc6 hardware validation also gates the final cut.

Verified end-to-end: `release.py generate 2.2.0 --preview` renders the narrative followed by the generated entries; tox green py311–314.

### DIFF
--- a/docs/release-notes/v2.2.0.md
+++ b/docs/release-notes/v2.2.0.md
@@ -48,6 +48,29 @@ chain (actions pinned to commit SHAs, least-privilege workflow permissions).
   but the typed-device API is the supported path going forward — new device
   types will appear there first.
 
+### Home Assistant
+
+The Home Assistant integration
+([givenergy-hass](https://github.com/dewet22/givenergy-hass)) ships its 1.2.0
+line on 2.2, and it is the most visible beneficiary of the typed device model.
+All-in-One systems now surface each removable battery module as its own HA
+device — parented to the inverter, identified by the module's serial, with its
+24 per-cell voltages and temperatures as diagnostic sensors — so the
+cell-balance heatmap and pack-health views work at module granularity,
+validated end-to-end on a real four-module unit. The 2.2 identity fixes also
+resolve a long-standing All-in-One collision where the inverter device could
+merge into one of its own battery modules.
+
+Beyond the device model, 1.2.0 adds always-positive Grid Power Import/Export
+sensors that suit the Energy Dashboard's two-sensor grid option directly,
+hardens debug-capture access behind admin tokens and single-capture signed
+links (following a security review of the integration that found no
+exploitable vulnerabilities), and inherits the library's rejection of the
+dongle's transient all-zero register blocks — last-good values are kept rather
+than zeros reaching your sensors. There's no migration on upgrade; AIO owners
+who ran an early 1.2.0 pre-release should remove and re-add the integration
+once to clear stale entities.
+
 ### givenergy-cli
 
 The CLI ([givenergy-cli](https://github.com/dewet22/givenergy-cli)) tracks 2.2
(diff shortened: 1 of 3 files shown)


scripts/release.py
```diff
@@ -322,6 +330,12 @@ def cmd_generate(args) -> None:
     body = _render_body(sections)
     today = date.today().isoformat()
 
+    notes_file = RELEASE_NOTES_DIR / f"v{args.version}.md"
+    if notes_file.is_file():
+        notes = notes_file.read_text(encoding="utf-8").strip()
+        if notes:
+            body = f"{notes}\n\n{body}" if body else f"{notes}\n"
```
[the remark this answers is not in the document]
dewet22: Fixed in 4a6af9a — the version arg is normalised at the top of cmd_generate so tag-style input behaves identically to a bare version (notes-file lookup and header both). I used removeprefix("v") rather than lstrip("v") since lstrip strips repeated leading characters. Regression test added.

dewet22-codex:

docs/release-notes/v2.2.0.md
```diff
@@ -0,0 +1,58 @@
...
+  as typed devices rather than reading raw register caches or model predicates.
+- **Reliable device identity**: serial reconciliation via `Plant.serial_index`
+  (#106), a single authoritative `Plant.inverter_serial` accessor that stays
+  correct from first detect through every refresh (#227), and a fix for
+  peripheral responses clobbering the inverter serial (givenergy-hass#95).
+- **Per-module AIO battery data** exposed as enumerable devices (#192).
+- **Share-safe diagnostics**: `RegisterCache.redact_serials()` and `Plant.redact()`
+  produce exports with serial numbers redacted, for attaching to issues (#212).
+- **Model-aware write commands** for three-phase inverters and EMS plants
+  (#75, #203), routed through per-model safe-register checks.
```
dewet22-codex: This overstates the write-safety implementation. `_ThreePhaseCommands` explicitly says a correct per-model allowlist is deferred because inherited single-phase helpers still expose wrong-register writes, and `_EmsCommands` likewise says its per-model `WRITE_SAFE_REGISTERS` is deferred; both currently rely only on the global PDU allowlist. Since this narrative will become the permanent CHANGELOG and GitHub Release description, I think this should describe model-specific command mixins/routing plus global PDU write-safety enforcement, rather than claiming per-model safe-register checks that do not yet exist.

dewet22: Agreed — fixed in 3482379. The bullet now describes what exists: model-specific mixins routing writes to the correct registers, write safety enforced by the global allowlist at the PDU layer, and the per-model allowlists named explicitly as groundwork not yet enforced. Good catch given this text becomes the permanent release description.

dewet22-codex:

dewet22-codex (approved):
I’ve re-reviewed the updated release tooling and narrative and have no further concerns. The write-safety description now matches the implementation accurately.

dewet22-codex (approved):
I’ve reviewed the final Home Assistant release-note expansion as well and have no further concerns.

Issue #238, opened 2026-06-11
GivEnergy - POSSIBLE issue


KeenOn0808:
Guys, this has been created using ChatGPT because my coding skills are way below what's required.

Quick bit of background.  I'm annoyed at the GivEnergy failure after spending £35,000 on my system that's still under warranty.  So, as someone that's dabbled with MS Access I thought 'what if I can do my own system'.  For the last couple of days, with the help of ChatGPT, I've been exploring various avenues an it insists Reg95 is active.  I got it to generate a report.  I hope you can make more sense of it than I can.

Not sure if you'll be interested or not, but if you have this, you can be the judge of that.  I'm happy to share any data that might help you guys.

Adrian.

LV Battery IR(95) Appears Active Despite Being Marked Unused
Summary

During investigation of LV battery telemetry (device addresses 0x32–0x33), IR(95) was observed returning dynamic, non-zero values despite being marked as unused in BatteryRegisterGetter.

The observed behaviour suggests IR(95) contains live battery telemetry rather than an unused or reserved register.

Code Reference

Current mapping in givenergy_modbus/model/battery.py:

# IR(95) unused
"num_cycles": Def(DT.uint16, None, IR(96)),
Test Environment

Two separate inverter systems were monitored:

G714
G436

Battery devices observed:

0x32
0x33

Polling was performed directly against the battery register block IR(60–119).

Observations
G714

Battery 1 (0x32)

IR95 = 1041
IR95 = 1041
IR95 = 1075
IR95 = 1075
IR95 = 4183

Battery 2 (0x33)

IR95 = -203
IR95 = 491
IR95 = 491
IR95 = 456
IR95 = 2950
G436

Battery 1 (0x32)

IR95 = 957
IR95 = 924
IR95 = -25
IR95 = 3912

Battery 2 (0x33)

IR95 = 734
IR95 = 771
IR95 = 3449
Additional Behaviour

During a forced EV charging session:

Battery discharge increased substantially.
IR(95) values increased significantly.
IR(91) converged to 3600 across all active battery packs.
IR(96) remained stable and behaved consistently with cycle count.

Observed IR(95) values include both positive and negative numbers:

-203
-25
456
491
734
771
924
957
1041
1075
2950
3449
3912
4183

This suggests the register is not a fixed configuration value, serial fragment, firmware version, or other static metadata.

What Can Be Concluded

Evidence supports the following:

IR(95) is populated on real LV battery hardware.
IR(95) changes during operation.
[...]

Issue #185, opened 2026-06-04
audit: review scale discrepancies vs v4.1.6 doc (3ph timing + power/current)


dewet22:
Umbrella issue for scale discrepancies the v4.1.6 doc audit flagged that I don't want to change blind — several need confirmation against hardware, and the doc's unit column has been demonstrably wrong before (its own revision history corrects unit errors).

**Three-phase grid-protection timing (doc 0.02s vs code `centi` = 0.01s → 2×):**
HR1034, HR1053, HR1067, HR1070, HR1075, HR1079. Six registers, identical pattern — smells like the doc unit is a 0.02s tick (50 Hz line period) that the code approximates as centiseconds. Low real-world impact (rarely-read 3ph safety params), but worth settling consistently.

**Other 3ph protection params:**
- HR1051 `f_under_derate_slope`, HR1052 `v_reactive_delay_time` — doc 1%/1s, code `milli` → 1000× off.
- HR1011 `grid_reconnect_slope` — doc 0.1%Pn/min, code raw → 10×.
- HR123 `frequency_load_limit_rate` — doc "10 Times", code raw.

**Three-phase power / current (hardware-confirm):**
- IR1083-1085 (`p_load_ac1-3`), IR1091-1093 (`p_out_ac1-3`) — doc 1W, code `deci` (0.1W).
- IR1140 (`i_battery`) — doc 0.01A, code `deci` (0.1A) → 10×.

**Likely doc imprecision, code probably right (no action unless contradicted):**
- IR8/IR9 PV current ("A" vs 0.1A), IR15 BUS voltage, HR60 Vpv start. The doc corrected IR10 from 0.01A to 0.1A in v4.1.1, so plain "A" on the PV currents is most likely stale doc rather than a code bug.

I'd rather confirm the 3ph items against a real three-phase unit (none to hand) than flip scales on the doc's word alone. Tracking them here so they're not lost. Generated from a local audit script (`scripts/audit_register_doc.py`) that diffs the protocol doc against the register map.

dewet22:
The device-section audit extension (#240) adds two PF entries to the pile, surfaced once bare numeric units ("0.01") parse as scales:

- **Meter IR(80–83) `PowerFator_L1..Total`** — doc 0.01, code `milli` (0.001) → 10×. The EMS capture noted on #48 showed PF "sane throughout" under the current decode, but I haven't re-checked the raw values against both candidate scales; one to settle from a capture rather than the doc's word.
[...]

Issue #184, opened 2026-06-04
feat: map per-direction battery energy totals (IR194-197) + IR198/199/213


dewet22:
The v4.1.5 protocol revision added dedicated per-direction battery energy totals that the library doesn't yet map. Today we only expose the *combined* `e_battery_throughput` (IR6/7), so charge vs discharge can't be separated.

New registers (all confirmed absent from the current LUTs):

| Reg | Doc name | Meaning | Unit |
|-----|----------|---------|------|
| IR194/195 | `dwChg_EAllEE` H/L | Battery charge energy total | 0.1 kWh |
| IR196/197 | `dwDis_EAllEE` H/L | Battery discharge energy total | 0.1 kWh |
| IR198 | `uwCTPower10T` | Grid CT power | 10 W |
| IR199 | `SystemStateBit` | System state bitfield | — |
| IR213 | `g_uwAfciModelVersion` | AFCI module version | — |

The charge/discharge split (IR194-197) is the valuable bit — it's what dewet22/givenergy-home-assistant#52 was probing for on real AC hardware, where the lifetime-totals registers were ambiguous. These are explicit per-direction totals at 0.1 kWh.

Plan: add `e_battery_charge_total` / `e_battery_discharge_total` (uint32 + `deci`, matching the existing totals), plus IR198 CT power and IR199 as a decoded bitfield once I work out the bit layout (the doc lists it as a state word but I haven't mapped the individual bits yet). I'll want to validate the new totals against a unit that's done a few charge/discharge cycles before relying on them.

dewet22:
Connecting this to **givenergy-home-assistant#52** (lifetime battery energy totals).

That thread had an unresolved gap: the Cloud API reports a "Battery Discharge Total" (~12,109 kWh on the reporter's AC units) with no corresponding value in anything the library currently polls. IR6/7 (`e_battery_throughput`) is the combined bidirectional figure (~18,540 kWh) and IR180/181 (the `alt1` total source) read a real 0 on AC hardware.

IR196/197 (`dwDis_EAllEE`, Battery Discharge Energy Total, 0.1 kWh) is the most likely local source — it's a dedicated per-direction lifetime total, and the library predates v4.1.5 so it has never been in the poll range.

A single probe on the AC hardware would confirm it before I map anything:

```
givenergy-cli probe --type ir --base 194 --count 4 --device 49
```

[...]

PR #240, opened 2026-06-11
feat(audit): diff device sections against their getters — close the IR95 blind spot

dewet22:
## Summary

#238 exposed a blind spot in the register-doc audit: `scripts/audit_register_doc.py` diffed only the two inverter sections (4.1.1/4.1.2) against the code's register map, while the device sections — meters (4.2.x), LV BCU (4.4.1.1), LV BMU (4.4.1.2), HV config (4.4.2.1) — were inventoried but never compared to their getters. IR(95) `Im_Avg` sat in the doc with a name, unit, and signedness while the code called it unused; it took field evidence to surface it. The scoping was a code comment, never a tracked TODO.

## Changes

- **Per-getter introspection**: `introspect_code()` additionally returns getter → register → converters, so each device section diffs against just its own getter. (Battery IRs share the IR number space with inverter IRs, so the global diff structurally couldn't see these gaps.)
- **`DEVICE_SECTIONS` wiring + `device_sections` report block**: per-section `in_doc_not_code` (self-describing rows with name/unit/rw), `in_code_not_doc`, and scale checks.
- **Scale parser**: understands milli-prefixed electrical units (mV/mA/mAh → ×0.001 of the modelled base unit) and slash-separated model-dependent alternatives ("0.1A/0.01A" passes if the code matches either). Deliberately narrow — `min`/`ms` and rate units like "0.1% Pn/min" keep their old reading, so the #185 inverter baseline is unchanged (same 29 mismatches before and after).
- **Regenerated inventory** with the new report. Also picks up legitimate drift since generation: HR(8–12) removal from #218.
- **Quirks doc**: records the LV BMU capacity-unit doc error — the doc claims `0.1mAh` for IR84–89/101–102, but fixtures prove the `centi`/Ah decode (raw 19097 → 190.97 Ah, sane for a 9.5 kWh/51 V pack; the doc's unit would give 1.9 Ah).
- **codespell**: skip the inventory JSON — it carries verbatim doc register names, typos and all ("Freqency").

## What the new diff found

- **Meters + HV config**: fully covered (code is a superset of the doc).
- **LV BMU**: IR(105)/IR(106) per-battery energy totals — doc-confirms the unverified fork lead on #184 — and IR(107) `Force_DisChg_Flag` (new).
- **LV BCU**: entire section unmapped, no getter (`udwBmsStatus1/2`, `RequestChgCurr`, `RequestDisChgCurr` at 60–63).

Follow-up issues for the residual gaps to come separately. Refs #48 #184 #238.

## Testing

- `tox` fully green (py311–py314, format, lint, build)
- Audit re-run verified: global `scale_mismatches` unchanged at the 29 #185 tracks; `device_sections` matches the findings above

### DIFF
--- a/scripts/audit_register_doc.py
+++ b/scripts/audit_register_doc.py
@@ -16,6 +16,7 @@ from __future__ import annotations
 
 import argparse
 import json
+import math
 import re
 import sys
 from dataclasses import asdict, dataclass, field
@@ -201,8 +202,13 @@ class CodeRegister:
     getters: set[str] = field(default_factory=set)
 
 
-def introspect_code() -> tuple[dict[tuple[str, int], CodeRegister], set[int]]:
-    """Collect every register the library maps + the WRITE_SAFE address set."""
+def introspect_code() -> tuple[dict[tuple[str, int], CodeRegister], set[int], dict[str, dict[int, set[str]]]]:
+    """Collect every register the library maps + the WRITE_SAFE address set.
+
+    Also returns a per-getter view (getter name -> register index -> converter
+    names) so device sections can be diffed against just their own getter,
+    without inverter mappings at the same index muddying the comparison.
+    """
     sys.path.insert(0, str(Path(__file__).resolve().parent.parent))
     from givenergy_modbus.model.battery import BatteryRegisterGetter
     from givenergy_modbus.model.ems import EmsRegisterGetter
@@ -233,6 +239,7 @@ def introspect_code() -> tuple[dict[tuple[str, int], CodeRegister], set[int]]:
         return getattr(c, "__name__", str(c))
 
     out: dict[tuple[str, int], CodeRegister] = {}
+    by_getter: dict[str, dict[int, set[str]]] = {gname: {} for gname in getters}
     for gname, g in getters.items():
         for attr, defn in g.REGISTER_LUT.items():
             for reg in defn.registers:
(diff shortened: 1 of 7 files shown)


dewet22-codex:

scripts/audit_register_doc.py
```diff
@@ -269,15 +289,17 @@ def conv_name(c) -> str | None:
...
     if not u or u in {"hex", "ascii", "-"}:
         return None
+    # Milli-prefixed electrical units (mV/mA/mAh) measure in thousandths of the
+    # base unit the code models (V/A/Ah), so fold the prefix into the scale.
+    # Deliberately NOT generalised: 'min'/'ms' must keep their existing scale-1
+    # reading so the #185 inverter-section baseline is undisturbed.
+    m = re.match(r"^(\d*\.?\d+)?\s*(mv|mah|ma)\b", u)
+    if m:
+        return (float(m.group(1)) if m.group(1) else 1.0) * 0.001
     m = re.match(r"^(\d*\.?\d+)\s*[a-z%]", u)
```
dewet22-codex, line 304: This still treats a numeric-only unit as unparseable because the regex requires a trailing letter or percentage sign. The meter section contains the four PowerFator_* rows with unit 0.01, while MeterRegisterGetter uses C.milli (0.001); the generated report consequently says scale_mismatches is empty for the meter section. That leaves a real device-section scale discrepancy invisible to the new audit. I think accepting a fully numeric token here (or equivalently in _doc_scales) would make the new coverage include those rows.

dewet22: Good catch — fixed in bc0612e. Bare numeric units now parse as scales, restricted to powers of ten so HR46's "101" (a version cell, not a scale) doesn't false-positive. The regenerated report picks up exactly the rows you predicted: the four meter PF registers (doc 0.01 vs our `milli`), plus IR1068 on the three-phase inverter (doc 0.001 vs raw) — the latter corroborates #209. I'll record the meter PF discrepancy on #185 rather than flip the scale on the doc's word alone.

PR #245, opened 2026-06-11
feat(model): auto-discover identifier registers for redaction across all model modules

### LINKED ISSUE #235: Harden serial-group auto-discovery so new device types can't silently miss redaction coverage
Follow-up to the unchecked batch-2 item in #224, split out so the umbrella can close.

## Context

`RegisterCache.redact_serials()` (the share-safe export guarantee from #212/#214) redacts the serial register groups built by `_get_serial_groups()` in `givenergy_modbus/model/register_cache.py`. That builder already auto-discovers most groups by walking the model `REGISTER_LUT`s for `Converter.serial` fields — but only across a hardcoded module list `(inverter, battery, ems, gateway)`, and only where `pre_conv is Converter.serial`. Three serials still need hand-appending:

- BMU serials (decoded manually, no LUT `Def`)
- the legacy AIO unit serial at HR(8-12) (removed from the LUT in #191, still byte-swap-recoverable so still redacted)
- the meter product serial MR(60-61) — the meter module isn't in the walked list, and its `Def` is `C.string`, not `C.serial` (this gap was found and hand-fixed in #228 / audit H2)

## Risk

The meter gap (H2) is the failure mode: a serial-bearing register on a device type the walk doesn't reach, or one that doesn't use `C.serial`, silently misses redaction and leaks in a shared export. As new device types / register maps are added, the hardcoded module list and the `is Converter.serial` check are easy to miss. Defence-in-depth — not a known current leak; the three above are covered.

## Proposal

- Discover serial groups by walking **all** `RegisterGetter` subclasses generically (e.g. `__subclasses__()` / a registry) rather than a hardcoded `(inverter, battery, ems, gateway)` list, so a new device module can't be forgotten.
- Make "this field is an identifier" explicit and discoverable — a marker on the `Def`, or a dedicated identifier-converter set — so serial-bearing fields that use `C.string` (like the meter MR) are caught without a manual append.
- Keep the manual appends only for genuinely LUT-less serials (BMU, HR 8-12), dedup-guarded (already done for MR).
- Add a test that fails if any `RegisterGetter` subclass exposes a `Converter.serial` (or identifier-marked) field not covered by `_get_serial_groups()` — turns "silently missed" into a CI failure.

Not a release blocker for 2.2.0 (deferred from #224 batch 2); fits the data-trust / hardening direction.

dewet22:
## Summary

Closes #235 (deferred batch-2 item from the #224 security audit).

`_get_serial_groups()` — the group builder behind the `redact_serials()` share-safe export guarantee — walked a hardcoded module tuple `(inverter, battery, ems, gateway)` and only matched `Converter.serial` pre-converters. That structure is exactly what let the meter product serial (MR 60-61, a `C.string` Def in an unwalked module) miss redaction until it was found and hand-fixed in #228 (audit H2). This makes the fix structural rather than enumerated.

## Changes

- **`RegisterDefinition.identifier`** (default `False`): explicit marker for unit-identifying fields whose converter isn't `C.serial`. Serial-converter fields are identifiers implicitly and need no marking.
- **`_get_serial_groups()`** now walks *every* `givenergy_modbus.model` submodule via `pkgutil` (import errors propagate — a broken module fails loudly rather than being skipped), matching `C.serial` **or** `identifier=True`. The manual MR(60,2) append is gone, auto-discovered via the marker; the BMU stride appends and legacy HR(8,5) stay, being genuinely LUT-less (manual decode / removed from the LUT in #191 but still byte-swap-recoverable).
- **`meter.serial_number`** marked `identifier=True`. `factory_code` deliberately not — it's a factory/model code shared across units, recorded in a comment.
- **Tests**: a coverage test walks all model modules independently and fails if any `C.serial`/identifier Def isn't covered by `_get_serial_groups()` — a future serial-bearing device module that misses redaction becomes a CI failure instead of a silent leak. A pinned-floor test locks the known 23-group set (inverter, legacy HR8-12, battery, BMU strides, gateway + both AIO layouts, EMS, meter) against discovery refactors.

No change to `redact_serials()` itself — only how its group list is built. The produced group set is verified identical before/after (23 groups, nothing added or removed).

## Testing

- TDD: coverage test written first, failed (`identifier` absent, meter group only present via the hand-append), passes after
- `tox` fully green: py311–py314, format, lint, build
- Existing `test_redact_serials_*` behavioural suite unchanged and green

### DIFF
--- a/tests/model/test_register_cache.py
+++ b/tests/model/test_register_cache.py
@@ -361,7 +361,7 @@ def test_redact_serials_meter_mr_group():
     so it can't be pattern-redacted — redact_serials() zeroes it instead, so a shared export
     doesn't leak the meter identity.
     """
-    val = "AB12".encode("latin1")  # 4-char meter identifier across MR(60-61)
+    val = b"AB12"  # 4-char meter identifier across MR(60-61)
     registers = {MR(60): int.from_bytes(val[0:2], "big"), MR(61): int.from_bytes(val[2:4], "big")}
 
     result = RegisterCache(registers).redact_serials()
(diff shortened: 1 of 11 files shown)


dewet22-codex:

tests/model/test_register_cache.py
```diff
@@ -405,3 +405,71 @@ def test_redact_serials_blanks_partial_meter_identifier():
...
+        for attr in dir(module):
+            cls = getattr(module, attr)
+            if not isinstance(cls, type):
+                continue
+            lut = getattr(cls, "REGISTER_LUT", None)
+            if not lut:
+                continue
+            for field, defn in lut.items():
+                pre_conv = defn.pre_conv[0] if isinstance(defn.pre_conv, tuple) else defn.pre_conv
+                if (pre_conv is Converter.serial or defn.identifier) and defn.registers:
```
dewet22-codex: This test cannot catch the missing-marker case described in its docstring. It selects fields using the same `pre_conv is Converter.serial or defn.identifier` predicate as production, so a future `serial_number: Def(C.string, ...)` whose author forgets `identifier=True` is skipped by both the builder and this test — exactly the original meter failure mode. For the promised CI guard, I think the test needs an independent signal for likely identifiers (at minimum serial-named LUT fields must use `C.serial` or set `identifier=True`), rather than repeating the discovery predicate.

dewet22-codex (changes requested):
I found one gap in the new CI guard: it repeats the production discovery predicate, so it cannot detect a non-C.serial identifier whose marker was forgotten. I’ve left the concrete failure mode inline.

tests/model/test_register_cache.py
dewet22: Right — the test shared the builder's blind spot. Fixed in 87e31fe: the coverage test now has an independent naming layer — any LUT field whose name contains "serial" must use `C.serial` or carry `identifier=True`, regardless of what the discovery predicate sees. Verified it catches the failure mode by stripping the meter marker locally: the test fails naming the exact field. (The two serial-named PDU header fields aren't LUT entries and are already strict-redacted in `Plant.redacted_dict()`, so they're out of this guard's scope.)

Issue #241, opened 2026-06-11
feat(battery): map remaining doc-listed LV registers — BMU IR(107) + the BCU block (60-63)


dewet22:
The extended register audit (#240) closed the device-section blind spot that hid IR(95) (#238). Two doc-listed LV register groups remain unmapped:

**LV BMU IR(107) `Force_DisChg_Flag`** — listed in §4.4.1.2 with no unit or range. Presumably a forced-discharge indicator (cell-balancing or maintenance?). Mapping is trivial once we know what values it takes on real hardware.

**LV BCU block (§4.4.1.1, registers 60–63)** — the library has no LV BCU model at all:

| Reg | Doc name | Unit |
|-----|----------|------|
| 60 | `udwBmsStatus1` | — (uint32 pair with 61, by the `udw` prefix?) |
| 61 | `udwBmsStatus2` | — |
| 62 | `RequestChgCurr` | 1 A |
| 63 | `RequestDisChgCurr` | 1 A |

The request-current pair is the interesting bit — it's the BMS's charge/discharge current envelope, useful for consumers that want to know what the stack will accept before commanding it. Open questions: the BCU's device address (the doc doesn't say; the BMU packs sit at 0x32+), and whether these read as a distinct device or alias onto one of the pack addresses.

The reporter on #238 has LV hardware and offered data — probing 60–63 at a few candidate addresses (0x30? 0x31? alongside the 0x32/0x33 packs) would settle the addressing question.

Refs #48 #184 #238 #240.

dewet22:
Interpretation note for the incoming probe data (ask posted on #238): a *response* at 0x31 is not by itself evidence of a BCU device — on AC/HYBRID_GEN1 units 0x31 is a confirmed read-alias of the inverter (#189), and the reporter's model isn't established yet. Disambiguate by shape: the doc's BCU block is 4 registers (two status words + requested charge/discharge current, 1 A), while the inverter alias would return its usual bank contents at those addresses. A timeout at 0x30/0x31 rules those addresses out; a response needs the shape check before concluding anything.

dewet22:
The probe data from #238 is in (two independent LV plants, full IR(60–119) sweeps at 0x30–0x33 — thanks again @KeenOn0808), and it part-answers the addressing question:

[...]

PR #251, opened 2026-06-12
feat(battery): map IR(105-107) — LV energy totals + force-discharge flag

dewet22:
Maps the three remaining doc-listed LV BMU registers in the IR(105–109) gap, prompted by the field data Adrian (@KeenOn0808) gathered on #238.

## What

- `e_battery_discharge_total` ← IR(105), `e_battery_charge_total` ← IR(106): lifetime energy totals, 0.1 kWh, unsigned (doc §4.4.1.2 "Battery discharge/charge energy total").
- `force_discharge_flag` ← IR(107) (`Force_DisChg_Flag`): raw uint16 — only 0 has been observed in the field and the doc gives no unit or range, so this follows the IR(115) precedent of exposing the raw value until semantics are known.

## Evidence

Probes from two independent LV plants (2× packs each, BMS firmware 3022) show IR(105)/(106) populated at plausible magnitudes — 6216.5/6127.6 kWh on one plant, 2893.6/2975.1 kWh on the other. The regression test encodes those exact raw values.

Two caveats are recorded in the code comments:

- both packs within a plant report **identical** values, so the counters appear stack-level and mirrored into each pack — summing across packs double-counts;
- these are single uint16s, wrapping at 6553.5 kWh (one observed plant is at ~6216 already), and that plant oddly shows lifetime discharge > charge, which I can't yet explain. The per-direction assignment rests on the doc plus the GivTCP fork's long-shipped identical mapping; Adrian's post-cycle re-probe should confirm it.

Field names deliberately match the GivTCP fork (`givenergy_modbus_async/model/battery.py`) so its modbusv2 migration picks these up with zero rename delta — this closes one of the unverified gaps from the fork-salvage triage.

The LV BCU addressing question (0x30/0x31) stays open on #241 pending control probes.

Refs #238, #241.


dewet22-codex:

dewet22-codex (approved):
I’ve reviewed this and have no concerns. The new mappings follow the register reference and two independent LV field reports, while the mirrored-counter and wraparound caveats are recorded without introducing any misleading plant-level aggregation.

PR #252, opened 2026-06-12
fix(inverter): pf formula + ChargeStatus enum (#209, #222)

### LINKED ISSUE #209: pf_inverter_output_now / power_factor — raw register exposed; conversion formula unverified and unapplied
Both power factor registers are defined with `C.uint16` and no scaling:

```python
"power_factor":           Def(C.uint16, None, HR(52)),  # /10_000 - 1
"pf_inverter_output_now": Def(C.uint16, None, IR(16)),
```

The comment on HR(52) notes a speculative conversion formula (`value / 10_000 - 1`) based on documentation, but it was never applied — and the formula itself was left unverified. IR(16) has no comment at all.

## 24 h data from a live SA2114G047

`pf_inverter_output_now` (IR 16) over 24 hours:

- Normal export operation: sustained cluster around **~10,000**
- Heavy grid-charge cycle (~04:00): dip to **~1,700**
- Occasional reactive bursts: spikes to **~19,000**

## Which formula fits?

| Formula | 10,000 | 1,700 | 19,000 |
|---------|--------|-------|--------|
| `÷ 10,000` | 1.0 ✓ | 0.17 ✓ | **1.9 — impossible** |
| `÷ 10,000 − 1` | 0.0 ✓ | −0.83 ✓ | +0.9 ✓ |

The simpler formula requires dismissing all >10,000 readings as noise, but those readings are structured and sustained — they track charging events and grid conditions rather than looking like glitches. Under `÷ 10,000 − 1` every data point is physically valid.

`value / 10_000 - 1` gives power factor in the standard EE signed convention: deviation from unity, where 0.0 = unity (no reactive component), positive = lagging (inductive), negative = leading (capacitive), range [−1, +1]. The billing-friendly cos(θ) form in [0, 1] is `1 − abs(converted)`.

This is consistent with the speculative comment on HR(52), and the live data provides the empirical confirmation it was waiting for.

## Test fixture evidence

Searching the test suite, the three IR(16) fixture captures are all sub-10,000:

| Raw value | `÷ 10,000` | `÷ 10,000 − 1` |
|-----------|------------|----------------|
| 4,790 | 0.479 ✓ | −0.521 ✓ |
| 9,531 | 0.953 ✓ | −0.047 ✓ |
| 8,160 | 0.816 ✓ | −0.184 ✓ |

Both formulas produce valid results for these values — the fixtures alone can't distinguish between them. The HR(52) fixture has a raw value of `0`, which under `÷ 10,000 − 1` gives −1.0 (fully leading) — likely a placeholder default rather than a real capture.

## Three-phase — needs separate investigation

`power_factor` at IR(1068) uses `C.int16`, yielding values like −100 after two's-complement conversion. Applying `÷ 10,000 − 1` to the post-int16 value gives out-of-range results (e.g. −1.01), so the three-phase register appears to use a different encoding. I'd suggest not touching IR(1068) or HR(1004) until that's verified independently.

## Proposed fix (single-phase only for now)

Add a `pf` converter to `Converter` in `givenergy_modbus/model/register.py`:

```python
@staticmethod
def pf(val: int) -> float | None:
    """Power factor in signed EE convention: (value / 10_000) - 1.
    0.0 = unity; positive = lagging (inductive); negative = leading (capacitive).
    Range [-1, 1]. The billing form cos(theta) in [0, 1] is 1 - abs(pf)."""
    if val is not None:
        return val / 10_000 - 1
```

Add to `_PRECISION_BY_CONVERTER` with `precision: 4`, then apply to:

- `pf_inverter_output_now` IR(16)
- `power_factor` HR(52) (remove the now-redundant comment)

The entity description should document the signed convention so users aren't confused by 0.0 at unity.

**Workaround:** givenergy-hass currently applies `÷ 10,000` (without the `− 1`) in the Analyst dashboard card as a stopgap; that will be corrected once the formula is confirmed here.

### LINKED ISSUE #222: feat(inverter): decode charge_status (IR14) — 0=idle, 2=charge, 3=finishing, 5=discharge
## Background

`charge_status` is read from IR(14) and currently decoded as a raw `uint16` with no enum (`Def(C.uint16, None, IR(14))` in `inverter.py`), so consumers see the bare code with no labels.

## Decode (field evidence + capture corroboration)

| Code | Meaning | Evidence |
|---|---|---|
| 0 | idle | my own system — battery power flat at 0 |
| 2 | charging | committed captures + live — battery power negative (~-2.5 kW) |
| 3 | finishing | my own system — appears near 100% SOC, presumably a protection taper that lowers charge current |
| 5 | discharge | committed captures + live — battery power positive |

Confirmed against the committed fixtures: `hybrid_gen1_arm449` reads `5` throughout a discharge window and `2` throughout a charge window; `ac_arm282` flips `2↔5` as the battery cycles. So `2` and `5` are corroborated by real captures on two model families (Gen1 hybrid + AC); `0` and `3` come from my live dashboard, where the status code tracks the sign of battery power exactly. Codes `1` and `4` haven't been seen anywhere yet.

## Proposal

Follow the `EmsInverterStatus` precedent from #108:

- keep the raw `charge_status` (IR14) untouched;
- add an additive `charge_status_label: ChargeStatus | None`, resolved by a lenient helper that maps only the observed `{0, 2, 3, 5}` and returns `None` for anything else, so an unexpected firmware code can't break the decode;
- `ChargeStatus(int, Enum)`: `IDLE = 0`, `CHARGING = 2`, `FINISHING = 3`, `DISCHARGING = 5`;
- docstring records which codes are capture-verified vs owner-observed and flags `1`/`4` as unconfirmed — feedback on those is what would close this.

Additive and non-breaking, in keeping with the register-provenance direction (Pillar B in the post-v2.2 roadmap).

## Notes

- `FINISHING` is my best read of code 3 (current taper approaching full) — happy to be corrected if anyone observes it in another context.
- Lands after the current 2.2.0 rc soak.

dewet22:
## Summary

- **#209 — pf formula**: `pf_inverter_output_now` (IR(16)) and `power_factor` (HR(52)) now decode with the verified offset-unsigned formula (`raw / 10,000 − 1`) via a new `Converter.pf` method. Under plain `÷ 10,000` the >10,000 readings captured from a live SA2114G047 over 24 h would be impossible (>1.0); under `÷ 10,000 − 1` every data point is physically valid. The new converter sits alongside `pf_signed` (used for meter IR(80-83), which has no offset and was shipped in #250).

- **#222 — ChargeStatus enum**: adds `ChargeStatus(IntEnum)` with four field-verified codes (`IDLE=0 / CHARGING=2 / FINISHING=3 / DISCHARGING=5`) and exposes them as `charge_status_label: ChargeStatus | None` on `SinglePhaseInverter`. Unknown codes decode to `None` rather than raising, following the `EmsInverterStatus` precedent from #108. The raw register code is still accessible as `charge_status_code`; the old `charge_status` property is deprecated with a `DeprecationWarning` and scheduled for removal in 3.x.

## Test plan

- [ ] `uv run pytest tests/ -q` — 969 tests pass
- [ ] `uv run ruff check` / `ruff format` — clean
- [ ] New focused tests: `test_pf_converter_inverter` (unity / charging-dip / reactive-burst cases) and `test_charge_status_enum` (all four known codes + unknown + deprecation warning)
- [ ] Existing whole-dict assertions in `test_from_registers`, `test_from_registers_actual_data`, `test_from_actual` updated for both changes
- [ ] tox matrix (CI will run)

Closes #209
Closes #222

## Release Notes

* **New Features**
  * Added charge status enumeration with code and label fields for clearer inverter charging state identification.

* **Bug Fixes**
  * Corrected power factor and inverter output power factor decoding to display accurate scaled values instead of raw register data.

* **Tests**
  * Added comprehensive test coverage for charge status enumeration and power factor converter functionality.


dewet22-codex:

givenergy_modbus/model/inverter.py
```diff
@@ -624,9 +650,10 @@ class SinglePhaseInverterRegisterGetter(RegisterGetter):
         "i_ac1": Def(C.deci, None, IR(10), min=0.0, max=500.0),
         "e_pv_total": Def(C.uint32, C.deci, IR(11), IR(12)),
         "f_ac1": Def(C.centi, None, IR(13), min=40.0, max=70.0),
-        "charge_status": Def(C.uint16, None, IR(14)),
+        "charge_status_code": Def(C.uint16, None, IR(14)),
+        "charge_status_label": Def(C.uint16, _charge_status_from, IR(14)),
```
dewet22-codex, line 651: This is not additive as proposed in #222: replacing the `charge_status` LUT field removes that key from the Pydantic model, `model_dump()`, generated schema, and constructor input immediately. The deprecated property only preserves direct attribute reads, so downstream consumers serialising the inverter or building one from a `charge_status` payload break before 3.x (and Pydantic may silently ignore the old input). I think `charge_status` needs to remain the raw register-backed field for the deprecation window, with `charge_status_label` added alongside it; a renamed `charge_status_code` can be an alias/property if useful, but removing the established field belongs in the 3.x change.

dewet22: Fixed in e103c57. `charge_status` is back as the raw LUT-backed field in `model_dump()` and the Pydantic constructor; `charge_status_label` is additive alongside it, reading the same IR(14). The deprecated `@property` shim is dropped — it cannot shadow a Pydantic field cleanly; the 3.x removal of `charge_status` is the right vehicle for that.

givenergy_modbus/model/inverter.py
```diff
@@ -438,7 +462,9 @@ class SinglePhaseInverterRegisterGetter(RegisterGetter):
         "enable_reversed_418_meter": Def(C.bool, None, HR(49)),
         "active_power_rate": Def(C.uint16, None, HR(50)),
         "reactive_power_rate": Def(C.uint16, None, HR(51)),
-        "power_factor": Def(C.uint16, None, HR(52)),  # /10_000 - 1
+        # Offset-unsigned PF: (raw / 10,000) − 1. See Converter.pf for the encoding
+        # rationale and its distinction from the meter's pf_signed registers.
+        "power_factor": Def(C.pf, None, HR(52)),
```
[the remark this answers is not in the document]
dewet22: Applied in 9dbdc3f — both `power_factor` (HR(52)) and `pf_inverter_output_now` (IR(16)) now have `min=-1.0, max=1.0`, consistent with the bounds on other scalar fields.

dewet22-codex:
I’ve started re-reviewing the updated PR.

dewet22-codex (approved):
I’ve reviewed the updated head and have no remaining concerns. `charge_status` now remains fully backwards-compatible as the raw Pydantic field while `charge_status_label` is additive, and the bounded PF conversion is supported by the live evidence and regression coverage.

PR #254, opened 2026-06-12
feat(model): non-negative directional power sensors (#205)

### LINKED ISSUE #205: feat: expose non-negative directional power sensors for grid and battery
Currently `grid_power` and `battery_power` are signed scalars (positive = export/discharge, negative = import/charge per the v1.1.3 rename). That convention is clean for a single reading, but several downstream consumers need non-negative directional values:

- **HA distribution cards** (`energy-distribution-card`, `power-flow-card-plus` segment mode) require non-negative inputs; negative values break their segment rendering.
- **HA Energy dashboard** grid sensors must be split into separate import/export sensors to feed the cost/income calculation correctly.
- **Future Coach mode** in givenergy-hass will route energy decisions on the basis of separate charge/discharge power figures rather than a signed composite.

PV string power is already exposed as unsigned (`p_pv1`, `p_pv2`), which makes a natural precedent.

The ask is four additional read-only sensors derived from the existing signed registers:

| Proposed name | Derived from | Semantics |
|---|---|---|
| `grid_import_power` | `grid_power` | W, non-negative; zero when exporting |
| `grid_export_power` | `grid_power` | W, non-negative; zero when importing |
| `battery_charge_power` | `battery_power` | W, non-negative; zero when discharging |
| `battery_discharge_power` | `battery_power` | W, non-negative; zero when charging |

These could be virtual `EntityDescription` entries derived at read time (no new Modbus registers needed), or separate computed properties on `Plant` — the mapping approach is left to the implementer's judgement; either would satisfy the HA integration's needs.

Happy to field questions on the givenergy-hass side of the integration.

dewet22:
## Summary

- Adds four `@computed_field` properties to `SinglePhaseInverter`: `grid_import_power`, `grid_export_power`, `battery_charge_power`, `battery_discharge_power`
- All return `int | None` (watts); zero when the direction is inactive, `None` when the source register is absent
- Derived from existing signed fields (`p_grid_out` IR(30), `p_battery` IR(52)) — no new register reads
- Sign convention: `p_grid_out > 0` = exporting to grid; `p_battery > 0` = discharging

Motivation: HA's Energy dashboard and distribution cards require non-negative sensors that split import/export and charge/discharge into distinct entities.

## Test plan

- [ ] `test_directional_power_sensors` covers export, import, idle, missing-register (None), discharge, and charging scenarios
- [ ] Three existing whole-dict assertions updated with expected values derived from fixture data
- [ ] 971 tests passing, 97% coverage

Closes #205

## Release Notes

* **New Features**
  * Added directional power metrics for both single-phase and three-phase inverters, including grid import/export power and battery charge/discharge power readings.

* **Tests**
  * Added comprehensive tests to verify the accuracy of new directional power sensor readings.


dewet22-codex:

givenergy_modbus/model/inverter.py
```diff
@@ -844,6 +844,38 @@ def e_battery_discharge_total(self) -> float | None:
         """Canonical total battery discharge energy (kWh), routed by model (see #76)."""
         return self._battery_energy("discharge", "total")
 
+    @computed_field  # type: ignore[prop-decorator]
+    @property
+    def grid_import_power(self) -> int | None:
```
dewet22-codex, line 849: This leaves the new canonical directional-power surface unavailable on `ThreePhaseInverter`, even though that model already exposes the corresponding non-negative values as `p_meter_import`, `p_meter_export`, `p_battery_charge`, and `p_battery_discharge`. Since `Plant.inverter` returns either model and the motivation is downstream HA consumption, callers still cannot use these four new names without model-specific branching. I think the three-phase model needs matching computed aliases (with its native fields as sources) before this closes #205.

dewet22: Good catch. Added four computed aliases to ThreePhaseInverter forwarding to its native non-negative registers (p_meter_import → grid_import_power, etc.) in c219dd1 — Plant.inverter callers can now use the canonical names on both models without branching.

dewet22-codex:

dewet22-codex (approved):
I’ve re-reviewed the latest changes. The three-phase aliases address my earlier concern, and I have no further concerns.

PR #255, opened 2026-06-12
fix(pdu): CRC-failed frames must not commit over good cache

dewet22:
## Problem

Five `wire=0x0000` CRC failures in ~1.5 h (field report from a Home Assistant install, 2026-06-12) were silently overwriting valid cache data with zero-filled payloads. Each event books a spurious TOTAL_INCREASING reset in HA — roughly +27,300 kWh on the lifetime PV counter per event.

The mechanism: `_validate_check_code()` has always logged a WARNING on CRC mismatch, but set no flag and took no other action, so `Plant.update()` committed the corrupted frame regardless. The `wire=0x0000` signature is a frame whose tail (including the CRC bytes themselves) was zeroed by the transport — consistent with bus stress immediately before each event (concurrent actor on the same dongle).

## Approach

Three-state CRC policy on `ReadRegistersResponse`:

- **Default** (`lenient_crc_commit=False`): `crc_failed=True` is set, WARNING fires, `Plant.update()` skips the commit — last-good cache is preserved, both warm (data already cached) and cold (cache stays empty until a valid frame arrives).
- **Opt-in lenient** (`lenient_crc_commit=True`): same flag + WARNING, but commit is allowed. Intended for frame-capture/forensic tools that want every frame regardless of CRC, or the edge case of a dongle that consistently emits bad CRCs on valid frames.
- **Strict** (`strict_crc=True`): raises `InvalidPduState`, unchanged from before.

The `lenient_crc_commit` flag is the signal that a consumer has made an explicit, informed choice to accept corrupt data. The old behaviour was effectively `lenient_crc_commit=True` by default, which I'd justified with a "some dongles emit bad CRCs" hypothesis that the corpus data doesn't support (3,106 register frames across all fixtures: 4 CRC failures, all `wire=0x0000`; zero non-zero mismatches).

## Changes

- `ReadRegistersResponse`: adds `crc_failed: bool = False` (per-instance), `lenient_crc_commit: ClassVar[bool] = False`; updates `_validate_check_code()` to set the flag and update the WARNING message.
- `Plant.update()`: guard after serial adoption, before `_commit_bank` — skips the commit for CRC-failed frames unless `lenient_crc_commit` is set.
- `TransparentMessage.__str__`: excludes `crc_failed` from the repr (it's a processing flag, not a PDU field).
- Tests: 7 new/updated PDU + plant tests covering default-skip, strict-raise, lenient-commit, warm-cache-protected, cold-cache-protected.

## Test plan

- [ ] All 996 tests pass (`uv run pytest --cov -q`)
- [ ] tox py311–py314 + lint + build all green
- [ ] After deployment: `wire=0x0000` events produce "Skipping commit of CRC-failed response" rather than silently zeroing registers; `block_age()` stays at the last valid poll time

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

## Release Notes

* **New Features**
  * Added optional lenient CRC validation mode allowing configurable handling of frames with integrity failures.

* **Bug Fixes**
  * Improved CRC error handling: frames with validation failures are now properly rejected and trigger automatic retries instead of committing invalid data to cache.

* **Tests**
  * Added comprehensive test coverage for CRC validation policies and frame rejection behaviour.

### DIFF
--- a/givenergy_modbus/client/client.py
+++ b/givenergy_modbus/client/client.py
@@ -1216,6 +1216,11 @@ class Client:
                 # because nothing yields between set_result and plant.update, but
                 # that's fragile to future refactors — make it explicit.
                 self.plant.update(message)
+                # Don't resolve the future for a discarded CRC-failed frame — leave it
+                # pending so send_request_and_await_response's timeout/retry fires a fresh
+                # request rather than treating a corrupt frame as a successful read.
+                if getattr(message, "crc_failed", False) and not getattr(message, "lenient_crc_commit", False):
+                    continue
                 future = self.expected_responses.get(message.shape_hash(), None)
                 if future and not future.done():
                     future.set_result(message)
--- a/givenergy_modbus/client/client.py
+++ b/givenergy_modbus/client/client.py
@@ -22,6 +22,7 @@ from givenergy_modbus.model.aio_battery import AioBatteryModule
 from givenergy_modbus.model.battery import Battery
 from givenergy_modbus.model.ems import EmsRegisterGetter
 from givenergy_modbus.model.inverter import Model, resolve_model
+from givenergy_modbus.model.lv_bcu import LV_BCU_ADDRESS, LvBcu
 from givenergy_modbus.model.meter import Meter
 from givenergy_modbus.model.plant import Plant, PlantCapabilities
 from givenergy_modbus.model.register import HR, IR
@@ -532,6 +533,40 @@ class Client:
                 continue
             caps.aio_battery_module_addresses.append(addr)
 
+    async def _detect_lv_bcu(
+        self,
+        caps: PlantCapabilities,
+        prior: PlantCapabilities | None,
+        probe_timeout: float,
+        probe_retries: int,
+    ) -> None:
+        """Probe the LV BCU page and set caps.lv_bcu_address when present (#241).
+
+        The stack-level BMS block (doc §4.4.1.1) answers at 0x31 IR(60-63) on LV
+        systems, firmware-gated: absent units still respond, just with an all-zero
+        block, so is_valid() (any-word-non-zero) is the presence test and the probe
+        costs nothing on healthy systems. Hinted mode follows the meter convention:
+        only re-check an address the prior captured — prior None-ness is trusted, so
+        a firmware upgrade that adds the block needs a cold detect to notice.
+        """
+        bcu_addr = prior.lv_bcu_address if prior is not None else LV_BCU_ADDRESS
+        if bcu_addr is None:
+            return
+        if not await self._probe(
+            ReadInputRegistersRequest(base_register=60, register_count=60, device_address=bcu_addr),
+            timeout=probe_timeout,
+            retries=probe_retries,
+        ):
+            return
+        bcu_cache = self.plant.register_caches.get(bcu_addr)
+        if not bcu_cache:
+            return
+        try:
+            if LvBcu.from_register_cache(bcu_cache).is_valid():
+                caps.lv_bcu_address = bcu_addr
+        except (KeyError, ValueError):
+            _logger.debug("detect: LV BCU probe at 0x%02x failed to decode — skipping", bcu_addr, exc_info=True)
+
     async def _ems_rollup_cross_check(self, timeout: float, retries: int) -> None:
         """Read IR(2040,55) at detect time and sanity-check the per-managed-inverter rollup.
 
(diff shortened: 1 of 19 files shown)


givenergy_modbus/model/lv_bcu.py
```diff
@@ -0,0 +1,73 @@
...
+    def is_valid(self) -> bool:
+        """Try to detect if the LV BCU block is present based on its attributes.
+
+        All-zero means absent/not supported (firmware-gated — see module docstring),
+        so presence requires at least one non-zero word.
+        """
+        return any(
+            getattr(self, f) not in (None, 0)
+            for f in ("bms_status_1", "bms_status_2", "request_charge_current", "request_discharge_current")
+        )
```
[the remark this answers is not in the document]
dewet22: Applied — direct attribute access in 7466c6b.

dewet22:
Reverted the Gemini suggestion in f90b230 — direct attribute access on `LvBcu` breaks mypy because the class is built with `create_model()` at runtime and mypy cannot see the dynamically-defined attributes. The `getattr` pattern is the correct approach here for the same reason other `create_model`-based classes in this codebase use it.

dewet22-codex:

givenergy_modbus/model/plant.py
```diff
@@ -643,6 +643,14 @@ def update(self, pdu: ClientIncomingMessage, *, received_at: datetime | None = N
         if device_address in (0x11, 0x31):
             self.inverter_serial_number = pdu.inverter_serial_number
 
+        if getattr(pdu, "crc_failed", False) and not getattr(pdu, "lenient_crc_commit", False):
```
dewet22-codex: This guard runs after several state mutations: a cache is created for the decoded device address, `data_adapter_serial_number` is replaced, and an 0x11/0x31 response replaces `inverter_serial_number`. The CRC covers the device address and inverter serial, so those values are untrusted on exactly the frames being rejected here; a corrupt 0x11 response can still clobber the stable inverter identity even though its register bank is skipped. I think the non-lenient guard needs to run before any Plant state is changed (including cache creation), with a regression test preserving the serials.

givenergy_modbus/model/plant.py
```diff
@@ -643,6 +643,14 @@ def update(self, pdu: ClientIncomingMessage, *, received_at: datetime | None = N
...
+            _logger.warning(
+                "Skipping commit of CRC-failed response from 0x%02x (base=%d) — last-good cache preserved",
+                device_address,
+                getattr(pdu, "base_register", 0),
+            )
+            return
```
dewet22-codex: Returning here only skips `Plant.update()`: `_task_network_consumer` immediately resolves the matching response future afterwards, so `send_request_and_await_response()` treats this CRC-failed frame as success and does not use its configured retries. On a cold `detect()`/`load_config()` this can leave the required cache empty or stale while the caller continues as though the read succeeded. Since the default policy calls the frame discarded, I think the consumer also needs to leave the future unresolved (or otherwise trigger a retry) for non-lenient CRC failures.

givenergy_modbus/model/plant.py
dewet22: Fixed in 2f01bae — guard now runs before any Plant state mutation, including cache creation and serial adoption. Added regression test covering the 0x11 serial case.

givenergy_modbus/model/plant.py
dewet22: Fixed in 2f01bae — _task_network_consumer now skips future.set_result() for non-lenient CRC-failed frames, so send_request_and_await_response's timeout/retry fires a fresh request instead of returning stale/empty cache to the caller.

dewet22-codex:

givenergy_modbus/client/client.py
```diff
@@ -1216,6 +1216,11 @@ async def _task_network_consumer(self):
                 # because nothing yields between set_result and plant.update, but
                 # that's fragile to future refactors — make it explicit.
                 self.plant.update(message)
+                # Don't resolve the future for a discarded CRC-failed frame — leave it
+                # pending so send_request_and_await_response's timeout/retry fires a fresh
+                # request rather than treating a corrupt frame as a successful read.
+                if getattr(message, "crc_failed", False) and not getattr(message, "lenient_crc_commit", False):
```
dewet22-codex, line 1222: The control flow now addresses the retry concern, but this new branch has no regression coverage: there is no client test feeding a CRC-failed response through `_task_network_consumer` and proving the first future remains unresolved while a subsequent retry can succeed. That is also leaving the patch/project coverage checks failing on this head. Because this is the behaviour that prevents a discarded frame from being reported as a successful read, I think it needs a focused consumer/retry test before sign-off.

dewet22: Added in 1dedd0d — test_consumer_does_not_resolve_future_for_crc_failed_frame feeds a CRC-failed frame through _task_network_consumer and asserts the registered future stays pending. Coverage gap closed; waiting for CI to confirm.

dewet22-codex:
I’ve started re-reviewing the concurrently updated head.

dewet22-codex (approved):
I’ve re-reviewed the concurrently updated head. The consumer regression test closes my remaining concern, all prior findings are addressed, and I have no further concerns.

Issue #256, opened 2026-06-12
fix(battery): guard battery banks against sub-bus splice corruption — valid-CRC garbage commits over good cache


dewet22:
Follow-up to the characterisation in #147 (see the closing comment there for the full evidence). Summary: BMS sub-bus corruption is re-framed by the dongle with a valid CRC, so it passes #255's guard and commits. Observed classes — all with rigidly deterministic register cohorts, none caught by current defences:

| Class | Cohort | Why nothing catches it |
|---|---|---|
| Capacity splice | {IR(88/89) → `temp << 8`, status words → shifted temp copies, IR(98) fw version → `0xXX00`} | `cap_remaining`/`cap_design`/`cap_calibrated` have no bounds |
| Cell + temp splice | {IR(74) → `0x0E10` (copy of IR(91)), IR(103) → 0, IR(104) → `0x000B`} — identical garbage across packs | every value inside its per-register bounds |
| Min/max temp splice | {IR(103) → 0, IR(104) → `0x000B`} | as above |
| Temp-cohort zeros | {IR(76-79), IR(81), IR(115)} zeroing together; single-poll or sustained multi-poll runs | 0 °C is in-bounds; bank isn't all-zero so #206 doesn't fire |

The field impact is the remaining "implausible reads" classes from the hass soak: garbage Ah spikes, temp dropouts to zero, and the recurring fake 3.600 V cell reading — each booking corrupt history into downstream consumers, roughly 30×/day on the reporting install.

Classifying all 1,103 consecutive-frame transitions in the capture corpus by the number of physically impossible per-register movements gives perfect separation — 1,086 clean, 17 corruption-related, zero false trips — and grounds the design empirically:

| Trip profile | Verdict | Corpus support |
|---|---|---|
| Any change to a constant register (IR(97), IR(98), serial block, IR(115)) | reject outright | present in the capacity and fw-version splices *and every poll of the temp-zero runs*; never in clean frames |
| ≥2 physics-impossible deltas | reject outright | all observed events; physically independent quantities (a cell voltage, a temperature, a lifetime capacity) cannot all step impossibly in one ~30 s poll |
| 1 physics-impossible delta | escrow: hold last-good, commit the *next* poll's value if it also reads away from last-good rather than snapping back (every observed splice reverts; genuine steps persist) | zero occurrences in corpus — pure defence-in-depth for unobserved shapes; worst case is one poll of latency |

Design notes:

[...]

Issue #147, opened 2026-05-31
BMS block dropouts on battery sub-bus (Pattern B from #78)


dewet22:
Split out from #78 (now closed — Pattern A, the dongle IR(0-59) substitution, is handled by `is_suspicious()` in `Plant.update()`).

## Pattern B — BMS block dropouts

Battery-side input register reads occasionally return zero for an entire response block. From the #78 investigation: 90 anomaly windows over 26 hours, 21 of them affecting 5 distinct battery fields at once (a complete block dropout); only 6 of 90 were isolated single-field events. Strongly suggests block-level read failure on the battery sub-bus, not per-field intermittency, and distinct from Pattern A (different layer, no temporal correlation).

## Why it's hard to action from here

Unlike Pattern A — which has a deterministic fixed-constant fingerprint the library can detect and discard — a zeroed BMS block is hard to distinguish from a genuine all-zero reading (an absent/empty battery slot legitimately reads zero) without more signal. So this isn't a simple discard guard.

## Next steps

- Paired capture to confirm the layer: modbus TCP from the dongle alongside RS485 from the inter-battery bus, so a dropout can be attributed to the sub-bus rather than the dongle or transport.
- Once characterised, decide whether the library can/should flag it (e.g. a freshness/staleness signal per battery — relates to #65 and #91) versus leaving it to the consumer.

Low priority / investigation — no release gating.

dewet22:
Field evidence from the hass-side soak (2026-06-12, on modbus 2.3.1) plus a scan of our own capture corpus has settled the characterisation question this issue was parked on — and broadened it well beyond block dropouts.

The mechanism: the batteries talk to the dongle on an internal sub-bus, and the dongle *re-frames* that data with a fresh CRC before it reaches the wire. Corruption on the sub-bus side therefore arrives with a perfectly valid CRC — the #255 guard structurally cannot see it. The paired-capture experiment this issue proposed turned out to be unnecessary: the corruption encodes its own origin. Garbage `cap_remaining` values are quantised as `temp_deci << 8` — e.g. 576.00 Ah = `0xE100` with the pack at 22.5 °C (`0xE1` = 225), tracking each pack's live temperature to the decimal, four-for-four across two packs in one evening.

[...]

PR #257, opened 2026-06-12
test(debug): add battery sub-bus splice corpus tools (#147, #256)

dewet22:
Checks in the two offline capture-analysis tools behind the #147 sub-bus splice characterisation and the #256 detection design — worth keeping for archaeology, and for re-running against future captures (HV stacks, new dongle firmwares) as they land.

- `tests/debug/byte_slip_scan.py` — fingerprints the known splice shapes (transient `0xXX00`-quantised `cap_remaining` garbage, temp-cohort zero runs with single-poll vs sustained classification, a cell pinned at the recycled status-word value) and locates the changed-byte window plus its byte-shifted donor region within the frame.
- `tests/debug/physics_delta_classify.py` — the generic classifier: counts per-poll register movements beyond per-register-class physics thresholds (bitfields exempt; constant registers flagged on any change). This produced the corpus numbers grounding #256: 1,103 transitions → 1,086 clean, 17 corruption-related, zero false trips.

Both are cited from #147 and #256. No library code changes; `Changelog: skip`.

## Release Notes

* **Tests**
  * Added diagnostic command-line scanner for analysing battery system capture files to identify and report potential data corruption patterns in battery monitoring.
  * Added diagnostic command-line tool for classifying battery system register state transitions and identifying physically impossible changes in register values detected in captured monitoring data.

### DIFF
--- a/tests/debug/byte_slip_scan.py
+++ b/tests/debug/byte_slip_scan.py
@@ -1,0 +1,245 @@
+#!/usr/bin/env python3
+r"""Scan wire captures for battery sub-bus splice/slip corruption — #147.
+
+Field evidence (givenergy-hass, 2026-06-12, modbus 2.3.1) identified the
+mechanism behind valid-CRC battery garbage: corruption on the battery
+sub-bus side of the dongle, re-framed with a fresh (valid) CRC. The wire
+CRC is computed by the dongle over already-corrupt payload, so #255's CRC
+guard structurally cannot catch it.
+
+Corpus analysis sharpened the shape: it is a *windowed splice*, not a
+tail-to-end byte slip. A short window of the payload is overwritten with a
+byte-shifted copy of data from elsewhere in the frame (observed donor: the
+temperature block), while bytes before and after the window are intact.
+E.g. ``cap_remaining`` raw ``0xED00`` = pack temp 23.7 °C (``0x00ED``)
+donating its low byte as the garbage value's high byte — the distinctive
+``0xXX00`` quantisation seen in the field.
+
+Distinct from (and also reported here): *sustained temp-group zeros* —
+IR(76-79) reading exactly 0 for several consecutive polls while cells stay
+live, then recovering. Multi-poll persistence means the BMS itself returned
+zeros (device-origin), not per-frame transport corruption; cross-pack
+clustering within the same wall-clock window implicates the shared sub-bus
+path. A different failure needing different handling (sub-group
+hold-last-good rather than frame rejection).
+
+The scanner replays capture files, watching LV battery banks (devices
+0x32-0x37, IR(60,120)) for:
+
+1. transient ``cap_remaining`` garbage — low byte 0x00 with a large jump
+   vs the previous poll (the splice fingerprint);
+2. temp-group all-zero runs (reported with run length: single-poll =
+   possibly splice, sustained = device-origin);
+3. a cell voltage of exactly 3.600 V (0x0E10) while pack-mates sit
+   elsewhere — likely the status word IR(91) (persistently 0x0E10 on
+   observed packs) spliced into a cell slot, not a real voltage.
+
+For frames flagged with (1) or (3) it extracts the changed-byte window vs
+the previous poll and searches the frame for a byte-shifted donor region
+matching the spliced content.
+
+Usage::
+
+    uv run python tests/debug/byte_slip_scan.py tests/fixtures/captures/*/*.log
+
+Capture files are ``givenergy-cli capture`` ``.log`` output
+(``<ts> rx|tx <hex>``).
+"""
+
+from __future__ import annotations
+
+import argparse
+import asyncio
+import re
+import statistics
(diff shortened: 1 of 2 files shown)


tests/debug/byte_slip_scan.py
```diff
@@ -0,0 +1,245 @@
...
+                reg_lo, reg_hi = BANK_BASE + start // 2, BANK_BASE + (start + len(needle)) // 2
+                return f"window is a copy of bytes at IR({reg_lo})-IR({reg_hi}) (offset {start - lo:+d})"
+        # tolerate one mismatched byte (live values drift between donor and splice)
+        for start in range(0, len(sus) - len(needle)):
+            if start == lo:
+                continue
+            region = sus[start : start + len(needle)]
+            if sum(1 for a, b in zip(region, needle) if a != b) <= 1:
+                reg_lo, reg_hi = BANK_BASE + start // 2, BANK_BASE + (start + len(needle)) // 2
+                return f"window ~matches bytes at IR({reg_lo})-IR({reg_hi}) (offset {start - lo:+d}, 1 byte drift)"
```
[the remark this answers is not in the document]
dewet22: Quite right — leftover from an earlier version that compared shifted byte streams; removed in 94626e3.

dewet22-codex:

tests/debug/byte_slip_scan.py
```diff
@@ -0,0 +1,244 @@
...
+    The observed splice copies a byte-shifted slice of the same frame (e.g.
+    the temperature block) over the window. Finding such a donor elsewhere in
+    the frame is the smoking gun that the garbage is recycled frame content.
+    """
+    sus = b"".join(v.to_bytes(2, "big") for v in suspect)
+    lo, hi = window
+    needle = sus[lo : hi + 1]
+    if len(needle) < 3:
+        return None
+    for start in range(0, len(sus) - len(needle)):
```
dewet22-codex: Both donor-search loops exclude the final valid start offset. For a 120-byte bank and an N-byte needle, `start == 120 - N` is a valid slice, but `range(0, len(sus) - len(needle))` stops one before it. A donor region ending exactly at IR(119) will therefore be missed (including by the one-byte-drift pass). I think these ranges need the inclusive `+ 1` bound.

tests/debug/byte_slip_scan.py
```diff
@@ -0,0 +1,244 @@
...
+                            f"  changed window: bytes {lo}-{hi} (IR({BANK_BASE + lo // 2})-IR({BANK_BASE + hi // 2}))"
+                        )
+                        donor = find_donor(regs, window)
+                        if donor:
+                            print(f"  donor: {donor}")
+
+            if not hits and not temp_zero:
+                tracker.prev = regs
+
+    print(
```
dewet22-codex, line 248: Active temperature-zero runs are only emitted when a later non-zero poll ends them. If the capture ends while IR(76-79) are still zero, the tracker retains the run and this summary returns without reporting it, so a sustained failure occupying the tail of a capture disappears from the analysis. I think the trackers need a final flush after the frame loop so those open runs are reported too.

tests/debug/physics_delta_classify.py
```diff
@@ -0,0 +1,161 @@
...
+        with open(path, encoding="utf-8") as f:
+            for line in f:
+                m = CAPTURE_LINE.match(line.strip())
+                if not m or m.group(2) != "rx":
+                    continue
+                try:
+                    raw = bytes.fromhex(m.group(3))
+                except ValueError:
+                    continue
+                entries.append((path.name, m.group(1), raw))
```
dewet22-codex: Using only `path.name` as the capture identity can merge unrelated captures that happen to share a basename. In that case `prevs[(fname, device)]` compares the final bank from one file with the first bank from another and reports a false physics transition; the splice scanner has the same state-key collision. Since these CLIs accept arbitrary paths, I think the full path (or a per-input unique identifier) needs to be carried through as the tracker key, even if the displayed label stays short.

tests/debug/byte_slip_scan.py
dewet22: Fixed in 59958c7 — both loops now use the inclusive bound.

tests/debug/byte_slip_scan.py
dewet22: Fixed in 59958c7 — added a flush after the frame loop; it immediately revealed a real 3-poll tail run (device 0x34, 0x11-poll capture) the scanner had been silently dropping. Good catch.

tests/debug/physics_delta_classify.py
dewet22: Fixed in 59958c7 — tracker keys now carry the full path in both tools; display labels stay short via Path(...).name.

dewet22-codex:

dewet22-codex (approved):
I’ve re-reviewed the latest changes. All three earlier concerns are addressed, and I have no further concerns.

PR #258, opened 2026-06-14
fix(battery): guard LV battery banks against sub-bus splice corruption (#256)

### LINKED ISSUE #256: fix(battery): guard battery banks against sub-bus splice corruption — valid-CRC garbage commits over good cache

dewet22:
## Summary

BMS sub-bus corruption is re-framed by the dongle with a **valid CRC** (#147), so it passes the #255 CRC guard and commits valid-looking garbage over good cache — Ah spikes, temp dropouts to zero, fake 3.600 V cells — ~30×/day on the reporting install. None of the existing defences catch it: corrupted values are in-bounds, the bank is never all-zero (Pattern B / #206 doesn't fire), and the serial can stay valid (`is_coherent` passes).

- **New module `givenergy_modbus/model/battery_splice.py`** — single source of truth for per-register-class physics thresholds, empirically grounded on the 1,103-transition capture corpus (1,086 clean / 17 corrupt / zero false trips). Lifted from the #257 debug tool so the live guard and the corpus classifier can never drift apart.
- **`Plant._commit_bank` splice guard** — runs last, after serial-coherence and bounds, on full (≥60 register) `BatteryRegisterGetter` banks only:
  - **≥2 physics trips or any IMMUTABLE change** → reject outright, keep last-good
  - **Exactly 1 physics trip** → escrow: hold last-good, commit only if the next poll reads the same value again (genuine steps persist; every observed splice reverts within one poll)
  - **0 trips** → commit normally
- **IR(115) exempt from IMMUTABLE** — `battery.py` documents it as mutable `usb_device_inserted` (observed values 0/8/11); a USB insert/remove would false-trip. The temp-zero corruption cohort it was meant to catch trips ≥2 temperatures on its own.
- **`tests/debug/physics_delta_classify.py` refactored** — now imports thresholds from the production module.

## Test plan

- [x] 10 unit tests in `tests/model/test_battery_splice.py` — pin threshold tables, trip logic, `present` gating, absolute IR numbering, pair-rule uint32 assembly
- [x] 13 integration tests in `tests/model/test_plant.py` — reject-outright (immutable + capacity cohort, cell+temp, temp-zeros), false-positive guard (clean jitter commits), single-trip escrow (confirm, snapback, two-wilds-in-a-row), cold-start, IR(115) exemption, short-read gate, non-battery-device gate, per-device escrow isolation, staleness preservation on reject
- [x] 1023/1023 tests pass; ruff, mypy, bandit clean
- [ ] Corpus re-validation: `uv run python tests/debug/physics_delta_classify.py tests/fixtures/captures/*/*.log` — expect 17 corruption transitions flagged, zero false trips (unchanged vs #257 baseline except any IR(115)-driven immutable lines drop)

## Known trade-offs

- **Genuine `bms_firmware_version` or serial change** (BMS firmware update, battery swap) is rejected until the next reconnect rebuilds the Plant (fresh cache → cold-start adopt). Rare and reconnect-coupled; integrity-preserving. If long-lived-Plant swaps ever bite, the fix is to extend the escrow machinery to immutable trips — noted as a follow-up, not built here.
- **Single-trip latency** — a genuine physics-singleton step (the one in the corpus: MOSFET temp +6.3 °C in one poll) is delayed by one poll while it confirms. Worst case: one stale reading in 30 s.

Closes #256.

## Release Notes

* **New Features**
  * Added battery splice guard to detect and reject physically impossible battery bank transitions, preventing corrupted data from being cached.
  * Implemented escrow logic to hold questionable readings pending confirmation on the next poll before committing to cache.

* **Tests**
  * Comprehensive unit and integration tests for battery transition validation and corruption detection.

### DIFF
--- a/givenergy_modbus/model/plant.py
+++ b/givenergy_modbus/model/plant.py
@@ -543,6 +543,15 @@ class Plant(GivEnergyBaseModel):
     # Ephemeral runtime state, rebuilt with a fresh Plant on reconnect; not serialised.
     _splice_escrow: dict[int, tuple[int, int]] = PrivateAttr(default_factory=dict)
 
+    # Splice-guard observation clock (#256): per battery device address, the ingestion time of the
+    # last full IR(60,60) bank the guard examined — accepted OR rejected. The stale-baseline bypass
+    # keys off this, NOT the last accepted commit: a sustained corruption run (every poll rejected)
+    # leaves the last-good commit ageing past the bypass window, but keeps arriving each poll, so
+    # this clock stays ~one poll old and the bypass never fires. Only a genuine polling gap (real
+    # outage, no banks at all) makes it stale and resets to cold-start adoption. Ephemeral; resets
+    # with a fresh Plant on reconnect; not serialised.
+    _splice_last_seen: dict[int, datetime] = PrivateAttr(default_factory=dict)
+
     # Direct-inverter register caches injected by add_direct_source() for multi-Client
     # reconciliation (#106 Phase 3). Stored separately from register_caches to avoid the
     # Modbus address collision (both EMS controller and direct inverter live at 0x11).
@@ -794,26 +803,22 @@ class Plant(GivEnergyBaseModel):
 
         Only full battery banks are guarded: a short read can't be physics-classified, a
         device with no prior committed bank (cold start) has nothing to compare against, and
-        a baseline older than ``STALE_BYPASS_SECONDS`` is too stale for per-poll thresholds
-        to apply (legitimate multi-field drift after a reconnect gap would trip them) — all
-        three fall through as a no-op commit so the cache self-heals without manual recovery.
+        a gap of more than ``STALE_BYPASS_SECONDS`` since the last *observed* full bank — a
+        genuine polling outage, not a rejection streak — is too stale for per-poll thresholds
+        to apply (legitimate multi-field drift over the gap would trip them) — all three fall
+        through as a no-op commit so the cache self-heals without manual recovery.
         """
         if register_count < 60:
             return True
-        # Stale-baseline bypass: after a long gap the per-poll thresholds don't hold (legitimate
-        # SOC/temp/cap drift would exceed them), and rejected banks never advance the timestamp,
-        # so without this the guard would pin the cache to the stale baseline indefinitely after
-        # a network outage or prolonged refresh failure. Adopt unconditionally and rebuild.
-        last_age = self.block_age(device_address, "IR", BANK_BASE, 60, now=now)
-        if last_age is not None and last_age > STALE_BYPASS_SECONDS:
-            self._splice_escrow.pop(device_address, None)
-            _logger.info(
-                "Battery bank for device 0x%02x: prior commit %.0f s old — splice guard bypassed "
-                "(stale baseline); adopting as new baseline.",
-                device_address,
-                last_age,
-            )
-            return True
+        now_ts = now if now is not None else datetime.now(UTC)
+        if now_ts.tzinfo is None:
+            now_ts = now_ts.replace(tzinfo=UTC)
+        # Record this observation up front — a rejected bank is still an observation, so the gap
+        # computed below measures time since we last *saw* a full bank, not since we last accepted
+        # one. This is what separates a real outage from a sustained corruption run (see below).
+        prev_seen = self._splice_last_seen.get(device_address)
+        self._splice_last_seen[device_address] = now_ts
+
         cache = self.register_caches[device_address]
         prev = [0] * 60
         new = [0] * 60
@@ -828,6 +833,23 @@ class Plant(GivEnergyBaseModel):
                 present.add(i)
         if not present:
             return True  # cold start: no last-good to compare against
+
+        # Stale-observation bypass: after a genuine polling gap the per-poll thresholds don't hold
+        # (legitimate SOC/temp/cap drift over the gap would exceed them), and rejected banks never
+        # advance the cache, so without recovery the guard would pin the cache to the stale baseline
+        # forever after a network outage. Crucially this keys off the last *observation*, not the
+        # last accepted commit: a sustained corruption run (e.g. a multi-poll temp-zero stream)
+        # keeps arriving and being rejected each poll, ageing the last *commit* past the window —
+        # but prev_seen stays ~one poll old, so this does NOT fire and the corruption stays rejected.
+        if prev_seen is not None and (now_ts - prev_seen).total_seconds() > STALE_BYPASS_SECONDS:
+            self._splice_escrow.pop(device_address, None)
+            _logger.info(
+                "Battery bank for device 0x%02x: %.0f s since the last observed bank — splice guard "
+                "bypassed (stale baseline); adopting as new baseline.",
+                device_address,
+                (now_ts - prev_seen).total_seconds(),
+            )
+            return True
 
         phys, immut = classify_transition(prev, new, present)
 
--- a/givenergy_modbus/model/plant.py
+++ b/givenergy_modbus/model/plant.py
@@ -8,7 +8,12 @@ from pydantic import BaseModel, ConfigDict, Field, PrivateAttr, model_validator
 from givenergy_modbus.model import GivEnergyBaseModel
 from givenergy_modbus.model.aio_battery import AioBatteryModule
 from givenergy_modbus.model.battery import Battery, BatteryRegisterGetter
-from givenergy_modbus.model.battery_splice import BANK_BASE, THRESHOLD_BY_CLASS, classify_transition
+from givenergy_modbus.model.battery_splice import (
+    BANK_BASE,
+    STALE_BYPASS_SECONDS,
+    THRESHOLD_BY_CLASS,
+    classify_transition,
+)
 from givenergy_modbus.model.devices import DeviceType, PlantDevice
 from givenergy_modbus.model.devices import Inverter as UnifiedInverter
 from givenergy_modbus.model.ems import Ems
@@ -665,7 +670,7 @@ class Plant(GivEnergyBaseModel):
 
         if isinstance(pdu, ReadHoldingRegistersResponse):
             incoming = {HR(k): v for k, v in pdu.to_dict().items()}
-            if self._commit_bank(device_address, incoming, pdu.register_count):
+            if self._commit_bank(device_address, incoming, pdu.register_count, received_at=received_at):
                 self._stamp_block(device_address, "HR", pdu.base_register, pdu.register_count, received_at)
                 self._track_content_change(
                     device_address, "HR", pdu.base_register, pdu.register_count, incoming, received_at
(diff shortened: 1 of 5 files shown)


dewet22-codex:
Starting a review round now.

givenergy_modbus/model/plant.py
```diff
@@ -745,9 +754,92 @@ def _commit_bank(self, device_address: int, incoming: dict, register_count: int
...
+            cached = cache.get(reg)
+            incoming_val = incoming.get(reg)
+            prev[i] = cached if cached is not None else 0
+            new[i] = incoming_val if incoming_val is not None else prev[i]
+            if incoming_val is not None and cached is not None:
+                present.add(i)
+        if not present:
+            return True  # cold start: no last-good to compare against
+
+        phys, immut = classify_transition(prev, new, present)
```
dewet22-codex, line 854: [P1] Gate physics classification on sample age. These thresholds are calibrated for ~30-second consecutive polls, but this compares against the cached last-good after an arbitrarily long gap. `Client.connect()` deliberately preserves `self.plant`, so after a network outage (or any prolonged refresh failure) SOC, cell voltage, temperature, and capacity can legitimately exceed multiple thresholds. This branch then rejects the bank outright; because rejected banks never advance the cache or timestamp, every subsequent poll is compared to the same stale baseline and can be rejected forever until the process rebuilds the Plant. Please bypass/reset the guard when the prior IR(60,60) commit is older than a bounded interval (or otherwise add a recovery path), and cover the long-gap multi-trip case.

dewet22-codex (changes requested):
One blocking lifecycle issue: the per-poll physics thresholds are applied after unbounded polling gaps, which can permanently pin the battery cache to stale data.

dewet22-codex:
Review round complete. I found one blocking lifecycle issue: the live guard applies ~30-second physics thresholds after unbounded polling gaps. Because reconnects preserve the existing Plant and rejected banks do not advance cache state, a legitimate multi-field change after an outage can cause every subsequent battery bank to be rejected against the same stale baseline indefinitely. I left the details and requested recovery behavior inline.

givenergy_modbus/model/plant.py
dewet22: Fixed in 3be1d05. Added  to `battery_splice.py` and a bypass check at the top of `_splice_guard`: if `block_age(device, "IR", 60, 60)` exceeds 300 s, the guard clears escrow, logs at INFO, and returns True (cold-start semantics). `received_at` is now threaded through `_commit_bank` → `_splice_guard` → `block_age(now=...)` so the age measurement is consistent with the rest of the stamping logic and controllable in tests without patching `datetime.now`. New test `test_splice_stale_baseline_bypass` seeds at `_T0`, feeds a 2-physics-trip bank at `_T0 + 400 s`, and asserts it commits with an INFO log and no rejection WARNING.

dewet22-codex:
Starting a re-review round now.

givenergy_modbus/model/plant.py
```diff
@@ -745,9 +769,112 @@ def _commit_bank(self, device_address: int, incoming: dict, register_count: int
...
+        # SOC/temp/cap drift would exceed them), and rejected banks never advance the timestamp,
+        # so without this the guard would pin the cache to the stale baseline indefinitely after
+        # a network outage or prolonged refresh failure. Adopt unconditionally and rebuild.
+        last_age = self.block_age(device_address, "IR", BANK_BASE, 60, now=now)
+        if last_age is not None and last_age > STALE_BYPASS_SECONDS:
+            self._splice_escrow.pop(device_address, None)
+            _logger.info(
+                "Battery bank for device 0x%02x: prior commit %.0f s old — splice guard bypassed "
+                "(stale baseline); adopting as new baseline.",
+                device_address,
```
dewet22-codex: [P1] Do not use last accepted commit age alone for this bypass. Rejected splice banks intentionally do not advance `register_block_updated_at`, so a sustained corruption run keeps increasing `last_age`; once it exceeds 300 seconds, this branch unconditionally adopts the still-corrupt bank. Sustained multi-poll temp-zero runs are explicitly one of #256's observed corruption shapes, so this reopens the data-integrity failure after five minutes. Please distinguish a real observation gap from continuously received/rejected banks, for example by tracking the last battery-bank observation/attempt separately and only bypassing when that timestamp is also stale. Add a test that repeatedly feeds a corrupt >=2-trip bank past 300 seconds and verifies it remains rejected.

dewet22-codex (changes requested):
The stale-baseline recovery fixes the original deadlock, but currently allows sustained corruption to be committed after five minutes.

dewet22-codex:
Re-review round complete. The stale-baseline bypass resolves the original post-outage deadlock, but it creates a follow-up blocking issue: because rejected banks do not advance the committed-block timestamp, a sustained corruption run ages past 300 seconds and is then adopted unconditionally. I left an inline request to distinguish a true observation gap from continuously received/rejected banks and to cover the sustained-corruption case.

givenergy_modbus/model/plant.py
dewet22: Fixed in 03e4ef9. You're right — keying off the last *accepted* commit let a sustained rejection streak age past the window and then adopt the corrupt bank. The bypass now keys off a new per-device observation clock (`_splice_last_seen`), updated on every full bank the guard examines, accepted *or* rejected. It fires only when the time since the last *observed* bank exceeds the window — a genuine polling gap — so a corruption run (which keeps arriving each poll) keeps the clock ~one poll old and stays rejected indefinitely. Added `test_splice_sustained_corruption_never_bypassed` (20 polls / 600 s of temp-zero corruption stays fully rejected, bypass never fires) and a complement test confirming an intervening rejected bank resets the clock.

dewet22:
The latest `CHANGES_REQUESTED` re-review targets `3be1d05`, which keyed the stale-baseline bypass off the last *accepted commit* — the exact case where a sustained rejection streak ages past the window and gets adopted.

That's fixed at HEAD (`03e4ef9`, pushed after that review): the bypass now keys off a per-device observation clock (`_splice_last_seen`) that advances on **every** full bank examined, rejected or not. A genuine polling outage lets it go stale → recover; a sustained corruption run keeps it ~one poll old → stays rejected indefinitely. Covered by `test_splice_sustained_corruption_never_bypassed` (20 polls / 600 s of temp-zeros, never bypassed) plus a complement test. The inline thread is resolved; ready for a fresh look at `03e4ef9`.

dewet22-codex:
Starting a final re-review round now.

dewet22-codex (approved):
Final re-review complete. The observation-clock recovery correctly distinguishes genuine polling gaps from sustained rejected corruption, and the regression coverage exercises both paths. No further concerns.

dewet22-codex:
Final re-review sign-off: the latest changes resolve the stale-baseline and sustained-corruption concerns. The observation clock now distinguishes genuine polling gaps from continuously rejected banks, the relevant regression cases are covered, and CI is green. Approved.

PR #267, opened 2026-06-17
feat(inverter): derive three-phase p_battery and e_battery_throughput

### LINKED ISSUE #262: Unify 3-phase battery fields (p_battery / e_battery_throughput) behind synthetic LUT overrides once hardware-validated
## Background

A field report on a three-phase hybrid (givenergy-hass#174) flagged inverter-level **Battery Power**, **Battery Temperature** and **Battery Throughput** as present-but-frozen, while per-pack SOC refreshed normally.

Tracing it: `p_battery` (IR52), `t_battery` (IR56) and `e_battery_throughput` (IR6/7) all sit in the live IR(0,60) block, which `refresh()` polls every tick for three-phase too — so this isn't a refresh-coverage gap. They look frozen because three-phase firmware doesn't appear to populate those single-phase addresses. The three-phase LUT (`_THREE_PHASE_LUT` in `givenergy_modbus/model/inverter_threephase.py`) inherits those three keys verbatim from the single-phase LUT while deliberately overriding `battery_soc` (→IR1132) and `i_battery` (→IR1140).

The canonical three-phase battery telemetry lives in the 1000-range instead:

- **Power** → `p_battery_charge` IR(1138/1139) + `p_battery_discharge` IR(1136/1137)
- **Energy** → `e_battery_charge_total`/`_today` IR(1394/1395, 1392/1393) and `e_battery_discharge_total`/`_today` IR(1390/1391, 1388/1389)
- **Temperature** → no inverter-level battery temp exists on three-phase; only component temps (`t_inverter`/`t_boost`/`t_buck_boost`). Battery temperature is only available from the per-pack block.

## Current state

Consumers resolve this per-model for now (hass branches onto the 1000-range keys directly). The library deliberately leaves the inherited single-phase keys raw rather than synthesising values.

## What this issue tracks

The eventual unification: adding synthetic three-phase LUT overrides so the inherited keys Just Work for any consumer, consistent with how `battery_soc`/`i_battery` are already shadowed:

- `p_battery` ← net of `p_battery_discharge` − `p_battery_charge`, matching the single-phase sign convention
- `e_battery_throughput` ← `e_battery_charge_total` + `e_battery_discharge_total`

`t_battery` has no inverter-level three-phase source, so it would stay per-pack regardless — parity here is partial by nature.

## Why it's gated on hardware

I haven't validated the sign convention or scaling against a real three-phase unit, and synthesising a signed net (discharge − charge) plus a summed total blind is exactly where a sign-flip or scale error ships silently. The unification should wait on a three-phase capture to confirm direction and units before it lands. The hass#174 reporter is gathering a capture; that, or a capture from a contributor with three-phase hardware, would unblock this.

dewet22:
## What

On three-phase, `p_battery` and `e_battery_throughput` were *inherited* from the single-phase LUT (via the `dict(SP_LUT, **_THREE_PHASE_LUT)` merge), so they pointed at single-phase registers (`IR52`, `IR6/7`) that three-phase firmware never populates — they read **frozen**. This replaces them with derived `@computed_field` properties from the native three-phase registers:

- `p_battery` = `p_battery_discharge − p_battery_charge` (+ve = discharging, −ve = charging — matching the single-phase sign convention where `battery_discharge_power = max(0, p_battery)`)
- `e_battery_throughput` = `e_battery_charge_total + e_battery_discharge_total`

## Why

This was the frozen-battery-fields symptom from [givenergy-hass#174](https://github.com/dewet22/givenergy-hass/issues/174). hass already resolves these consumer-side; this is the upstream facade so any consumer keying off `p_battery` (cli, others) gets correct values on three-phase without per-model branching. The premise was validated against the GIV-3HY-11 capture: `p_battery_charge`/`p_battery_discharge` decode correctly and the sign is unambiguous.

## Implementation note

Both keys are dropped from the merged `REGISTER_LUT` before the computed properties are defined — pydantic v2 will not let a `@computed_field` shadow a generated model field. `build()`/`to_fields()`/`precision_of()`/`registers_of()` simply stop emitting them (confirmed: `precision_of` → None, `registers_of` → ()), and the metadata path already tolerates non-LUT computed fields (the existing `battery_charge_power` precedent). Return type is `float | None` (inputs are deci floats), None-propagating like the single-phase computed fields.

## Out of scope

The inherited single-phase `t_battery` (IR56) is likewise frozen on three-phase but has no inverter-level three-phase source register, so it can't be derived the same way — left as a separate open question.

Full suite green (1033 passed).

Closes #262

### DIFF
--- a/givenergy_modbus/model/inverter_threephase.py
+++ b/givenergy_modbus/model/inverter_threephase.py
@@ -447,9 +447,20 @@ class ThreePhaseInverterRegisterGetter(RegisterGetter):
     Merges the single-phase SinglePhaseInverterRegisterGetter LUT with three-phase-specific
     registers. Three-phase entries win for any key that appears in both (e.g.
     battery_soc moves from IR(59) to IR(1132)).
+
+    p_battery and e_battery_throughput are dropped from the merged LUT: on three-phase
+    they are DERIVED @computed_field properties on ThreePhaseInverter (from the native
+    p_battery_charge/discharge and the charge/discharge energy totals). Left register-backed
+    they would inherit single-phase IR(52) / IR(6,7), which three-phase firmware does not
+    populate (they read frozen) — and a generated pydantic field cannot be shadowed by a
+    same-named computed_field.
     """
 
-    REGISTER_LUT = dict(SinglePhaseInverterRegisterGetter.REGISTER_LUT, **_THREE_PHASE_LUT)
+    REGISTER_LUT = {
+        k: v
+        for k, v in dict(SinglePhaseInverterRegisterGetter.REGISTER_LUT, **_THREE_PHASE_LUT).items()
+        if k not in ("p_battery", "e_battery_throughput")
+    }
 
 
 _ThreePhaseInverterBase = create_model(  # type: ignore[call-overload]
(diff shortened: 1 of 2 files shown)


dewet22-codex:
Starting review.

dewet22-codex (approved):
Looks good. Deriving the three-phase battery facade from native charge/discharge registers avoids the frozen inherited single-phase registers, keeps the expected computed fields in `model_dump()`, and the tests cover sign, None propagation, and metadata behavior.

PR #275, opened 2026-06-17
fix(client): tear down connection when detect() fails (atomic connect+detect, #274)

### LINKED ISSUE #274: connect()+detect() atomicity: reset `connected` on detect failure to avoid "connected but undetected" state
## The footgun

A `detect()` that times out *after* `connect()` has already opened the socket leaves a client with `connected == True` but `capabilities is None`:

- `connect()` opens the streams and sets `connected = True` with no dependency on detection.
- `detect()` has several un-guarded `send_request_and_await_response()` calls (HR(0,60), the 0x32 battery read) that can raise `TimeoutError` / `CommunicationError`; it only assigns `capabilities` on full success, and never touches `connected`.
- `connected` is a plain stored boolean, reset only by `close()` or socket EOF in the consumer/producer tasks — none of which a failed `detect()` triggers.

The natural consumer idiom — "reconnect if `not client.connected`, otherwise `refresh()`" — then skips reconnecting and calls `refresh()`, which hits its `capabilities is None` guard and raises `PlantNotDetected`. hass#183 worked around it by discarding the client on any connect-failure, but every consumer (cli, third parties) hits the same wire.

## Proposal

Make detect-failure reset `connected` so connect+detect is atomic: either you get a fully-detected client or none, and the standard reconnect idiom gets the invariant for free.

Alternative considered and not preferred: lazily `detect()` inside `refresh()` — it hides topology-probe cost on a hot path and can mask a wedged link.

## Why it matters — the stress case

This is effectively invisible on a clean bus (detect almost always completes first try) and fires repeatedly on a contended/flaky one — i.e. exactly when graceful degradation matters most. It surfaced on a tester's deficient RS485 setup, whose logs over a ~5h window showed the library's existing guards all working as intended:

- CRC-rejection (#255) skipping zero-filled `wire=0x0000` frame tails on 0x01/0x03/0x11
- the sub-bus splice guard (#256) holding last-good on a constant-register change
- timeouts mid-flight ("Timeout awaiting …Response after 2 tries", "writer is closing") and full-poll loss ("All 3 register reads failed; treating plant as unreachable")

…and it was the reconnect path after full-poll loss where a `detect()` timeout produced the `PlantNotDetected` storm. The root cause is physical (contended bus / marginal wiring / flaky dongle), not something the library fixes — but the half-initialised window opening often is the realistic case, and a consumer shouldn't have to hand-roll the cleanup.

## Secondary (docs)

Worth stating the "success ≠ fresh" contract explicitly: the keep-last-good guard reports a *successful* poll while serving frozen data, so any client displaying device data should gate on `register_age()`. Closely related to the AIO register-getter ask — without it, that one device family can't be gated at all.

---

Reported via the cross-repo coordination note from the hass agent.

dewet22:
## Summary

Closes #274. Makes `connect()` + `detect()` atomic: a `detect()` that fails on a connection-level error now tears the connection down, instead of leaving the client half-initialised with `connected=True` / `capabilities=None`.

The footgun: a `detect()` timing out *after* `connect()` opened the socket left `connected=True` but `capabilities=None`. The natural consumer idiom — "reconnect if `not connected`, otherwise `refresh()`" — then skipped reconnecting and called `refresh()`, which raised `PlantNotDetected`. On a flaky/contended RS485 bus this fired repeatedly. `givenergy-hass` worked around it by discarding the client on any connect-failure, but every consumer hits the same wire.

## What changed

- **`detect()` now closes the connection on a connection-level failure** (`TimeoutError` / `CommunicationError`), so `connected` flips to `False` and the standard reconnect idiom recovers with no consumer-side guard.
- **`PlantTopologyMismatch` is exempt** — it's raised on a *healthy* connection (the registers read fine, only the `prior` hint was wrong), already clears `capabilities`, and is a deliberately caller-handled signal. The socket is left up so the caller can retry a cold `detect()`.
- The body moved into a private `_detect()`; `detect()` is now a thin try/except wrapper — the implementation is otherwise unchanged.
- **Docs:** the "success ≠ fresh" contract is now stated on `refresh()` / `load_config()` — the keep-last-good guards (CRC #255, splice #256, bank holds) report a *successful* poll while serving last-known-good data, so display consumers should gate on `Plant.register_age()` / `Plant.block_age()`.

## Test plan

- [ ] `uv run pytest tests/client/test_detect.py -v` — 4 new tests: timeout closes, missing-HR(0) closes, topology-mismatch keeps connection, success keeps connection
- [ ] `uv run pytest -q` — full suite (1079) green
- [ ] `uv run tox` — py311–py314 + lint + build green

Surfaced via a cross-repo report from the `givenergy-hass` side (hass#183), driven by a tester on a deficient RS485 bus where the #255/#256 guards were working as intended but the half-init window opened repeatedly.

### DIFF
--- a/givenergy_modbus/client/client.py
+++ b/givenergy_modbus/client/client.py
@@ -676,7 +676,40 @@ class Client:
         Uses a two-tier timeout: `timeout`/`retries` for the known inverter device
         (where a response is expected), and `probe_timeout`/`probe_retries` for
         speculative probes where absence is the common case.
+
+        On a connection-level failure (TimeoutError / CommunicationError) the
+        connection is torn down via close(), so connect()+detect() is atomic:
+        `connected` flips to False and the standard "reconnect if not connected"
+        idiom recovers (#274). A PlantTopologyMismatch is raised on a healthy
+        connection (only the hint was wrong) and leaves it up so the caller can
+        retry a cold detect().
         """
+        try:
+            return await self._detect(
+                timeout=timeout,
+                retries=retries,
+                probe_timeout=probe_timeout,
+                probe_retries=probe_retries,
+                prior=prior,
+            )
+        except PlantTopologyMismatch:
+            # Healthy connection — only the hint was wrong; capabilities already cleared.
+            raise
+        except (TimeoutError, CommunicationError):
+            # A connection-level failure leaves a half-open socket with capabilities
+            # unset. Tear down so connect()+detect() is atomic (#274).
+            await self.close()
+            raise
+
+    async def _detect(
+        self,
+        timeout: float,
+        retries: int,
+        probe_timeout: float,
+        probe_retries: int,
+        prior: PlantCapabilities | None,
+    ) -> PlantCapabilities:
+        """Implementation of detect(); see detect() for the contract and error semantics."""
         if prior is not None:
             _logger.info(
                 "detect: hinted mode — assuming device_type=Model.%s, inverter=0x%02x, "
@@ -887,6 +920,11 @@ class Client:
 
         Returns the populated plant on full success. On partial/total read
         failure raises ``RefreshPartiallySucceeded`` / ``RefreshFailed``.
+
+        Success does not imply *fresh*: the keep-last-good guards (CRC #255, sub-bus
+        splice #256, bank holds) report a successful poll while serving last-known-good
+        content for a device whose live read was rejected. Display consumers should gate
+        on ``Plant.register_age()`` / ``Plant.block_age()``, not on a poll returning.
         """
         caps = self.plant.capabilities
         if caps is None:
@@ -951,6 +989,11 @@ class Client:
 
         Returns the populated plant on full success. On partial/total read
         failure raises ``RefreshPartiallySucceeded`` / ``RefreshFailed``.
+
+        Success does not imply *fresh*: the keep-last-good guards (CRC #255, sub-bus
+        splice #256, bank holds) report a successful poll while serving last-known-good
+        content for a device whose live read was rejected. Display consumers should gate
+        on ``Plant.register_age()`` / ``Plant.block_age()``, not on a poll returning.
 
         The ``timeout=2.0, retries=1`` defaults are tuned for a contended bus: the
         inverter serialises requests, so when other clients (GivTCP, the vendor app,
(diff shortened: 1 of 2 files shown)


givenergy_modbus/client/client.py
```diff
@@ -676,7 +676,40 @@ async def detect(
...
+                probe_retries=probe_retries,
+                prior=prior,
+            )
+        except PlantTopologyMismatch:
+            # Healthy connection — only the hint was wrong; capabilities already cleared.
+            raise
+        except (TimeoutError, CommunicationError):
+            # A connection-level failure leaves a half-open socket with capabilities
+            # unset. Tear down so connect()+detect() is atomic (#274).
+            await self.close()
```
dewet22-codex: This makes `detect()` depend on `close()` completing successfully before the original connection-level error is re-raised. If `close()` itself raises during teardown (for example from `wait_closed()` or reader cleanup), callers will see that cleanup failure instead of the `TimeoutError` / `CommunicationError` that explains why detection failed. The important state flip already happens at the start of `close()`, so please guard/log cleanup failures here and re-raise the original detect failure unchanged.

dewet22-codex (changes requested):
Thanks, this is the right place to fix the half-initialized client state, and the topology-mismatch exception path looks well scoped. I found one teardown-path issue worth fixing before merge.

givenergy_modbus/client/client.py
dewet22: Fixed in 4dfe565 — wrapped the close() call in its own try/except that logs the teardown error, then re-raises the original TimeoutError / CommunicationError unchanged. Added a regression test (close() raising RuntimeError still surfaces the original TimeoutError). The connected=False flip happens at the start of close() regardless.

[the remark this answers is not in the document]
dewet22: Fixed in 4dfe565 — close() is now wrapped in its own try/except that logs the teardown failure, so the original TimeoutError / CommunicationError propagates unchanged. Added a regression test covering close() raising during teardown.

dewet22-codex (approved):
Re-reviewed the teardown masking fix. The guarded `close()` preserves the original detect failure while still attempting cleanup, and the regression test covers the masking case. No remaining actionable concerns from me.

PR #279, opened 2026-06-22
fix(plant): de-noise #256 splice-guard singleton path (thresholds + logging)

dewet22:
## Summary

Addresses [givenergy-hass#186](https://github.com/dewet22/givenergy-hass/issues/186): holdestmade (AC3 + 2× Giv-Bat 5.2) is seeing repeated WARNINGs from the #256 battery splice guard's **singleton-escrow** path — lone, near-threshold deltas like `IR(67) cell_mV 3327->3224` (−103 mV) and `IR(84) cap_centiAh 10200->9190` (−1010).

Analysis (corpus + `tests/debug/physics_delta_classify.py`) shows these are **benign physics, not corruption**: legitimate AC3 load-step cell sag (I×R) and a capacity recalibration, grazing thresholds that were calibrated on calmer Gen1 data. The escrow already handles them correctly (held one poll, then committed) — so no data is lost; the problem is a false alarm plus an observability gap (the hold logged WARNING but its resolution logged INFO / was silent, so a self-healing event read as unresolved at HA's default level).

## Changes

**1. Widen two thresholds (`battery_splice.py`):** `cell_mV` 100→150, `cap_centiAh` 1000→1500.

This is safe by construction: the corpus's **only** corruption signature is the temperature-zero cohort (≥2 temp registers → rejected outright). Temp thresholds are untouched, and no corruption has ever presented as a lone `cell_mV` or `cap_centiAh` delta — so widening exactly these two classes cannot weaken detection.

Corpus re-validated under the new thresholds:
- 1086 clean / 17 corruption — unchanged.
- **Every one of the 17 still rejected outright** (each trips ≥2 physics or an immutable violation).
- **Zero** corruption demoted to a lone `1 physics + 0 immutable` (escrow) shape; zero false trips.

**2. De-alarm the self-healing singleton path to INFO (`plant.py`):**
- The lone-delta **hold**, its **confirm-on-re-read**, and a **new snapback reversion** log all emit at INFO — reworded to drop "physics-impossible" / "Please report if seen".
- The ≥2 / immutable **REJECT** stays at WARNING with "Please report if seen" — that's the genuine corruption signal.

So a self-healing one-poll hold no longer alarms, and the held→resolved story is told at one consistent level.

## Test plan

- [ ] `uv run pytest tests/model/test_battery_splice.py tests/model/test_plant.py -v` — updated boundary/level assertions + new snapback-INFO test
- [ ] `uv run python tests/debug/physics_delta_classify.py tests/fixtures/captures/*/*.log` — 1086/17 separation, all 17 rejected
- [ ] `uv run pytest -q` (1085) and `uv run tox` (py311–py314) green

The thresholds are calibrated on holdestmade's three field data points plus physics reasoning — I'll ask for a fuller AC3 capture over a charge↔discharge cycle to tighten the calibration if the singleton path keeps firing.

## Release Notes

* **Bug Fixes**
  * Improved corruption detection accuracy for high-power battery systems by adjusting detection thresholds to reduce false positives from legitimate power fluctuations.
  * Enhanced diagnostic logging with clearer messages to distinguish between actual system issues and normal operational variations.
  * Refined escrow and revert handling to provide better visibility when temporary anomalies self-correct.

### DIFF
--- a/givenergy_modbus/model/battery_splice.py
+++ b/givenergy_modbus/model/battery_splice.py
@@ -16,13 +16,23 @@ This module is the single source of truth for those thresholds. ``Plant._commit_
 imports it to guard live commits; ``tests/debug/physics_delta_classify.py`` imports it to
 re-validate the corpus separation against the very same constants.
 
-Thresholds are in raw register units and set at roughly 10x what the quantity can really do
-in a ~30 s poll interval — cell voltages barely move (electrochemical inertia), cell-mass
+Thresholds are in raw register units, set comfortably above what the quantity can really do
+in a ~30 s poll interval — cell voltages move little near idle but sag under load, cell-mass
 temperatures drift, the junction-adjacent MOSFET sensor steps (hence its much wider
-threshold), capacities move with charge current (~0.5 Ah/poll at LV power). Status/bitfield
+threshold), capacities move with charge current and step on recalibration. Status/bitfield
 words carry no physics and are exempt (IR(91) legitimately toggles 0x0E10 <-> 0x0610 every
 few minutes). ``num_cells``, ``bms_firmware_version`` and the serial block are constant —
 ANY transient change is corruption on its own.
+
+The original Gen1-calibrated corpus separated cleanly, but field reports from higher-power
+hardware (hass#186, an AC3 + 2x Giv-Bat 5.2: a ~103 mV cell sag on a load step, a ~10 Ah
+capacity recalibration) showed legitimate per-poll deltas just past the ``cell_mV`` (100) and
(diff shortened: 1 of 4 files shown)


dewet22-codex (approved):
Looks good. The change keeps the hard rejection path intact for immutable/>=2-trip corruption while narrowing the de-noising to singleton escrow cases, and the updated tests cover the new threshold boundaries plus hold/confirm/snapback logging behavior. I also re-ran the focused splice tests and the corpus classifier; the corpus still separates as 1086 clean / 17 flagged transitions with no new singleton corruption shape.

Issue #281, opened 2026-06-23
Battery splice guard: a poisoned cold-start bms_firmware_version baseline permanently rejects a healthy bank (only a reboot recovers)


dewet22:
**Summary.** On a battery sub-bus, a single corrupt frame ingested at cold-start can poison the *immutable* `bms_firmware_version` (IR98) baseline. Because a lone immutable-register mismatch rejects the entire bank — and immutable mismatches are never escrowed/re-baselined — every subsequent *correct* frame is then rejected. The stale-baseline bypass can't recover (it keys off observation time, which refreshes on each rejected poll, not a rejection streak), so the pack stays held indefinitely until a comms reboot. This is the guard's "Please report if seen" — here's a sighting.

**Observed (field, modbus 2.5.1, via givenergy-hass v1.3.6).** A battery pack (device `0x33`) was unavailable for ~5 min. Per-poll, ~every 6–15 s:

```
Rejected battery bank for device 0x33 — sub-bus splice (constant-register change); keeping last-good.
Trips: IR(98) IMMUTABLE 2241->3009, IR(100) soc_pct 78->100, IR(88) cap_centiAh 58880->15088
```

- `IR(98)` = `bms_firmware_version`. Real value is **3009** (confirmed live after a dongle reboot; the sibling pack reads `3007`). `2241` is corrupt.
- Recorder history showed `bms_firmware_version` came up as **2241** immediately after a dongle restart — i.e. a bad *first* frame was adopted as the cold-start baseline.
- The SoC (78→100) and capacity (→150.88 Ah) in the rejected frames were **correct** (sunny morning, recharging) — they were collateral of the bank-level reject.
- Only a power-cycle cleared it; the guard then re-baselined cleanly on `3009`.

**Root cause (confirmed in source).**
1. **Cold-start adopts the first frame unconditionally**, with no plausibility check on immutable registers (`plant.py:834-835`; `bms_firmware_version` has no min/max in `battery.py:68`; `IMMUTABLE = [37, 38, *range(50,55)]` in `battery_splice.py`, idx 38 → IR98). One bad frame poisons the inviolate baseline.
2. **A lone immutable mismatch rejects the whole bank** (`plant.py:856`, `if immut or len(phys) >= 2`) and — unlike a single physics-class delta (`plant.py:868-892`) — is **never escrowed/re-baselined**. So the correct `3009` can never replace `2241`.
[...]

PR #285, opened 2026-06-23
feat: per-device comms-quality counters on Plant (CRC, splice, retries)

### LINKED ISSUE #284: Expose per-device comms-quality counters on Plant (CRC, splice-guard, retries — noise-floor diagnostics)
**What I'm after.** A way to read a plant's comms-quality "noise floor" — how often CRC-failed responses get skipped, how often the battery splice guard trips, and how often reads need retrying — without grepping WARNING logs after the fact. givenergy-hass surfaces these as Home Assistant diagnostic counters; the CLI could report them too.

Today these are only logged (or not surfaced at all). None is queryable, and they don't reach a `refresh()` caller: the keep-last-good guards report a *successful* poll while skipping the bad frame internally, and a read that fails then succeeds on retry is also reported as a clean success. So a consumer can't see any of them.

## Proposed API

Four cumulative, per-device event counters on `Plant`, keyed by `device_address`:

- `crc_failure_count: dict[int, int]` — +1 each time a CRC-failed response is skipped
- `splice_reject_count: dict[int, int]` — +1 each time a battery bank is hard-rejected (keep-last-good)
- `splice_held_count: dict[int, int]` — +1 each time a bank is escrowed/held pending confirmation
- `retry_count: dict[int, int]` — +1 per retry attempt consumed (a re-request beyond the initial read)

## Semantics that matter to the consumer

- Counts that only ever increment — in particular `splice_held_count` should count *hold events*, not currently-held banks (please don't decrement on confirm/clear). I want "how often did this happen", not "is it happening now".
- `retry_count` is the early-warning one: a read that recovers on retry is otherwise invisible, so a creeping per-device retry rate is the best "link starting to struggle" signal. Whether you count attempts or retried-reads is your call — I just need it monotonic per device like the rest.
- In-memory / cumulative-since-`Plant`-construction is fine — I accumulate reset-aware deltas my side, so a reconnect that re-instantiates the Plant doesn't break monotonicity.
- Public attributes, initialised empty (`Field(default_factory=dict)`).

## Likely sites (orientation, not prescription)

CRC-skip branch (`model/plant.py` ~657), hard-reject path (~886), escrow/hold path (~917), and the per-read retry loop inside `Client.refresh()` for `retry_count`. Final shape is yours.

## Consumer side is ready and decoupled

The hass integration reads all four defensively (`getattr(plant, …, {})`), so it can merge and ship independently as a no-op, then light up once I bump the modbus floor to whatever release adds them — no coordination window beyond the version bump.

This came out of a field noise-floor sweep (≈4 CRC skips over 14h on a real plant — benign, the guards doing their job), where the only way to see it was grepping container logs. Standing counters make "what's normal for my plant" visible.

dewet22:
Closes #284.

## What

Four cumulative, per-device event counters on `Plant`, keyed by `device_address`, so a consumer can read a plant's comms "noise floor" without grepping WARNING logs after the fact:

- `crc_failure_count` — +1 each time a CRC-failed response is skipped
- `splice_reject_count` — +1 each time a battery bank is hard-rejected (keep-last-good)
- `splice_held_count` — +1 each time a bank is escrowed/held pending confirmation
- `retry_count` — +1 per consumed read retry, per device

Today these are log-only and invisible to a `refresh()` caller — the keep-last-good guards report a *successful* poll while skipping the bad frame, and a read that recovers on retry also looks clean — so a consumer can't see any of them.

## Design

- **Counter == log event.** Each increment sits next to the exact log line it mirrors (CRC-skip in `update()`; both "Rejected battery bank" WARNINGs including the #281 backstop; both "held one poll" INFOs including the #281 coherent scalar-immutable hold), so logs and counters always agree.
- **Monotonic, never decremented.** `splice_held_count` counts *hold events*, not currently-held banks.
- **`retry_count` is gated on `warn_timeout`.** Absent-device detect probes pass `warn_timeout=False` and time out by design, so they're excluded — only genuine reads/writes (where a response was expected) count toward the noise floor.
- **Public attributes**, `Field(default_factory=dict, exclude=True)` — live, gettable, but kept out of `model_dump()` like the sibling `register_block_updated_at` runtime bookkeeping. In-memory / cumulative-since-construction; the consumer tracks reset-aware deltas across reconnects.

## Consumer side

The givenergy-hass integration already reads all four defensively (`getattr(plant, …, {})`), so it ships as a no-op on current modbus and lights up once the floor bumps to whatever release adds these — no coordination window beyond the version bump.

## Tests

Per-counter coverage: CRC-skip increments per device; the hard-reject vs escrow-hold split; the #281 scalar-immutable coherent hold; a consumed retry bumps `retry_count`; and a `warn_timeout=False` probe does not. Full suite green, tox green across py311–py314.

## Release Notes

* **New Features**
  * Added per-device communication quality diagnostics to monitor device connection reliability, including failure and retry counters.

* **Improvements**
  * Refined retry logic for improved fault handling and recovery accuracy.

### DIFF
--- a/givenergy_modbus/client/client.py
+++ b/givenergy_modbus/client/client.py
@@ -1451,11 +1451,17 @@ class Client:
                     f"Timeout awaiting {expected_response} (future: {response_future}), "
                     f"attempting retry {tries} of {retries}"
                 )
-                if tries <= retries and retry_delay > 0:
-                    # Discard the orphaned future so a late response from this attempt
-                    # doesn't accidentally resolve into the next attempt's future.
-                    response_future.cancel()
-                    await asyncio.sleep(retry_delay)
+                if tries <= retries:
+                    # Count the consumed retry (#284), but only where a response was genuinely
+                    # expected — absent-device detect probes pass warn_timeout=False and their
+                    # expected timeouts shouldn't pollute the per-device retry noise floor.
+                    if warn_timeout:
+                        self.plant.record_retry(request.device_address)
+                    if retry_delay > 0:
+                        # Discard the orphaned future so a late response from this attempt
+                        # doesn't accidentally resolve into the next attempt's future.
+                        response_future.cancel()
+                        await asyncio.sleep(retry_delay)
                 continue
             response = response_future.result()
             if tries > 0:
@@ -1466,8 +1472,11 @@ class Client:
                 # Unlike the timeout path above, no response_future.cancel() is needed here:
                 # the future is already resolved (we just called .result()), so cancel() would
                 # be a no-op, and the next attempt overwrites expected_responses[hash] anyway.
-                if tries <= retries and retry_delay > 0:
-                    await asyncio.sleep(retry_delay)
+                if tries <= retries:
+                    if warn_timeout:  # count the consumed retry (#284); skip absent-device probes
+                        self.plant.record_retry(request.device_address)
+                    if retry_delay > 0:
+                        await asyncio.sleep(retry_delay)
                 continue
             return response
 
--- a/givenergy_modbus/model/plant.py
+++ b/givenergy_modbus/model/plant.py
@@ -530,6 +530,18 @@ class Plant(GivEnergyBaseModel):
     # model_dump(): it's ephemeral runtime bookkeeping, not part of the plant's dumpable state.
     register_block_updated_at: dict[tuple[int, str, int, int], datetime] = Field(default_factory=dict, exclude=True)
 
+    # Per-device comms-quality counters (#284): cumulative, monotonic event counts keyed by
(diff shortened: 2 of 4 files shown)


dewet22-codex:
I’ve started a review pass on this now.

dewet22-codex (approved):
I’ve reviewed this pass and have no blocking concerns. The four Plant counter names match the hass-side consumer contract, the counters are excluded from dumps while remaining public attributes, and the retry hook is scoped to consumed retries with `warn_timeout=False` probes excluded. The splice and CRC increments sit beside the corresponding log events, which keeps the observable counters aligned with the existing diagnostics.

I ran the focused client and plant tests locally: `uv run pytest tests/client/test_client.py tests/model/test_plant.py -q` passed with 154 tests.

PR #290, opened 2026-06-23
fix(mock): disambiguate EMS managed-inverter slot serials by slot index

### LINKED ISSUE #288: mock: extend distinct-serial disambiguation to EMS managed-inverter slots (follow-up to #283)
Follow-up to #283 (`_make_device_serials_distinct`). #283 gives replayed multi-instance devices distinct serials by re-tailing with the device's bus address — covering LV battery packs (`0x32–0x37`) and AIO modules (`0x50–0x53`), and explicitly scoping out meters and HV BMUs as different layouts. EMS managed inverters fall in that same gap.

## The case
On an EMS plant the managed inverters aren't separate-address devices — they're up to 4 sub-slots in the EMS controller's cache at `0x11` (serials at IR2066-2070 / 2071-2075 / 2076-2080 / 2081-2085, surfaced via `Ems.managed_inverters → InverterSummary.serial_number`). The fixtures redact every serial by zeroing the suffix while keeping the `CExxxxG` manufacture-week prefix, so two managed inverters from a paired install collapse to the same redacted serial.

## Observed
Driving `ems_2_inv_3_bat_a/ems_arm1036_*.log` through `MockPlant` + a real client (givenergy-modbus 2.5.3, which has #283):

```
controller serial: EMS2522000
managed_inverters: 2
  CE2231G000  status=2 power=1042 soc=31 temp=50.3
  CE2231G000  status=2 power=1469 soc=16 temp=50.3   <- distinct telemetry, same redacted serial
raw inverter_1_serial_number = CE2231G000
raw inverter_2_serial_number = CE2231G000
```

## Impact
Any consumer keying on the managed-inverter serial collapses the two into one. Concretely, givenergy-hass#199 now models each managed inverter as its own HA device keyed by serial; against this fixture (via the mock) it creates a single device and drops the second — so the fixture can't validate the multi-managed-inverter device tree, though real hardware (distinct suffixes) is fine.

## Ask
Extend `_make_device_serials_distinct` to also disambiguate colliding managed-inverter slots within the `0x11` EMS block — re-tail each colliding slot's serial by its **slot index** (they share the device, so address-based re-tailing doesn't apply), preserving the prefix and keeping it obviously synthetic, exactly as #283 does for packs/modules. In-memory mock caches only; never touch fixture bytes. Final shape yours.

Touch points: `givenergy_modbus/testing/mock_plant.py` `_make_device_serials_distinct` (~129); managed-inverter serial registers on the `Ems` model (IR2066-2085); `Ems.managed_inverters` is the consumer surface that should then return distinct serials.

Scope: mock/replay fidelity for EMS only — real hardware (distinct serial suffixes) is unaffected. The same shape will matter for gateway parallel-AIO captures later.

dewet22:
Closes #288. Follow-up to #283.

## The gap

#283's `_make_device_serials_distinct` gives replayed multi-instance devices distinct serials by re-tailing with the device's **bus address** — covering LV battery packs (`0x32–0x37`) and AIO modules (`0x50–0x53`). EMS managed inverters fall outside it: they aren't separate addresses but up to **4 sub-slots in the single `0x11` EMS cache** (`EmsRegisterGetter.inverter_1..4_serial_number`, IR2066-2085), surfaced via `Ems.managed_inverters`. Fixtures redact serials by zeroing the suffix, so two managed inverters from a paired install collapse to one serial (`CE…G000`) — and a serial-keyed consumer (givenergy-hass#199's per-managed-inverter HA devices) drops the second. Real hardware (distinct suffixes) is unaffected.

## The fix

A small EMS-slot pass appended to `_make_device_serials_distinct`, reusing the existing `_decode_serial`/`_encode_serial` helpers: where two managed-inverter slots in the `0x11` cache share a non-empty serial, re-tail each by its **slot index** (1..4), preserving the prefix and keeping it obviously synthetic — exactly as #283 does for packs/modules by address.

- **Collision-only + register-presence gated**: a non-EMS `0x11` cache has no IR2066+ managed-inverter serials, so every slot decodes to `None` → no-op. The EMS controller's own serial (and the LV/AIO passes) are untouched.
- Mock caches only; fixture bytes never modified.

## Tests

Unit (two slots collide → re-tailed `…G001`/`…G002`, a distinct third left alone, absent slots ignored) and the reported regression: the `ems_2_inv_3_bat_a` fixture through `MockPlant` now yields two **distinct** `Ems.managed_inverters` serials (was both `CE2231G000`). Full suite green, tox green across py311–py314.

## Release Notes

* **Tests**
  * Enhanced test coverage for serial number collision handling in test utilities.

* **Chores**
  * Refactored internal serial collision-detection logic in test fixtures for improved accuracy.

### DIFF
--- a/givenergy_modbus/testing/mock_plant.py
+++ b/givenergy_modbus/testing/mock_plant.py
@@ -174,9 +174,10 @@ def _make_device_serials_distinct(plant: Plant) -> None:
         (range(lv_start, 0x38), BatteryRegisterGetter),  # LV battery packs
         (range(0x50, 0x54), AioBatteryModuleRegisterGetter),  # AIO battery modules
     ):
+        # registers_of returns () for getters without a serial_number; _disambiguate_serials then
+        # decodes every member to None and no-ops, so we can iterate unconditionally (#247 contract).
         regs = getter.registers_of("serial_number")
-        if regs:
-            _disambiguate_serials([(caches[a], regs, f"{a:02x}") for a in addr_range if a in caches])
+        _disambiguate_serials([(caches[a], regs, f"{a:02x}") for a in addr_range if a in caches])
 
     # EMS managed-inverter slots (#288): up to 4 sub-slots in the single 0x11 EMS cache (not separate
     # device addresses), so address keying can't see them — re-tail by slot index instead. A non-EMS
--- a/givenergy_modbus/testing/mock_plant.py
+++ b/givenergy_modbus/testing/mock_plant.py
@@ -32,6 +32,7 @@ from typing import cast
 from givenergy_modbus.framer import ServerFramer
 from givenergy_modbus.model.aio_battery import AioBatteryModuleRegisterGetter
 from givenergy_modbus.model.battery import BatteryRegisterGetter
+from givenergy_modbus.model.ems import EmsRegisterGetter
 from givenergy_modbus.model.plant import Plant
 from givenergy_modbus.model.register import HR, IR, MR, Register, RegisterGetter
 from givenergy_modbus.model.register_cache import RegisterCache
@@ -135,9 +136,10 @@ def _make_device_serials_distinct(plant: Plant) -> None:
     keeping it obviously synthetic. Distinct or singular serials are left untouched; only the
     in-memory mock caches are changed, never the fixture bytes.
 
-    Only IR/``C.serial`` 5-register devices are handled (LV battery packs, AIO modules); meters
-    (MR/``C.string``) and HV BMUs (stride within a single 0x70 cache) use other layouts and are
-    out of scope.
+    Handles the IR/``C.serial`` 5-register devices: LV battery packs and AIO modules (re-tailed by
+    bus address), and EMS managed-inverter slots within the 0x11 cache (re-tailed by slot index,
+    since they share an address). Meters (MR/``C.string``) and HV BMUs (stride within a single 0x70
+    cache) use other layouts and are out of scope.
     """
     caches = plant.register_caches
     # Unified addressing (inverter at 0x11/0x31) puts LV battery pack #1 at 0x32; legacy bare-plant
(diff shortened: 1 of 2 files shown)


dewet22-codex:

givenergy_modbus/testing/mock_plant.py
```diff
@@ -162,6 +164,22 @@ def _make_device_serials_distinct(plant: Plant) -> None:
...
+    # managed-inverter serials) is a no-op; the EMS controller's own serial is never touched.
+    ems_cache = caches.get(0x11)
+    if ems_cache is not None:
+        slots = [(i, EmsRegisterGetter.registers_of(f"inverter_{i}_serial_number")) for i in range(1, 5)]
+        named = {i: s for i, regs in slots if regs and (s := _decode_serial(ems_cache, regs))}
+        shared = {s for s, n in Counter(named.values()).items() if n > 1}
+        for i, regs in slots:
+            slot_serial = named.get(i)
+            if slot_serial and slot_serial in shared and len(slot_serial) >= 2:  # only colliding slots
+                distinct = slot_serial[:-2] + f"{i:02d}"  # slot index (1..4), preserving the prefix
```
dewet22-codex: Red (blocking): the generated slot suffix can collide with a serial that was already distinct and is therefore left untouched. I reproduced slots `CE0000G000`, `CE0000G000`, `CE0000G001` becoming `CE0000G001`, `CE0000G002`, `CE0000G001`, so a serial-keyed consumer still drops one inverter. I think the rewrite needs to account for all occupied serials when choosing replacements (with a regression case for this mixed input), or consistently rewrite every populated slot when any collision is present.

dewet22: Good catch — this was a real latent bug, and it applies to the address pass too. Re-tailing only the colliding members could mint a suffix that happens to match an already-distinct member (e.g. `CE0000G000`, `CE0000G000`, `CE0000G001` → `…G001`, `…G002`, `…G001`, recreating the collision). Refactored to `_disambiguate_serials`, which rewrites **every** populated member in a group by its unique index suffix the moment any collision is detected — so the result is provably distinct rather than collision-by-collision. Both passes (address-keyed and EMS-slot) now share it. Added an adversarial regression test for each pass covering exactly this suffix-collision case. Fixed in a16a16d.

givenergy_modbus/testing/mock_plant.py
```diff
@@ -162,6 +164,22 @@ def _make_device_serials_distinct(plant: Plant) -> None:
...
                     caches[addr][reg] = value
 
+    # EMS managed-inverter slots (#288): up to 4 sub-slots in the single 0x11 EMS cache (not separate
+    # device addresses), so the address-keyed pass above can't see them. Re-tail collisions by slot
+    # index instead. Collision-only + register-presence gated, so a non-EMS 0x11 cache (no IR2066+
+    # managed-inverter serials) is a no-op; the EMS controller's own serial is never touched.
+    ems_cache = caches.get(0x11)
+    if ems_cache is not None:
+        slots = [(i, EmsRegisterGetter.registers_of(f"inverter_{i}_serial_number")) for i in range(1, 5)]
+        named = {i: s for i, regs in slots if regs and (s := _decode_serial(ems_cache, regs))}
```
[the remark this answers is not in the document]
dewet22: Fixed — `_decode_serial` now strips whitespace as well as nulls (`.replace("\x00", "").strip()`), matching the model's `strip("\x00 ")`. A space-padded empty EMS slot now reads as no serial rather than a whitespace string that the collision check would treat as a phantom inverter. Added a regression test with a space-padded empty slot 4 (a16a16d).

dewet22-codex:

dewet22-codex (approved):
I’ve re-reviewed the latest head and the blocking collision case is addressed. Rewriting every populated member when a group collides keeps both the EMS slot path and the address-based path unique, with regression coverage for the adversarial mixed-serial input and space-padded empty slots. The focused mock-plant integration tests pass locally (20 tests).

Issue #286, opened 2026-06-23
Splice guard (#281): over-eager scalar-immutable adoption commits sustained corruption + co-corrupted SOC


dewet22:
## Summary

#281 (shipped 2.5.2) turned the battery splice guard's scalar-immutable handling — `num_cells` (IR97) and `bms_firmware_version` (IR98) — from a hard-reject *canary* into a *recoverable* signal. Field data shows that the 6-poll backstop now **adopts** sustained-but-transient splice corruption: the BMS firmware sensor bounces between the real value and corrupt ones, and because the co-corrupted bank is committed, SOC briefly dips to implausible values (e.g. ~15 %) before self-healing.

## Diagnosis (data-backed)

Decoded from the HA recorder on my loft pack, cross-checked against a parallel **GivTCP** install (a different library, no splice guard, polling the same pack via dongle fan-out):

- The corruption is **real and wire-level** — GivTCP bounces the firmware value far *more* than we do (≈56 vs ≈18 changes/24 h) and dips SOC at the same times, plus extra dips we reject. So it's genuine windowed-splice corruption, not something the library invents.
- **#281 is a net win** — the pre-#281 failure mode (firmware stuck on a poisoned cold-start baseline, pack repeatedly unavailable for hours) is gone.
- **But** removing IR(98)'s hard-reject canary lost the *incidental* protection it gave co-corrupted physics: a "drift" bank (firmware + SOC both spliced) now reaches the backstop and commits the lot → the SOC dips. Pre-#281 the immutable trip dropped that whole bank.
- The clean discriminator is **time**: corruption reverts within ~3 min, while a genuinely poisoned baseline persists for *hours*.

## Fix direction

Hold every scalar-immutable change as last-good (which also holds the co-corrupted physics), and self-heal only after the same value has persisted **uninterrupted well beyond the max corruption dwell** — a long, consumer-tunable window. Corruption reverts before it (rejected); a genuine poison outlasts it (heals, no reboot). Staleness stays visible to consumers via `block_age` (grows during a hold) and the `splice_held_count` diagnostic.

The first-frame cold-start adoption (how a poisoned baseline forms in the first place) is a separate hardening item I'll track on its own.

PR #291, opened 2026-06-23
feat(plant): confirm cold-start battery baseline before adopting (#289)

### LINKED ISSUE #289: Splice guard: harden cold-start baseline adoption (don't trust the first frame unconditionally)
Follow-up to #286.

## The gap

When a battery device's register cache is empty — a genuine cold start, or the first poll after a comms gap / reconnect — the splice guard has no last-good to compare against, so it adopts the incoming bank **unconditionally** as the new baseline. If that first frame happens to be a sub-bus splice (corrupt), it *poisons* the baseline. Everything #281/#286 do afterwards is recovery from that poisoned state.

## Why it matters

The #286 fix (hold scalar-immutable disagreements, heal only after a long `splice_heal_seconds` insistence) recovers from a poisoned baseline without a reboot, but at the cost of holding stale/last-good data for the heal window. Closing the cold-start hole at the source would mean poisoned baselines largely don't form in the first place — which would let `splice_heal_seconds` be set *shorter* (faster recovery from the rare genuine poison) without re-opening the corruption-adoption window.

## Possible directions (not yet decided)

- **Confirm-the-first-frame**: don't baseline a fresh cold-start bank until a second poll reads it the same (one extra poll of "unknown" before the device's data appears).
- **Plausibility-gate the first frame**: reject an obviously-implausible first bank (e.g. internally-incoherent SOC vs cell voltages) before adopting.
- A combination, scoped to battery banks (the only place the splice guard runs).

Care needed not to regress the legitimate post-outage stale-bypass path (#256), which deliberately adopts the first post-gap bank.

dewet22:
Closes #289. Follow-up to #286.

## The hole

When a battery device's cache is empty — a genuine cold start — the splice guard had no last-good to compare against, so it adopted the first bank **unconditionally** (`plant.py`, `if not present: return True`). If that first frame was a sub-bus splice (#147/#256), it poisoned the baseline, and everything #281/#286 do afterwards is recovery from that poisoned state (holding last-good for the `splice_heal_seconds` window).

## The fix — confirm-the-first-frame

The first cold-start frame is now **held** pending a corroborating read, reusing the pure `classify_transition` classifier exactly like the existing physics escrow ("held one poll, committed only if the next reads the same"):

- **First frame** → held (cache untouched, device serves "unknown").
- **Next frame** → `classify_transition(pending, incoming)`. Agree (no trip) → corroborated → commit. Disagree → one of them is a splice → keep the **most recent** as the candidate and wait (recovers if the *first* frame was the splice).

A transient splice reads different garbage each poll, so it never corroborates and never becomes the baseline.

### Design boundary (deliberate)

This defends against *transient/windowed* splices — the real #256 signature. A **persistently**-identical wrong value corroborates itself and is still adopted — by design, because recovering that is the #286 heal's job. The two layers compose: #289 stops transient poison forming, #286 heals persistent poison. Closing the hole at the source should let `splice_heal_seconds` be set shorter later, without re-opening the adoption window.

### Scope

Scoped to the genuine cold-start (empty-cache) branch only. The #256 post-outage stale-bypass is left untouched, so the deliberate fast-readopt-after-outage behaviour isn't regressed.

## Cost & observability

~1 extra poll of "unknown" battery data at startup — safer than serving a poisoned SOC/firmware for the heal window. Surfaced via a new per-device `cold_start_held_count` counter (#284-style, a benign "initialising" signal distinct from `splice_held_count`), plus the existing `block_age()` / absent-register signals.

## Tests

6 new cold-start tests (held-pending, corroborate-and-adopt, transient-splice-never-adopted, most-recent-wins, stale-pending-reset, persistent-value-corroborates). A behaviour change ripples through the existing splice suite — a single battery feed no longer seeds the cache — so those tests now establish a baseline with two corroborating feeds via a new `_establish_baseline` helper that is post-condition-identical to the old single feed. The #286 heal tests prime their poisoned baseline the same way, which nicely demonstrates the #289/#286 layering.

Full suite (1149) green, tox py311–py314, ruff/mypy clean, new code fully patch-covered.

## Known limitation (noted, not bounded)

A pathological *alternating* clean/splice every single poll would hold indefinitely. It's corpus-unsupported (splices are ~1.5%, 1086/17/0), and "unknown" is safer than an adopted poison — bounding it would reintroduce a poison window, so I've left it unbounded and documented.

## Update (review)

Codex caught a real hole in the original "persistent poison → #286 recovers it" boundary: it only holds for *scalar-immutable* poison. A persistent **temp-zero corruption cohort** (IR76-79=0) corroborates itself at cold start and, if adopted, hard-rejects every healthy frame forever — physics-only poison the scalar heal can't recover. Making the ≥2-physics hard-reject healable would be symmetric with, and re-open, the #256 sustained-corruption hole, so instead the cold-start confirmation now **refuses to baseline the known corruption signature** (new `battery_splice.is_corruption_cohort`): a corroborated cohort is held (WARNING) until a healthy pair corroborates instead. Scalar-immutable poison still corroborates and is left to #286, by design. Fixed in 4e1c36a with a bad/bad→sustained-healthy regression + predicate unit tests.

## Release Notes

* **Bug Fixes**
  * Improved detection of corrupted battery temperature sensor data to prevent invalid readings from affecting system performance.
  * Enhanced cold-start battery baseline establishment with a confirmation process that requires consistent readings before adoption, reducing false data acceptance.

* **Tests**
  * Added test coverage for data corruption detection and cold-start baseline confirmation workflows.

### DIFF
--- a/givenergy_modbus/model/plant.py
+++ b/givenergy_modbus/model/plant.py
@@ -16,6 +16,7 @@ from givenergy_modbus.model.battery_splice import (
     STALE_BYPASS_SECONDS,
     THRESHOLD_BY_CLASS,
     classify_transition,
+    is_corruption_cohort,
 )
 from givenergy_modbus.model.devices import DeviceType, PlantDevice
 from givenergy_modbus.model.devices import Inverter as UnifiedInverter
@@ -575,8 +576,9 @@ class Plant(GivEnergyBaseModel):
     # ingestion time). A cold-start frame is not adopted until the next poll reads it the same (no
     # physics/immutable trip between the two), so a transient sub-bus splice (#256) — different
     # garbage each poll — never corroborates and never poisons the baseline. Most-recent-wins on a
-    # disagreement. A persistently-identical wrong value DOES corroborate; that is left to the #286
-    # heal, by design. Ephemeral; rebuilt with a fresh Plant on reconnect; not serialised.
+    # disagreement. A persistently-identical scalar-immutable poison corroborates and is left to the
+    # #286 heal; a corroborated temp-zero corruption cohort is refused outright (it would hard-reject
+    # all healthy data, which the scalar heal can't recover). Ephemeral; rebuilt on reconnect.
     _splice_pending_baseline: dict[int, tuple[list[int], datetime]] = PrivateAttr(default_factory=dict)
 
     # Splice-guard observation clock (#256): per battery device address, the ingestion time of the
@@ -1004,8 +1006,13 @@ class Plant(GivEnergyBaseModel):
         and adopted only once the next poll reads it the same — ``classify_transition`` finds no physics
         or immutable trip between the two. A transient splice reads different garbage each poll, so it
         never corroborates; on a disagreement the most-recent frame becomes the new candidate (so a
-        splice in the *first* frame is recovered from). A persistently-identical wrong value does
-        corroborate and is adopted — by design: recovering that is the #286 heal's job, not this guard's.
+        splice in the *first* frame is recovered from).
+
+        Two exceptions to "corroboration adopts": a frame in the temp-zero corruption cohort
+        (:func:`is_corruption_cohort`) is refused even when corroborated — adopting it would
+        hard-reject every healthy frame forever, an unrecoverable physics-only poison. A
+        persistently-identical *scalar*-immutable poison (IR97/IR98) does corroborate and is adopted,
+        because recovering that one is the #286 heal's job, not this guard's.
 
         Returns True to adopt (corroborated), False to keep holding last-good ("unknown").
         """
@@ -1016,6 +1023,18 @@ class Plant(GivEnergyBaseModel):
             pending = None
         if pending is not None:
             phys, immut = classify_transition(pending[0], frame, incoming_present)
+            if not phys and not immut and is_corruption_cohort(frame, incoming_present):
+                # Corroborated, but the frame IS the temp-zero corruption signature (#256/#289 review).
+                # Baselining it would be unrecoverable: every later healthy frame then trips >=2 physics
+                # and is hard-rejected forever, and the #286 heal only recovers scalar-immutable poison.
+                # Keep holding (cache untouched) — a healthy corroborated pair will baseline instead.
+                self._bump(self.cold_start_held_count, device_address)
+                _logger.warning(
+                    "Battery bank for device 0x%02x: cold-start frame corroborated but is the temp-zero "
+                    "corruption cohort; refusing to baseline it, holding for a healthy read. Please report if seen.",
+                    device_address,
+                )
+                return False
             if not phys and not immut:
                 self._splice_pending_baseline.pop(device_address, None)
                 _logger.info(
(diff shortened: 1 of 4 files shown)


dewet22-codex:

givenergy_modbus/model/plant.py
```diff
@@ -971,6 +993,49 @@ def _splice_guard(
...
+        Returns True to adopt (corroborated), False to keep holding last-good ("unknown").
+        """
+        pending = self._splice_pending_baseline.get(device_address)
+        # A pending older than the stale-bypass window straddles a genuine polling gap — don't
+        # corroborate across it; treat the incoming frame as a fresh first read.
+        if pending is not None and (now_ts - pending[1]).total_seconds() > STALE_BYPASS_SECONDS:
+            pending = None
+        if pending is not None:
+            phys, immut = classify_transition(pending[0], frame, incoming_present)
+            if not phys and not immut:
```
dewet22-codex, line 1038: Red (blocking): the stated persistent-poison recovery boundary does not hold for the observed physics-only splice shape. I reproduced two identical cold-start frames with IR76-79 all zero being corroborated here and committed; every subsequent healthy frame then has four physics trips and is hard-rejected indefinitely (`splice_reject_count` reached 20 while the cached temperatures remained zero). The #286 heal path only handles scalar immutable disagreements, so it cannot recover this baseline. I think corroborated physics-only poison needs a recovery route, or the cold-start confirmation needs to avoid adopting this candidate shape, with a regression covering bad/bad followed by sustained healthy reads.

dewet22: Good, sharp catch — you're right, and the docstring overclaimed. Two identical temp-zero frames corroborate, and adopting that baseline hard-rejects every healthy frame forever; the #286 heal only recovers scalar-immutable poison, so this physics-only poison was unrecoverable.

I deliberately avoided the recovery-route option: making the ≥2-physics hard-reject healable is symmetric with the #256 sustained-corruption case (baseline X / incoming Y, repeated, either direction) and would re-open that hole. Instead the cold-start confirmation now refuses to *baseline* the known corruption signature — a new absolute predicate `battery_splice.is_corruption_cohort` (≥2 cell-mass temps at zero) gates the adopt branch. A corroborated cohort is held (cache untouched, surfaced at WARNING) until a healthy pair corroborates instead. Scalar-immutable poison still corroborates and is left to #286, by design — docstrings corrected to say exactly that.

Regression added (`test_cold_start_corroborated_temp_zero_cohort_not_adopted`): bad/bad then sustained healthy reads adopts the healthy baseline with `splice_reject_count == {}`, plus unit tests for the predicate. Fixed in 4e1c36a.

dewet22-codex:

dewet22-codex (approved):
I’ve re-reviewed the latest head and the blocking cold-start poison case is addressed. The corroborated temp-zero cohort is now refused without weakening the established hard-reject path; replaying the bad/bad/healthy/healthy sequence leaves the bad pair uncommitted and adopts the healthy baseline with no rejects. The focused splice and plant suites pass locally (144 tests).

PR #292, opened 2026-06-23
fix(mock): disambiguate serials in main() + deprecate `python -m` serve path

dewet22:
## Problem

The `mock_plant` module docstring advertised `python -m givenergy_modbus.testing.mock_plant --capture …` as the way to serve a mock. That path has two defects:

1. **Double-import** — `testing/__init__.py` imports `mock_plant`, so running it as `-m` loads the module twice (once as `__main__`, once as the package member) and emits a `RuntimeWarning`.
2. **Collided serials (the real bug)** — `main()` built the `MockPlant` from `plant_from_capture()` directly and never called `_make_device_serials_distinct()` (which lives only in `MockPlant.from_capture()`). So the served mock collided multi-instance device serials — e.g. an EMS capture served both managed inverters as `CE…G000`, defeating #283/#288/#290.

Surfaced hass-side (after ~2h of debugging) while validating [givenergy-hass#199](https://github.com/dewet22/givenergy-hass/issues/199) (EMS managed-inverter devices) against the mock.

## Fix

- `main()` now runs `_make_device_serials_distinct()` after loading the capture, exactly like `MockPlant.from_capture()` — the scripted serve path serves distinct serials.
- Docstring points at `givenergy-cli mock-server` (the canonical serve tool) and `MockPlant.from_capture()` for in-process/tests; the `-m` invocation is marked **deprecated**.
- `__main__` emits a one-time deprecation warning, then runs anyway so existing scripts don't break.

## Test

A regression test drives `main()` with the server stubbed (spy on construction, no-op `start`/`serve_forever`) and asserts the built mock serves two **distinct** managed-inverter serials. Verified it fails without the fix.

Full suite (1153) green, tox py311–py314.

### DIFF
--- a/givenergy_modbus/testing/mock_plant.py
+++ b/givenergy_modbus/testing/mock_plant.py
@@ -15,9 +15,19 @@ Fidelity choices (Phase 1 — read-serving, static topology):
 
 Mutable writes and dynamic bus/device add-remove are a later phase.
 
-Run it directly to point another client at a capture::
+To serve a mock from a capture, use the CLI (the canonical serve tool)::
 
-    python -m givenergy_modbus.testing.mock_plant --capture path/to/plant.log --port 8899
+    givenergy-cli mock-server --capture path/to/plant.log   # see --help for --bind/--port
+
+In-process (tests, scripting), build one with :meth:`MockPlant.from_capture` — which also
+re-tails multi-instance devices to distinct serials (#283/#288/#290)::
+
+    mock = MockPlant.from_capture("path/to/plant.log")
+    host, port = await mock.start()
+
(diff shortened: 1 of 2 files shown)


dewet22-codex:

dewet22-codex (approved):
I’ve reviewed the change and have no blocking concerns. The legacy `main()` path now applies the same serial disambiguation invariant as `MockPlant.from_capture()`, while preserving its existing overrides and server behaviour; the documented `givenergy-cli mock-server` flags also match the current CLI implementation. The focused mock-plant integration suite passes locally (21 tests), and the deprecated module invocation still provides usable help with the migration warning.

PR #296, opened 2026-06-24
feat(client): enable battery_*_limit_ac (HR313/314) writes for single-phase AC (#295)

### LINKED ISSUE #295: Enable HR313/314 (battery_*_limit_ac) writes for single-phase AC — exclusion's #75 blocker has landed
Cross-repo from givenergy-hass [dewet22/givenergy-hass#188](https://github.com/dewet22/givenergy-hass/issues/188). A user preparing Predbat against the HA integration can't write the AC charge/discharge limits on a single-phase AC-coupled inverter:

```
InvalidPduState: HR(313) is not permitted for AC inverter
```

The integration exposes `set_battery_charge_limit_ac` (HR313) / `set_battery_discharge_limit_ac` (HR314) for AC-coupled inverters (`has_ac_config_block and not is_three_phase`), but `Client.one_shot_command`'s per-model write-safe guard rejects them for `Model.AC` — they're not in `_InverterCommands.WRITE_SAFE_REGISTERS`.

## The exclusion's stated blocker has landed

The exclusion comment holds 313/314 out of the universal set because `ThreePhaseInverter` remaps the read-backs to HR1110/1108 (read≠write on three-phase), "until [per-model command-register selection, #75] lands." But #75 is closed — `WRITE_SAFE_REGISTERS` is already selected per model (`_InverterCommands` / `_ThreePhaseCommands` / `_EmsCommands`). So the concern is now separable: single-phase AC uses HR313/314 for both read and write; only three-phase needs the remap. This mirrors **HR311** (EXPORT_PRIORITY), already in the single-phase set, annotated "confirmed writable via hass#52 portal observations."

## Field evidence (single-phase, `Model.AC`)

GivEnergy's own Android app ("Direct Control") exposes Registers 313/314 as "Inverter Charge Power Percentage" / "Inverter Discharge Power Percentage" (Battery category, 0–100%) in its writable **Control** tab — vendor-intended writable controls. On the reporter's *older* firmware the app shows "Failed to read register" (the registers aren't present there); newer firmware is expected to read them, at which point the control populates. So they're real and writable where the firmware supports them.

**Consumer-side note:** on the hass end I plan to gate these controls on register *readability* (the control-side of the skip-if-None we use for sensors), so enabling 313/314 in the write-safe set won't produce broken controls on firmware that lacks them — hass just won't offer the control there.

## Requested outcome (API boundary)

- Single-phase AC (`Model.AC`): accept `set_battery_charge_limit_ac(v)` / `set_battery_discharge_limit_ac(v)` — i.e. add HR313/314 to `_InverterCommands.WRITE_SAFE_REGISTERS`, to the same bar HR311 met.
- Three-phase AC: route to HR1110/1108 in `_ThreePhaseCommands` (separate, lower priority — no reporter yet).
- If you'd prefer a wire capture before flipping single-phase, the reporter (Predbat-motivated) can grab a `Capture debug frames` recording across a portal write.

## hass side

The entities already exist (gated on `has_ac_config_block`), so once HR313/314 are writable for `Model.AC` no hass change is needed. If you conclude single-phase AC shouldn't write them, I'll gate the entities off for `Model.AC` instead.

dewet22:
Closes #295 (cross-filed from [givenergy-hass#188](https://github.com/dewet22/givenergy-hass/issues/188) — a Predbat user can't write the AC charge/discharge limits on a single-phase AC inverter).

## What

Enable `set_battery_charge_limit_ac` (HR313) / `set_battery_discharge_limit_ac` (HR314) for single-phase AC-coupled inverters. The GE app's "Direct Control → Control" tab exposes these as writable "Inverter Charge/Discharge Power Percentage" (0–100%), but `Client.one_shot_command`'s model-aware guard rejected them — they sat out of `_InverterCommands.WRITE_SAFE_REGISTERS`.

## Why now

The exclusion comment held them out "until per-model command-register selection (#75) lands." **#75 is closed** — the write-safe sets are already per-model (`_InverterCommands` / `_ThreePhaseCommands` / `_EmsCommands`). Single-phase AC uses HR313/314 for both read and write (no remap), so it's now separable from three-phase. Same bar as HR311 (EXPORT_PRIORITY), already enabled on "app Control-tab confirmed writable" evidence.

## The three-phase nuance (handled in this change, not deferred)

The command builders write HR313/314 **unconditionally**, and `_ThreePhaseCommands.WRITE_SAFE_REGISTERS` is *derived from* `_InverterCommands` by subtraction. So adding 313/314 to the single-phase set would leak them onto three-phase via inheritance — where the read-backs remap to HR1110/1108, so a write would target the wrong register. This change therefore **also removes 313/314 from the three-phase set** in the same commit, keeping three-phase rejecting until the HR1110/1108 routing lands (a separate, lower-priority follow-up — no reporter yet).

## Changes

- `_InverterCommands.WRITE_SAFE_REGISTERS`: **+313/314**.
- `_ThreePhaseCommands.WRITE_SAFE_REGISTERS`: **−313/314** in the derivation.
- Refreshed the now-stale exclusion comment (it cited #75 as the blocker).
- `set_battery_charge_limit_ac` stays a `commands.*` function (not promoted to a model method — that would inherit onto `ThreePhaseInverter` and target the wrong register).

## Tests

- Single-phase AC `one_shot_command` accepts HR313/314 (the reporter's case).
- Three-phase still raises `InvalidPduState` for HR313/314.
- Updated the allowlist-membership and three-phase-exclusion invariant tests.

Value bounds are already enforced in the builders (`1 ≤ val ≤ 100`). The change was run through the register-safety reviewer — no hardware-safety concerns (no address changes, no MRO leak to three-phase, EMS unaffected). Full suite green, tox py311–py314.

## hass side

The entities already exist (gated on `has_ac_config_block`), and hass plans to gate the controls on register *readability* — so on firmware that doesn't expose 313/314, no broken control is offered. No hass change needed once this ships.

### DIFF
--- a/givenergy_modbus/client/commands.py
+++ b/givenergy_modbus/client/commands.py
@@ -815,17 +815,19 @@ class _InverterCommands:
     # Single-phase shape: contains HR(96/110/116) and single-phase slot pairs (94/95,
     # 31/32 charge; 56/57, 44/45 discharge). ThreePhaseInverter replaces these via
     # _ThreePhaseCommands.WRITE_SAFE_REGISTERS (defined below, overrides this per MRO).
-    # Also excludes 313/314 (BATTERY_*_LIMIT_AC): the *single-phase* AC charge/discharge
-    # limits, but ThreePhaseInverter remaps the read-backs to HR1110/1108, so read !=
-    # write on three-phase AC. A correct three-phase write needs per-model
-    # command-register selection (#75); until that lands they stay out of the
-    # universal write-safe set. Also excludes 318-320 (pause mode, firmware-gated),
+    # Includes 313/314 (BATTERY_*_LIMIT_AC) for single-phase AC (#295, GE app Control-tab
+    # confirmed writable, same bar as HR311). _ThreePhaseCommands removes them again: there
+    # the read-backs remap to HR1110/1108 (read != write), and the command builders don't
+    # route to those yet (the per-model command-register selection blocker, #75, has landed,
+    # so single-phase is now separable). Also excludes 318-320 (pause mode, firmware-gated),
     # 1078/1109/1111-1123 (native three-phase), and 2040/2062-2069 (EMS).
     WRITE_SAFE_REGISTERS: ClassVar[frozenset[int]] = frozenset(
         {
             20,  # ENABLE_CHARGE_TARGET
             27,  # BATTERY_POWER_MODE
             311,  # EXPORT_PRIORITY (AC-coupled; confirmed writable via hass#52 portal observations)
+            313,  # BATTERY_CHARGE_LIMIT_AC (single-phase AC; app Control-tab confirmed writable, #295)
+            314,  # BATTERY_DISCHARGE_LIMIT_AC (single-phase AC; app Control-tab confirmed writable, #295)
             29,  # SOC_FORCE_ADJUST
             31,
             32,  # CHARGE_SLOT_2
@@ -1052,6 +1054,7 @@ class _ThreePhaseCommands:
             - {94, 95, 31, 32}  # charge slots 1-2
             - {56, 57, 44, 45}  # discharge slots 1-2
             - {96, 110, 116}  # ENABLE_CHARGE, BATTERY_SOC_RESERVE, CHARGE_TARGET_SOC
+            - {313, 314}  # BATTERY_*_LIMIT_AC: three-phase remaps to HR1110/1108, not routed yet (#295)
         )
         | {
             1078,  # BATTERY_RESERVE_SOC (three-phase only)
--- a/givenergy_modbus/client/commands.py
+++ b/givenergy_modbus/client/commands.py
@@ -781,6 +781,19 @@ def set_mode_storage(
     return ret
 
 
+# HR(300-359) AC-output config-block writes that are gated on capabilities.has_ac_config_block
+# rather than the model-class allowlist — one_shot_command unions these into the safe set only when
+# the detected model exposes the block (Model.AC / All-in-One), never for a DC-coupled hybrid or an
+# undetected client (#295/#296 review). Currently just the AC charge/discharge limits; HR311/317
+# (also HR300-359) predate the gate and stay in the universal set pending a separate cleanup.
+_AC_CONFIG_WRITE_SAFE_REGISTERS: frozenset[int] = frozenset(
+    {
+        RegisterMap.BATTERY_CHARGE_LIMIT_AC,  # 313
+        RegisterMap.BATTERY_DISCHARGE_LIMIT_AC,  # 314
+    }
+)
+
+
 class _InverterCommands:
     """Commands every inverter supports, mixed into Inverter model classes.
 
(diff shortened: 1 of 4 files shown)


dewet22-codex:

givenergy_modbus/client/commands.py
```diff
@@ -815,17 +815,19 @@ def slot_map(self) -> SlotMap: ...
...
+    # route to those yet (the per-model command-register selection blocker, #75, has landed,
+    # so single-phase is now separable). Also excludes 318-320 (pause mode, firmware-gated),
     # 1078/1109/1111-1123 (native three-phase), and 2040/2062-2069 (EMS).
     WRITE_SAFE_REGISTERS: ClassVar[frozenset[int]] = frozenset(
         {
             20,  # ENABLE_CHARGE_TARGET
             27,  # BATTERY_POWER_MODE
             311,  # EXPORT_PRIORITY (AC-coupled; confirmed writable via hass#52 portal observations)
+            313,  # BATTERY_CHARGE_LIMIT_AC (single-phase AC; app Control-tab confirmed writable, #295)
+            314,  # BATTERY_DISCHARGE_LIMIT_AC (single-phase AC; app Control-tab confirmed writable, #295)
```
dewet22-codex: Red (blocking): adding these to the shared `_InverterCommands` set enables them for every non-three-phase, non-EMS model, not just single-phase AC. `one_shot_command()` selects this set for DC-coupled hybrids and even undetected clients, although `PlantCapabilities.has_ac_config_block` explicitly says those models lack HR300–359 and time out on reads. I think these registers need to remain outside the universal set and be added at the client boundary only when capabilities confirm `has_ac_config_block` and the model is not three-phase (which covers Model.AC and the confirmed AIO case); the undetected fallback should continue to reject them. A negative HYBRID/undetected regression would pin the safety boundary alongside the AC acceptance and three-phase rejection tests.

dewet22: Good catch — fixed. You're right that `_InverterCommands.WRITE_SAFE_REGISTERS` serves DC-coupled hybrids and the undetected fallback, neither of which exposes the HR(300-359) AC-output config block. HR313/314 are now gated on `capabilities.has_ac_config_block` at the client boundary: `one_shot_command` unions a new `_AC_CONFIG_WRITE_SAFE_REGISTERS` into the safe set only when `caps.has_ac_config_block and not caps.is_three_phase` — so Model.AC and the All-in-One accept them, while DC hybrids, three-phase (which remaps to HR1110/1108), EMS, and undetected clients all reject. Removed them from the class sets.

Added the regression coverage you asked for: parametrised reject tests for DC hybrid (HYBRID_GEN1/2/3), AC_3PH, HYBRID_3PH, and the undetected fallback, alongside the AC/AIO accept cases. Register-safety re-reviewed (the gate is strictly more conservative than the prior approach). Fixed in a63534f.

tests/client/test_client_write_safety.py
```diff
@@ -58,6 +58,29 @@ async def test_single_phase_rejects_three_phase_register():
...
+        await client.one_shot_command([WriteHoldingRegisterRequest(reg, 50)], dry_run=True)  # must not raise
+
+
+@pytest.mark.asyncio
+async def test_three_phase_rejects_battery_limit_ac_writes():
+    """Three-phase must still reject HR313/314 — it remaps to HR1110/1108, not yet routed (#295)."""
+    client = _client(_caps(Model.HYBRID_3PH))
+    for reg in _AC_LIMIT_REGS:
+        with pytest.raises(InvalidPduState, match=rf"HR\({reg}\)"):
+            await client.one_shot_command([WriteHoldingRegisterRequest(reg, 50)], dry_run=True)
```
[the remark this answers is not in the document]
dewet22: Added — `test_battery_limit_ac_commands_encode` invokes `encode()` on every request from `set_battery_charge_limit_ac(50)` / `set_battery_discharge_limit_ac(50)`, so the builders are exercised end-to-end through the PDU encode path (a63534f).

dewet22-codex:

dewet22-codex (approved):
I’ve re-reviewed the latest head and the model-scope concern is resolved. HR313/314 now remain outside the universal and three-phase sets, and the client adds them only for detected, non-three-phase plants whose capabilities confirm the AC config block. AC and AIO accept them; hybrid, three-phase, EMS and undetected clients reject them. The focused write-safety suite passes (251 tests), as does the full suite (1,160 tests). I have no blocking concerns remaining.

Issue #299, opened 2026-06-24
Splice guard: hard-reject has no recovery for a sustained legitimate cell-voltage step (heal-adopt / plausibility follow-up to #300)


dewet22:
Field finding (maintainer's loft Gen1, 2026-06-24). The splice guard hard-rejected a battery bank **continuously for ~18.5 minutes** (device 0x33, 15:03:34 → 15:22:05), holding last-good and freezing the battery telemetry — but the rejected data was **legitimate**, not corruption.

## Evidence

Trips throughout the run:
```
IR(65/68) cell_mV  3453 -> 3607..3644
IR(82)    v_out_mV 53901 -> 55926..56224
```
- The values are **plausible** (3.6 V cells, 56 V v_out) and **drift smoothly** over the 18 min (v_out 56224→56164→56045→55985→55956→55926; cells 3644→3631→3622→3614→3607) — a real settling voltage, not a splice (corruption doesn't decay smoothly).
- The held baseline stays pinned at exactly `3453 / 53901` the whole time, because every frame is rejected and the cache never catches up.
- v_out rose ~2263 mV ≈ a ~150 mV rise across ~15 cells — internally consistent with a genuine all-cell surge.
- Cell-voltage chart confirms the cells climbing toward the 3.5 V warn band as the pack tops out, with telemetry thinning through the rejection window.

## Root cause

As the pack approaches 100 % SOC, the LiFePO4 voltage **knee** makes cell voltage rise steeply — legitimately exceeding the `cell_mV` per-poll threshold (currently **150**, observed 154–198 mV) and `v_out_mV` (currently **2000**, observed 2263 mV). Two physics trips → the **hard-reject** path, which — unlike #286's scalar-immutable hold-then-heal — has **no adopt/recovery for a sustained *legitimate* step**, so it stays stuck rejecting plausible data until the voltage settles. The maintainer reports these spikes recur regularly near full charge.

## Fix (chosen: widen)

[...]

Issue #298, opened 2026-06-24
Splice guard: sub-bank salvage — commit good registers instead of rejecting the whole bank on a noisy bus


dewet22:
Enhancement / design discussion. On a noisy sub-bus, the splice guard (#256) rejects the **whole** battery bank (IR60-119) when ≥2 physics-impossible deltas trip. When corruption is frequent, a bank that carries **both** a legitimate update *and* a corruption gets thrown out wholesale — so the legit fields are held at last-good too, and telemetry lags reality.

## Field evidence (maintainer's loft Gen1, 2026-06-24)

A genuinely noisy bus: **159 rejections in ~3h** across both packs (0x32/0x33), trips like:
```
0x32 — IR(76) cell_temp 303->0, IR(77) 283->0, IR(78) 292->0, IR(79) 275->0, IR(81) mosfet 322->0
0x33 — IR(76) 281->58, IR(81) mosfet 331->0
0x32 — IR(103) 304->0, IR(104) 283->11
```
These are healthy→zero (real incoming temp-zero splices; the guard is working — the cache holds *good* last-good, not poison). But because rejection is whole-bank, the battery telemetry lagged a GivTCP control on the same packs by ~3 °C (modbus 30.0 vs GivTCP 33.5, climbing) — legit temp/SOC updates riding in the same frames as a corruption were discarded along with the corruption.

## Proposal

Sub-bank salvage: commit the registers that didn't trip (and aren't in the same atomic group as a tripped register), holding only the corrupt ones at last-good — so a temp-zero splice holds the temps but lets a clean SOC / cell-voltage update through, cutting the lag.

## Why it's non-trivial (and why #256 chose whole-bank)

- **Atomic groups:** uint32 pairs (v_out, cap_centiAh) and physics cohorts must commit together; salvage granularity has to respect them.
- **Windowed corruption:** #256 found splices are *contiguous windows* — a register that looks clean but sits *inside* the splice window may be silently wrong. Committing it would defeat the guard. Safe salvage needs to bound the splice window, which we can't currently do reliably.
- **Trade-off inversion:** whole-bank rejection is the safe-but-blunt choice (deliberate in #256); salvage trades less lag for more risk of admitting subtle corruption. Needs corpus validation that salvaged registers are provably never corrupt.

## Lighter alternative

[...]

PR #300, opened 2026-06-24
fix(splice): widen cell_mV/v_out thresholds for the near-full-SOC knee (#299)

dewet22:
Refs #299 — this is the threshold-widening step. The deeper sustained-step recovery fix stays tracked on #299 as follow-up (per review), so this PR deliberately does **not** auto-close it.

As a LiFePO4 pack enters the charge knee near 100% SOC, all cells surge together and the pack terminal voltage steps with them. On my loft Gen1 this tripped the splice guard's physics-delta thresholds (`cell_mV` 150, `v_out` 2000) on ≥2 registers at once, so the whole bank hit the **hard-reject** path and stayed rejected for the entire ~18.5 min surge (device 0x33), freezing the battery telemetry at the last-good baseline. The rejected data was legitimate — plausible 3.6 V cells / 56 V v_out, drifting smoothly as the pack settled — not corruption.

## Change

- `cell_mV` 150 → 300, `v_out_mV` 2000 → 4000 in `battery_splice.py` (`SCALAR_RULES` / `PAIR_RULES`), clearing the observed knee (cells 154–198 mV/poll, v_out 2263 mV/poll) with margin.
- Module docstring + rule comments updated to fold the #299 knee rationale in alongside the hass#186 cell_mV widening.

## Why it's safe

The corpus's only corruption signature is the temperature-zero cohort (≥2 cell-mass temps → 0), which is untouched. No corruption has ever presented as a lone `cell_mV` or `v_out` delta, so widening exactly these voltage classes can't weaken detection — the same argument as the earlier hass#186 `cell_mV` 100→150 widening. Re-validated against the capture corpus: separation **unchanged** at 1086 clean / 1 genuine singleton / 16 corruption events (all ≥2 trips). Codex independently re-ran the classifier and confirmed the same.

## Tests

- New `test_near_full_soc_knee_surge_commits` reproduces the field signature (two cells +198 mV, v_out +2263 mV in one poll) and asserts it commits rather than rejecting.
- The existing escrow/classifier tests that used ~200 mV cell deltas (chosen to be "just over 150") are bumped to clearly exceed 300 — a mechanical consequence of the threshold move, behaviour intent unchanged.
- Full suite green; `tox` py311–py314 + format + lint green.

## Not in this PR

The deeper fragility #299 notes: a ≥2-physics hard-reject still has no recovery for a *sustained legitimate* step, so a larger future spike could stick again. The heal-adopt / plausibility-gate follow-up (extending #286's time-discriminator, or the #289-deferred plausibility gate) is left as the proper fix and stays open on #299; this widening is the pragmatic first step. Related: #298 (sub-bank salvage / telemetry-lag).

### DIFF
--- a/givenergy_modbus/model/battery_splice.py
+++ b/givenergy_modbus/model/battery_splice.py
@@ -25,14 +25,17 @@ few minutes). ``num_cells``, ``bms_firmware_version`` and the serial block are c
 ANY transient change is corruption on its own.
 
 The original Gen1-calibrated corpus separated cleanly, but field reports from higher-power
-hardware (hass#186, an AC3 + 2x Giv-Bat 5.2: a ~103 mV cell sag on a load step, a ~10 Ah
-capacity recalibration) showed legitimate per-poll deltas just past the ``cell_mV`` (100) and
-``cap_centiAh`` (1000) thresholds — caught by the singleton escrow but noisy. Both were widened
-(cell_mV->150, cap_centiAh->1500). This is safe because the corpus's *only* corruption signature
-is the temperature-zero cohort (>=2 temp registers -> rejected outright): no corruption has ever
-presented as a lone ``cell_mV`` or ``cap_centiAh`` delta, so widening exactly these two classes
-(temps untouched) cannot weaken detection, and any residual over-threshold singleton still
-escrows.
+hardware showed legitimate per-poll deltas exceeding the voltage/capacity thresholds. hass#186
+(an AC3 + 2x Giv-Bat 5.2) saw a ~103 mV cell sag on a load step and a ~10 Ah capacity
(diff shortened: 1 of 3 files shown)


dewet22-codex:
I’m starting a review of the latest changes now.

dewet22-codex (approved):
I’ve completed the review and have no blocking concerns. The field regression is internally coherent: the two +198 mV cell steps, +2.263 V pack step, and matching cell-sum movement now classify cleanly and commit directly. I re-ran the production corpus classifier against the amended constants: all 1,103 transitions retain the stated separation (1,086 clean, one genuine singleton, 16 corruption transitions with at least two total trips). The focused splice/Plant suite (145 tests) and full suite with coverage (1,161 tests) pass locally. All CI checks are green. The larger sustained-step recovery gap remains a clearly documented follow-up rather than being obscured by this threshold change.

PR #301, opened 2026-06-24
feat(commands): widen battery power-limit bounds to 0–100% (HR111/112, HR313/314)

dewet22:
Reported via givenergy-hass (Nick, EMS tester): `set_battery_charge_limit` / `set_battery_discharge_limit` (HR111/112) reject >50%, while the GivEnergy app exposes "Battery Charge/Discharge Power" as 0–100%.

## Investigation — is the 50% real?

It looked like it might be a genuine hardware limit, not conservatism:
- **GivTCP** (downstream fork, large install base) caps HR111/112 at `[0,50]` with the comment *"50% (2.6 kW) is the maximum for most inverters"*, and routes AC-coupled/3-phase to HR313/314 `[1,100]`.
- **All four capture fixtures** read the DC limit (HR111/112) at exactly 50; my hybrid's battery is rated 2600 W of a 5000 W inverter ≈ 52% — so on a DC-coupled hybrid the battery subsystem is ~half the inverter rating.

## What changed it

I tested on my own Gen1 hybrid: the GivEnergy app wrote **62% to HR111** and **59% to HR112**, both accepted. So the register tolerates >50 — GivTCP's 50 is a software clamp to the *effective* max, and it's model-specific (a hard `[0,50]` would wrongly reject valid higher values on a hybrid with a larger battery subsystem).

These registers are a percentage-of-rated **ceiling**, not a current command — the BMS/firmware clamp to the battery's actual rating — so a too-high value is clamped (ineffective), not unsafe. Register-safety reviewed: the library only ever writes the percentage; nothing downstream turns it into a current/power command; gating and encoding are unchanged.

## Change

Widen all four setters to `[0,100]`:
- HR111/112 (DC): `[0,50]` → `[0,100]`
- HR313/314 (AC): `[1,100]` → `[0,100]` (0 = disable AC charge/discharge, matching the DC pair and GivTCP's own model `valid=(0,100)`)

A thorough docstring records the provenance so the bound isn't re-capped at 50.

## Tests

Updated `tests/client/test_commands.py`: 51/100 now commit on the DC pair, 0/100 on the AC pair, 101 is the new rejection. Full suite + tox (py311–py314) green.

## Pending / notes (not blocking)

- A physical clamp test on my hybrid (set discharge 50%, start an EV charge, bump past 50%, watch whether output rises past ~2.5 kW or clamps) — characterises whether >50% does anything on this unit; result folds back into the tracking issue.
- Pre-existing #75 (three-phase HR313/314 → HR1110/1108 read remap) is untouched.

Durable tracking issue to follow (hass asked us to capture this on the modbus side).

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -265,8 +265,8 @@ An unmarked row means the method is available on both inverter types. On `ThreeP
 | `set_charge_target_soc(soc)` | Set just the charge-target SOC (4–100), leaving the enable bits untouched | |
 | `disable_charge_target()` | Remove SOC limit, target 100% | |
 | `set_enable_charge(enabled)` | Enable or disable charging | |
-| `set_battery_charge_limit(val)` | Charge power limit (0–50%) | |
-| `set_battery_charge_limit_ac(val)` | AC charge power limit (1–100%) | ⛔ commands-only |
+| `set_battery_charge_limit(val)` | Charge power limit (0–100%; firmware clamps per-model) | |
+| `set_battery_charge_limit_ac(val)` | AC charge power limit (0–100%) | ⛔ commands-only |
 | `set_shallow_charge(val)` | Set shallow charge threshold (deprecated — use `set_battery_soc_reserve`) | ⛔ commands-only |
 
 ### Discharging
@@ -274,8 +274,8 @@ An unmarked row means the method is available on both inverter types. On `ThreeP
 | Function | Description | Surface |
 |---|---|---|
 | `set_enable_discharge(enabled)` | Enable or disable discharging | |
-| `set_battery_discharge_limit(val)` | Discharge power limit (0–50%) | |
-| `set_battery_discharge_limit_ac(val)` | AC discharge power limit (1–100%) | ⛔ commands-only |
+| `set_battery_discharge_limit(val)` | Discharge power limit (0–100%; firmware clamps per-model) | |
+| `set_battery_discharge_limit_ac(val)` | AC discharge power limit (0–100%) | ⛔ commands-only |
 | `set_battery_soc_reserve(val)` | Minimum SOC to maintain (4–100%) | |
 | `set_battery_power_reserve(val)` | Battery power reserve (4–100%) | |
 
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -265,7 +265,7 @@ An unmarked row means the method is available on both inverter types. On `ThreeP
 | `set_charge_target_soc(soc)` | Set just the charge-target SOC (4–100), leaving the enable bits untouched | |
 | `disable_charge_target()` | Remove SOC limit, target 100% | |
 | `set_enable_charge(enabled)` | Enable or disable charging | |
-| `set_battery_charge_limit(val)` | Charge power limit (0–100%; firmware clamps per-model) | |
+| `set_battery_charge_limit(val)` | Charge power limit (0–100%; firmware expected to clamp per-model) | |
 | `set_battery_charge_limit_ac(val)` | AC charge power limit (0–100%) | ⛔ commands-only |
 | `set_shallow_charge(val)` | Set shallow charge threshold (deprecated — use `set_battery_soc_reserve`) | ⛔ commands-only |
 
@@ -274,7 +274,7 @@ An unmarked row means the method is available on both inverter types. On `ThreeP
 | Function | Description | Surface |
 |---|---|---|
 | `set_enable_discharge(enabled)` | Enable or disable discharging | |
-| `set_battery_discharge_limit(val)` | Discharge power limit (0–100%; firmware clamps per-model) | |
+| `set_battery_discharge_limit(val)` | Discharge power limit (0–100%; firmware expected to clamp per-model) | |
 | `set_battery_discharge_limit_ac(val)` | AC discharge power limit (0–100%) | ⛔ commands-only |
 | `set_battery_soc_reserve(val)` | Minimum SOC to maintain (4–100%) | |
 | `set_battery_power_reserve(val)` | Battery power reserve (4–100%) | |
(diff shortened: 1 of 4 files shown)


dewet22-codex:
I’m starting a review of the latest changes now.

givenergy_modbus/client/commands.py
```diff
@@ -394,18 +406,18 @@ def set_enable_eps(enabled: bool) -> list[TransparentRequest]:
 
 
 def set_battery_charge_limit_ac(val: int) -> list[TransparentRequest]:
-    """Set the battery AC charge power limit as a percentage."""
+    """Set the battery AC charge power limit as a percentage (0-100; 0 disables AC charge)."""
     val = _as_int(val, "val")
-    if not 1 <= val <= 100:
-        raise ValueError(f"Specified AC Charge Limit ({val}%) is not in [1-100]%")
+    if not 0 <= val <= 100:
```
dewet22-codex, line 417: Red (blocking): this lower-bound change is not supported by the evidence in the PR. The v4.1.6 register inventory documents HR313/314 as 1–100, the field observation here was against HR111/112, and GivTCP’s `valid=(0, 100)` only shows what its model accepts; it does not establish that writing zero disables AC charge or discharge. PR #296 deliberately retained the 1–100 command bound. I don’t think we should add a previously rejected hardware write or document its semantics without a portal/app capture showing zero being written and read back with the intended effect. Keeping the AC pair at 1–100 would let the supported DC widening ship independently.

givenergy_modbus/client/commands.py
```diff
@@ -327,18 +327,30 @@ def set_charge_target_soc_3ph(target_soc: int) -> list[TransparentRequest]:
 
 
 def set_battery_charge_limit(val: int) -> list[TransparentRequest]:
-    """Set the battery charge power limit as a percentage of rated charge power (0–50)."""
+    """Set the battery charge power limit as a percentage of rated charge power (0-100).
+
+    This is a %-of-rated ceiling, not a current command: the BMS/firmware clamp it to the battery
```
dewet22-codex: Amber: the PR body says the physical clamp test is still pending, so “the BMS/firmware clamp it” is stronger than the current evidence. The app observation proves that this inverter accepts values above 50; it does not yet establish what happens when the requested ceiling exceeds the battery subsystem’s capability. I’d keep the useful observed provenance here, but describe the clamp behaviour as unverified until that test has been run.

dewet22-codex (changes requested):
I found one blocking hardware-write contract issue: the observed HR111/112 writes support widening the DC pair, but they do not support changing HR313/314 from the documented 1–100 range or asserting that zero disables those controls. I’ve left that detail inline, along with a smaller documentation caveat around the still-untested firmware-clamp claim. The public usage table also still lists the old 0–50 and 1–100 ranges, so I think it needs updating alongside this API change. The implementation otherwise stays within the existing write-safety routing; the full suite passes (1,161 tests), lint is clean, and CI is green.

givenergy_modbus/client/commands.py
dewet22: Maintainer decision: keeping the AC 0-floor to field-confirm rather than revert. The GE Android app exposes HR313/314 as 0–100; the v4.1.6 inventory (1–100) is hybrid-scoped and has had errata, and HR313/314 are AC-coupled registers absent on hybrids, so it isn't authoritative for the AC pair. The docstring now marks the 0-disables semantics as expected-but-unconfirmed, and a tester on AC/EMS hardware (via givenergy-hass) will write 0 and confirm it stops AC charge/discharge. The DC widen is independently evidenced and unaffected.

givenergy_modbus/client/commands.py
dewet22: Addressed in 4d0ccc7: the docstring no longer states the firmware clamps as established — it reads that the app field-test proves the register accepts >50, while clamp behaviour above the battery's capability is the expectation pending the physical test (tooling added at `tests/debug/probe_discharge_limit_clamp.py`).

dewet22-codex:

docs/usage.md
```diff
@@ -265,17 +265,17 @@ An unmarked row means the method is available on both inverter types. On `ThreeP
 | `set_charge_target_soc(soc)` | Set just the charge-target SOC (4–100), leaving the enable bits untouched | |
 | `disable_charge_target()` | Remove SOC limit, target 100% | |
 | `set_enable_charge(enabled)` | Enable or disable charging | |
-| `set_battery_charge_limit(val)` | Charge power limit (0–50%) | |
-| `set_battery_charge_limit_ac(val)` | AC charge power limit (1–100%) | ⛔ commands-only |
+| `set_battery_charge_limit(val)` | Charge power limit (0–100%; firmware clamps per-model) | |
```
dewet22-codex: Amber (non-blocking): this table still says the firmware clamps per model as an established fact, while the updated command docstring correctly records that behaviour as expected but not yet field-verified. I’d use the same caveat here when next touching the branch so the public surfaces agree.

dewet22-codex (approved):
I’ve re-reviewed the latest head and the earlier blocking concern is now an explicit maintainer decision with the uncertainty recorded in the command API. The DC widening remains backed by the observed app writes; the AC zero floor is clearly marked as awaiting field confirmation rather than claiming proven disable semantics. The new live probe is model-gated, prompt-gated and restores the original limit through its failure path. I left one non-blocking consistency note on the usage-table wording. Focused command tests pass (133), lint is clean, and CI is green. I have no blocking concerns remaining.

docs/usage.md
dewet22: Addressed in c68c92b: the usage table now reads "firmware expected to clamp per-model", matching the docstring's expected-but-unverified framing — public surfaces agree.

Issue #302, opened 2026-06-24
Widen battery charge/discharge power-limit bounds to 0–100% (HR111/112, HR313/314)


dewet22:
Reported via givenergy-hass (a tester on an EMS plant): `set_battery_charge_limit` / `set_battery_discharge_limit` (HR111/112) reject anything above 50%, but the GivEnergy app exposes "Battery Charge/Discharge Power" as 0–100%.

## Is the 50% a real hardware limit?

At first it looked like it might be, not just conservatism:

- **GivTCP** (a downstream fork with a large install base) caps HR111/112 at `[0,50]` with the comment *"50% (2.6 kW) is the maximum for most inverters"*, and routes AC-coupled/3-phase systems to HR313/314 `[1,100]`.
- **All four capture fixtures** read the DC limit (HR111/112) at exactly 50, and my hybrid's battery is rated 2600 W of a 5000 W inverter ≈ 52% — so on a DC-coupled hybrid the battery subsystem is ~half the inverter's AC rating.

## What changed my mind

I wrote 62% to HR111 and 59% to HR112 from the GivEnergy app on my own Gen1 hybrid, and the inverter accepted both. So the register tolerates >50 — GivTCP's 50 is a software clamp to the *effective* max, not a register-level limit, and the ~50% is model-specific (a hard `[0,50]` wrongly rejects valid higher values on a hybrid with a larger battery subsystem).

These registers are a percentage-of-rated *ceiling*, not a current command — the BMS/firmware enforce the real physical limits — so a too-high value is expected to be clamped (ineffective), not unsafe. (Register-safety reviewed: the library only ever writes the percentage; nothing downstream turns it into a current/power command.)

## Change (#301)

Widen all four setters to `[0,100]`:
- **HR111/112 (DC)** `[0,50]` → `[0,100]` — fully evidenced by the field test.
- **HR313/314 (AC)** `[1,100]` → `[0,100]` — the GE app exposes 0–100; the v4.1.6 register doc lists 1–100 but it's hybrid-scoped (HR313/314 are AC-coupled registers absent on hybrids), so its bound isn't authoritative for the AC pair. The **0-disables semantics are not yet field-confirmed** — documented as such, pending a write-0 test on AC hardware.

A docstring records the provenance so the bound isn't "corrected" back to 50.

## Still to characterise (not blocking)

[...]

PR #306, opened 2026-06-25
fix(commands): restore the 1-floor on AC battery-limit setters (HR313/314)

dewet22:
Field feedback closing the AC arm of #302.

2.5.8 (#301) widened the AC setters `set_battery_charge_limit_ac` / `set_battery_discharge_limit_ac` (HR313/314) to `[0,100]`, dropping the `1`-floor on a "0 disables AC charge/discharge" assumption that wasn't field-confirmed (Codex flagged exactly this at the time).

**An AC-coupled tester has now confirmed** (via givenergy-hass): writing **0** to HR313/314 **ERRORs** on the hardware — `WriteHoldingRegisterResponse(ERROR 313 -> 0)` then a write timeout. Writing **1** works (drives the battery to near-zero). So `0` is invalid on the AC pair.

## Change

- Restore `1 <= val <= 100` on the two **AC** setters → a `0` now raises a clean `ValueError` instead of dispatching a doomed hardware write.
- The **DC** pair (`set_battery_charge_limit` / `set_battery_discharge_limit`, HR111/112) is **unchanged** at `[0,100]` — `0` is legitimately accepted there (separately confirmed on a DC inverter). The asymmetry is now documented in the docstrings.

## Tests

AC tests revert to `0` raising `[1-100]`; DC tests unchanged. Full suite + tox (py311–py314) green. Register-safety reviewed (strictly safer — a bound tightening that turns a known-erroring write into a clean client-side ValueError).

Closes the AC arm of #302 (the DC clamp was already field-confirmed there).

## Summary by CodeRabbit

* **Bug Fixes**
  * Tightened validation for AC battery charge and discharge limit settings so only values from 1–100% are accepted.
  * Updated error messages to clearly reflect the supported range and avoid invalid hardware writes.

* **Documentation**
  * Corrected the command reference for AC charge/discharge limits to show the 1–100% range and revised hardware behaviour notes.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -266,7 +266,7 @@ An unmarked row means the method is available on both inverter types. On `ThreeP
 | `disable_charge_target()` | Remove SOC limit, target 100% | |
 | `set_enable_charge(enabled)` | Enable or disable charging | |
 | `set_battery_charge_limit(val)` | Charge power limit (0–100%; firmware expected to clamp per-model) | |
-| `set_battery_charge_limit_ac(val)` | AC charge power limit (0–100%) | ⛔ commands-only |
+| `set_battery_charge_limit_ac(val)` | AC charge power limit (1–100%; 0 errors on hardware) | ⛔ commands-only |
 | `set_shallow_charge(val)` | Set shallow charge threshold (deprecated — use `set_battery_soc_reserve`) | ⛔ commands-only |
 
 ### Discharging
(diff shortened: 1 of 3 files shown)


dewet22-codex:

dewet22-codex (approved):
I’ve reviewed the current head and have no blocking concerns. The 1–100 AC bounds now match the field-confirmed hardware behaviour, reject the known-doomed zero write client-side, and leave the separately confirmed DC 0–100 contract untouched. Documentation and boundary tests agree, and the downstream hass controls have already restored their AC minima to 1. The full suite passes (1,161 tests), lint is clean, and CI is green.

PR #308, opened 2026-06-25
fix(plant): gate the temp-zero corruption cohort on the stale-bypass + short-read paths (#294)

### LINKED ISSUE #294: Splice guard: temp-zero cohort can still poison the baseline via stale-bypass / short-read paths (#289 follow-up)
Follow-up to #289. Seen live on a maintainer Gen1 (running `main`, post-#289): the splice guard rejecting a battery bank with the cache **stuck at temp-zero**, refusing the real temps —

```
Rejected battery bank for device 0x32 — sub-bus splice (>=2 physics-impossible deltas); keeping last-good.
Trips: IR(76) cell_temp_deci 0->303, IR(77) 0->289, IR(78) 0->295, IR(79) 0->283, IR(81) mosfet_temp_deci 0->304.
```

The baseline (`prev`) has cell temps at 0 and is rejecting the recovery to ~30 °C — the unrecoverable physics-poison #289 was meant to prevent (the #286 heal only recovers scalar-immutable IR97/98, not physics).

## The gap

#289's `is_corruption_cohort` gate is applied in exactly one place — the cold-start corroboration path (`_confirm_cold_start_baseline`, `plant.py:1026`). It does **not** guard the other two paths that commit a bank without a physics comparison:

1. **Stale-bypass** (`_splice_guard`, `plant.py:913`): after a gap > `STALE_BYPASS_SECONDS`, the incoming bank is adopted unconditionally (`return True`) — a temp-zero cohort arriving post-gap poisons the baseline.
2. **Short reads** (`plant.py:871`): `register_count < 60` returns True and commits directly, skipping the guard entirely — a partial/spliced read that zeroes temps seeds the cache.

So `is_corruption_cohort` closes the cold-start hole but not these two, and a temp-zero baseline can still form and then reject every real-temp frame.

## What's not yet pinned

Which path produced the field observation isn't confirmed — the cold-start path (with the gate) should have refused a temp-zero baseline, so it came in via stale-bypass or a short read. Nailing it needs the frame sequence: a `capture_frames` dump over a window when it recurs, replayed through the guard with debug logging (`tests/debug/replay.py`).

## Proposed direction

- Apply `is_corruption_cohort` to the stale-bypass adoption too (refuse-and-hold, consistent with cold-start). Care needed re the #256 stale-bypass intent (it deliberately re-adopts post-outage to avoid pinning to a stale baseline) — but refusing *the corruption signature specifically* and holding "unknown" until a healthy frame is the right trade, same as cold-start.
- Decide whether short-read commits should be physics-gated for battery temp registers, or left as-is (short reads are rare and the next full read re-asserts).

Refs: #289, #286 (heal only covers scalar-immutable), #256 (stale-bypass intent).

dewet22:
Closes #294.

#289 added `is_corruption_cohort` — refuse to *baseline* a frame with ≥2 cell-mass temps (IR76-79) at zero (the corpus's only corruption-cohort signature). Adopting it is unrecoverable: every later healthy frame then trips ≥2 physics vs the poisoned baseline and is hard-rejected forever (the #286 heal only recovers scalar-immutables, not physics).

But that gate was wired into only one of the three `_splice_guard` paths that adopt a bank **without** a physics comparison — cold-start. The **stale-bypass** (gap > `STALE_BYPASS_SECONDS`) and **short-read** (`register_count < 60`) paths still adopted a cohort blindly. Field-observed live on a Gen1 (#294): a temp-zero baseline rejecting the real ~30 °C temps.

## Change (`_splice_guard` only)

- **Short-read**: build the incoming frame before the early return, then hold instead of commit if it's the cohort. A lone temp-zero (`<2`) still commits.
- **Stale-bypass**: hold instead of adopt if the post-gap frame is the cohort. **Key subtlety:** a refused cohort restores the observation clock to `prev_seen` so the bypass stays armed — otherwise the cohort would consume the bypass and a healthy recovery frame would be rejected against the stale baseline, re-pinning the exact failure the bypass exists to prevent.

Cold-start was already covered, so the cohort is now held on **every** adopt-without-comparison path.

## Safety

Strictly safer — it only refuses *more* corruption; `classify_transition` (the corpus 1086/1/16 separation) is untouched, confirmed via `tests/debug/physics_delta_classify.py`. Read/commit guard, not a write path.

## Tests

- `test_splice_short_read_refuses_temp_zero_cohort` — ≥2-temp-zero short read held; lone temp-zero commits.
- `test_splice_stale_bypass_refuses_temp_zero_cohort` — post-gap cohort held, then a healthy frame recovers via the still-armed bypass.
- Existing stale-bypass / short-read / sustained-corruption tests unaffected. Full suite (1164) + tox (py311–py314) green.

## Follow-on (separate)

#299 — recovery for a sustained *legitimate* ≥2-physics step (smooth-streak plausibility heal-adopt). This PR's comprehensive cohort gate is its prerequisite.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of short battery data reads so invalid temperature patterns are now held back instead of being accepted.
  * Refined stale-data recovery so suspicious readings are no longer adopted after a long gap, helping preserve the last known good state.
  * Prevented recovery logic from being advanced by short reads, making the guard more reliable.

* **Tests**
  * Added regression coverage for short-read rejection and stale-baseline recovery behaviour.

### DIFF
--- a/givenergy_modbus/model/plant.py
+++ b/givenergy_modbus/model/plant.py
@@ -866,18 +866,13 @@ class Plant(GivEnergyBaseModel):
         a gap of more than ``STALE_BYPASS_SECONDS`` since the last *observed* full bank — a
         genuine polling outage, not a rejection streak — is too stale for per-poll thresholds
         to apply (legitimate multi-field drift over the gap would trip them) — all three fall
-        through as a no-op commit so the cache self-heals without manual recovery.
+        through as a no-op commit so the cache self-heals without manual recovery. The one
+        exception is the temp-zero corruption cohort, which is held (never adopted) on every
+        path, not just cold start (#294).
         """
-        if register_count < 60:
-            return True
         now_ts = now if now is not None else datetime.now(UTC)
         if now_ts.tzinfo is None:
             now_ts = now_ts.replace(tzinfo=UTC)
-        # Record this observation up front — a rejected bank is still an observation, so the gap
-        # computed below measures time since we last *saw* a full bank, not since we last accepted
-        # one. This is what separates a real outage from a sustained corruption run (see below).
-        prev_seen = self._splice_last_seen.get(device_address)
-        self._splice_last_seen[device_address] = now_ts
 
         cache = self.register_caches[device_address]
         prev = [0] * 60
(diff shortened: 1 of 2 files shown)


dewet22-codex:

dewet22-codex (approved):
I’ve reviewed the current head and have no blocking concerns. The short-read gate only counts temperatures present in the incoming response, and restoring the previous observation timestamp after a rejected post-gap cohort keeps the stale bypass available for the next healthy frame. The focused plant tests (132 passed), full suite (1,164 passed), Ruff, and CI are clean.

PR #310, opened 2026-06-25
feat(splice): opt-in heal-adopt for sustained legitimate >=2-physics step (#299)

### LINKED ISSUE #299: Splice guard: hard-reject has no recovery for a sustained legitimate cell-voltage step (heal-adopt / plausibility follow-up to #300)

dewet22:
## Problem

The `>=2-physics` hard-reject path in the splice guard has no recovery: any battery frame that trips two or more physics thresholds against the frozen baseline is permanently rejected. That is correct for corruption — but a *legitimate* charge event (LiFePO4 near-100%-SOC charge knee) can trip two thresholds in a single poll and then freeze telemetry indefinitely.

Field-observed on Gen1: an 18.5-min frozen battery state during a charge event. `cell_mV` stepped 154–198 mV/poll and `v_out` ~2263 mV/poll — both above the `>=2-physics` threshold, but plausible values evolving *smoothly* as the battery settled. #300 (2.5.7) widened the thresholds to clear that specific knee; the structural gap remained for future surges.

## Approach

Opt-in `splice_reject_heal_seconds` knob (default `None` = off, zero behaviour change). When enabled, a `>=2-physics, not scalar_immut` reject is recoverable if all three gates pass:

1. **Class restriction** — every trip vs the frozen baseline is a voltage/capacity class (`cell_mV`, `v_out_mV`, `v_cells_sum_mV`, `cap_centiAh`). Any temp/SOC/energy trip → not eligible → hard-reject as today.
2. **Absolute-range** — every tripping register's new value is within its absolute physical bound (raw units).
3. **Smooth multi-poll streak** — incoming frames are smooth continuations of *each other* (not the frozen baseline) for `SPLICE_REJECT_HEAL_POLLS=10` consecutive polls **and** `splice_reject_heal_seconds` elapsed.

On heal: return `True`, adopting the new values as the baseline; subsequent frames settle clean against it.

## Safety rationale

**Class restriction is the safety spine.** Across all 1,103 consecutive battery-bank transitions in the wire-capture corpus, every `>=2-physics` corruption event is either a *temp-zero* shape (cell-mass IR76-79, mosfet IR81, t_max/t_min IR103-104) or a *constant-register* flip — never a `cell_mV`/`v_out`/`v_cells_sum`/`cap_centiAh` surge. This is corpus-provable today.

The IR(103/104) `t_max`/`t_min` zeroing shape deserves explicit mention: it produces exactly two physics trips (`cell_temp_deci` class), no scalar-immutable, no serial change — the exact bucket this heal targets. `is_corruption_cohort` (which gates only IR76-79) misses it. The class restriction catches it by construction: `cell_temp_deci` is not in `RANGE_BY_CLASS`, so `heal_eligible()` returns `False`.

**Corpus safety proof** (`tests/model/test_splice_heal_corpus.py`): replays every consecutive battery-bank transition in all wire captures and asserts that *every* `>=2-physics` transition is `not heal_eligible`. This is the non-vacuous check — a plain "nothing heals" replay passes vacuously (the corpus has zero N-poll smooth-reject runs).

## Shipping posture

Default off (`splice_reject_heal_seconds=None`). The *positive* path's tuning rests on a smoothness assumption the corpus can't yet confirm: a real `>=18-min` charge-knee capture exceeding #300's thresholds (Gen1 and ideally a higher-power unit) is needed before flipping the default. Until then the knob exists, is safe to opt into, and the structural gap is documented in the issue.

## Changes

- **`battery_splice.py`**: `SPLICE_REJECT_HEAL_POLLS = 10`, `RANGE_BY_CLASS` absolute-bounds table, `heal_eligible(phys)` helper
- **`plant.py`**: `splice_reject_heal_seconds` field (default `None`), `_splice_reject_streak` PrivateAttr, hard-reject branch split (serial stays terminal; `>=2-physics` gated on knob + `heal_eligible`), `_handle_reject_streak` method, streak-pop discipline at all reset points
- **`client.py`**: opt-in `splice_reject_heal_seconds` ctor param, wired like `splice_heal_seconds`
- **Tests**: 8 plant tests, 3 `heal_eligible` unit tests, 1 client-wiring test, 1 corpus safety proof (new file)

Closes #299.

### DIFF
--- a/givenergy_modbus/model/plant.py
+++ b/givenergy_modbus/model/plant.py
@@ -1007,10 +1007,10 @@ class Plant(GivEnergyBaseModel):
         # corruption shape the corpus contains), hand off to the smooth-streak heal: hold now, adopt
         # only after a sustained smooth in-range run. Otherwise terminal reject, as before.
         if len(phys) >= 2 and not scalar_immut:
-            if self.splice_reject_heal_seconds is not None and heal_eligible(phys):
-                return self._handle_reject_streak(device_address, new, present, phys, now_ts)
             self._splice_escrow.pop(device_address, None)
             self._splice_immut_streak.pop(device_address, None)
+            if self.splice_reject_heal_seconds is not None and heal_eligible(phys):
+                return self._handle_reject_streak(device_address, new, present, phys, now_ts)
             self._splice_reject_streak.pop(device_address, None)
             self._bump(self.splice_reject_count, device_address)
             _logger.warning(
--- a/givenergy_modbus/model/plant.py
+++ b/givenergy_modbus/model/plant.py
@@ -13,9 +13,11 @@ from givenergy_modbus.model.battery_splice import (
     IMMUTABLE_SCALAR,
     IMMUTABLE_SERIAL,
     SCALAR_IMMUT_HEAL_POLLS,
+    SPLICE_REJECT_HEAL_POLLS,
     STALE_BYPASS_SECONDS,
     THRESHOLD_BY_CLASS,
     classify_transition,
+    heal_eligible,
     is_corruption_cohort,
 )
 from givenergy_modbus.model.devices import DeviceType, PlantDevice
@@ -554,6 +556,17 @@ class Plant(GivEnergyBaseModel):
     # Client(splice_heal_seconds=…). Runtime config; excluded from model_dump.
     splice_heal_seconds: float = Field(default=900.0, exclude=True)
 
+    # How long (seconds) a sustained *legitimate* >=2-physics battery step must persist, evolving
+    # smoothly and in-range, before the terminal hard-reject heals to it (#299). ``None`` (default)
+    # DISABLES the heal — the >=2-physics path stays terminal, zero behaviour change. A float opts
+    # in (recommended 300, = STALE_BYPASS_SECONDS; never set below it). Separate from
+    # splice_heal_seconds: that defends scalar poison which persists indefinitely (long window),
+    # this a transient charge-knee surge (short window). The positive (heal-fires) path can't be
+    # validated against the existing corpus, so it ships off by default until a real near-full-SOC
+    # knee capture confirms the smoothness assumption. Consumer-tunable via
+    # Client(splice_reject_heal_seconds=…). Runtime config; excluded from model_dump.
+    splice_reject_heal_seconds: float | None = Field(default=None, exclude=True)
+
     # Content-staleness tracker — the duration substrate for frozen-BMS-cache detection (#91).
     # Keyed identically to register_block_updated_at; each value is (content_hash, unchanged_since)
     # where unchanged_since is the ingestion time of the FIRST commit in the current
@@ -601,6 +614,17 @@ class Plant(GivEnergyBaseModel):
     _splice_immut_streak: dict[int, tuple[tuple[tuple[int, int], ...], datetime, int]] = PrivateAttr(
         default_factory=dict
     )
+
+    # Sustained-step heal streak (#299): per battery device address, (prev_incoming_frame,
+    # started_at, consecutive_count). When a >=2-physics reject is heal-eligible (voltage/cap-class
+    # trips, in absolute range — see battery_splice.heal_eligible) and the heal is enabled, the bank
+    # is held but the incoming frames are tracked here. The streak advances only while each new frame
+    # is a SMOOTH continuation of the *previous incoming* frame (classify_transition between
+    # consecutive incomings is clean), so a legitimate charge-knee surge — which drifts smoothly —
+    # accumulates, while corruption (which reverts or jumps) restarts the count. After
+    # SPLICE_REJECT_HEAL_POLLS smooth polls AND splice_reject_heal_seconds, the latest frame is
+    # adopted as the new baseline. Ephemeral; not serialised.
+    _splice_reject_streak: dict[int, tuple[list[int], datetime, int]] = PrivateAttr(default_factory=dict)
 
     # Direct-inverter register caches injected by add_direct_source() for multi-Client
     # reconciliation (#106 Phase 3). Stored separately from register_caches to avoid the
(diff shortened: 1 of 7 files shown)


dewet22-codex:

givenergy_modbus/model/plant.py
```diff
@@ -961,30 +986,54 @@ def _splice_guard(
...
 
+        # >=2 physics deltas with no recoverable scalar immutable: corruption (the temp-zero cohort,
+        # incl. the IR103/104 t_max/t_min pair) OR — rarely — a legitimate sustained step (the
+        # near-full-SOC charge knee, #299). When the heal is enabled (opt-in) AND the bank is
+        # heal-eligible (every trip a voltage/cap-class surge in absolute range, so it can't be any
+        # corruption shape the corpus contains), hand off to the smooth-streak heal: hold now, adopt
+        # only after a sustained smooth in-range run. Otherwise terminal reject, as before.
+        if len(phys) >= 2 and not scalar_immut:
+            if self.splice_reject_heal_seconds is not None and heal_eligible(phys):
+                return self._handle_reject_streak(device_address, new, present, phys, now_ts)
```
dewet22-codex: [Red / blocking] This new return path no longer clears the other guard state machines as the old >=2 hard-reject branch did. In particular, an existing `_splice_immut_streak` survives a physics-only heal-eligible poll, so the same scalar signature can resume afterwards and satisfy its “uninterrupted” count/time gates despite the interruption. `_splice_escrow` likewise survives, allowing a singleton held before the >=2 run to be confirmed afterwards. I think entering this mutually-exclusive sustained-step path needs to clear both states (and a regression test should cover at least the scalar streak interruption).

dewet22-codex (changes requested):
I found one blocking state-transition issue: the new recoverable >=2-physics path returns before clearing the singleton escrow and scalar-immutable streak that the former hard-reject path invalidated. That allows evidence to survive an interrupting, mutually-exclusive run and can shorten a later heal or confirmation. I’ve left the concrete concern inline. The eligibility restrictions, corpus proof, Ruff, focused tests (195 passed), and full suite (1,177 passed) are otherwise clean.

givenergy_modbus/model/plant.py
dewet22: Valid — both pops were missing from the heal-eligible path. Fixed by hoisting `_splice_escrow.pop` and `_splice_immut_streak.pop` to the top of the `>=2-physics` branch, before branching on `heal_eligible`, so they fire unconditionally. Regression test added: builds a scalar streak to count=9, sends a pure `>=2-physics` heal-eligible frame (IR98 forced to match the baseline so there is no co-occurring scalar_immut trip), then verifies 9 further drift polls don't heal — proving the interrupt reset the streak.

dewet22-codex:

dewet22-codex (approved):
I’ve re-reviewed the latest head and the blocking state-transition issue is fixed. The >=2-physics branch now clears both singleton escrow and scalar-immutable streak state before entering either outcome, and the regression test demonstrates that an interrupted scalar streak cannot resume its prior count. The focused suite (196 passed), full suite (1,178 passed), Ruff, and the exact CI merge commit under the Python 3.11 tox environment are clean; I’ve treated the reported Actions failure as transient.

PR #311, opened 2026-06-25
feat: add e_self_consumption_today/total computed fields

dewet22:
## Summary

- Adds `e_self_consumption_today` and `e_self_consumption_total` as `@computed_field` properties on `SinglePhaseInverter`, following the existing `e_consumption_today` pattern
- Definition (GivTCP parity): `max(0, PV generation − grid export)` — clamped at zero to handle battery-to-grid export where `grid_out` can temporarily exceed `pv_generation`
- Documented in each docstring that battery-to-grid export causes a slight under-count of true self-consumption; matches GivTCP's accepted approximation
- Both fields appear in `model_dump()` and return `None` if any input register is unavailable
- Placement on `SinglePhaseInverter` only — same as `e_consumption_today`; three-phase units are not verified to have all four input registers decoded

Closes #223

## Test plan

- [ ] New `test_e_self_consumption_computed_formula` — parametrized: normal positive case, clamp-to-zero (battery-to-grid), and exact-zero (export = PV)
- [ ] New `test_e_self_consumption_none_when_any_input_missing` — any absent input register → `None`
- [ ] Existing snapshot dicts (null inverter, Hybrid fixture, AC fixture, `test_from_actual`) updated with the new fields
- [ ] `uv run pytest --cov -q` → 1182 passed
- [ ] `uv run tox` → py311–py314 + format + lint + build all OK

## Summary by CodeRabbit

* **New Features**
  * Added new self-consumption energy metrics for single-phase inverters, showing today’s and total self-consumed energy.
  * Introduced an optional recovery mode for sustained battery telemetry spikes, helping prevent long-lived freezes in supported cases.

* **Bug Fixes**
  * Improved handling of unusual battery step changes so valid data can recover more gracefully while still rejecting unsafe values by default.
  * Added extra safeguards to avoid accepting physically impossible battery readings.

### DIFF
--- a/givenergy_modbus/model/inverter.py
+++ b/givenergy_modbus/model/inverter.py
@@ -949,6 +949,37 @@ class SinglePhaseInverter(  # type: ignore[valid-type,misc]
             return None
         return pv + grid_in - grid_out - ac_charge
 
+    @computed_field  # type: ignore[prop-decorator]
+    @property
+    def e_self_consumption_today(self) -> float | None:
+        """Self-consumption energy today (kWh): PV generation used on-site (GivTCP parity).
+
+        DERIVED: PV generation today − grid export today, clamped at ≥ 0.
+        Battery-to-grid export is counted in grid_out, so this slightly under-counts
+        true self-consumption when the battery discharges to grid — matching GivTCP's
+        accepted approximation. Returns None if any input is unavailable.
+        """
+        pv = self.e_pv_generation_today  # type: ignore[attr-defined]
+        grid_out = self.e_grid_out_day  # type: ignore[attr-defined]
+        if None in (pv, grid_out):
+            return None
+        return max(0.0, pv - grid_out)
+
+    @computed_field  # type: ignore[prop-decorator]
+    @property
+    def e_self_consumption_total(self) -> float | None:
+        """Lifetime self-consumption energy (kWh): PV generation used on-site (GivTCP parity).
+
+        DERIVED: PV generation total − grid export total, clamped at ≥ 0. Same
+        battery-to-grid under-count caveat as e_self_consumption_today.
+        Returns None if any input is unavailable.
+        """
+        pv = self.e_pv_generation_total  # type: ignore[attr-defined]
+        grid_out = self.e_grid_out_total  # type: ignore[attr-defined]
+        if None in (pv, grid_out):
+            return None
+        return max(0.0, pv - grid_out)
+
     # Plain @property (not @computed_field) so the deprecated alias doesn't
     # appear in model_dump() output. See #84 — renamed to work_time_total_hours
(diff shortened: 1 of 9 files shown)


dewet22-codex:

givenergy_modbus/model/inverter.py
```diff
@@ -949,6 +949,37 @@ def e_consumption_today(self) -> float | None:
...
+
+        DERIVED: PV generation total − grid export total, clamped at ≥ 0. Same
+        battery-to-grid under-count caveat as e_self_consumption_today.
+        Returns None if any input is unavailable.
+        """
+        pv = self.e_pv_generation_total  # type: ignore[attr-defined]
+        grid_out = self.e_grid_out_total  # type: ignore[attr-defined]
+        if None in (pv, grid_out):
+            return None
+        return max(0.0, pv - grid_out)
```
dewet22-codex, line 981: [Red / blocking] The downstream request in `givenergy-hass#223` intends to expose these as `state_class: total_increasing`, but this difference is not monotonic during battery-to-grid export. For example, PV/export counters moving from 10/1 kWh to 10/3 kWh makes this field fall from 9 to 7; Home Assistant treats that decrease as a meter reset and will add the post-drop value into long-term statistics, substantially over-counting self-consumption. The zero clamp only bounds the value, it does not preserve the cumulative invariant (the daily field has the same issue). I think the contract needs resolving before these fields back a total-increasing sensor: either the HA surface must explicitly use non-monotonic `total`, or the value needs a genuinely monotonic source/stateful derivation rather than the raw cumulative difference.

dewet22-codex (changes requested):
I found one blocking cross-repo contract issue: PV minus cumulative grid export can decrease during battery-to-grid export, while the linked Home Assistant request intends to publish these fields as total-increasing energy sensors. Home Assistant interprets such a decrease as a meter reset and would over-count long-term statistics. I’ve left the concrete sequence inline. The implementation otherwise matches the stated formula; the focused tests (213 passed), full suite (1,182 passed), and Ruff are clean.

dewet22:
Closing in favour of a clean branch without the already-merged #310 commits. Replacement PR incoming.

givenergy_modbus/model/inverter.py
dewet22: Good catch — the raw difference genuinely isn't monotonic during battery-to-grid export. Both docstrings in the replacement PR ([#312](https://github.com/dewet22/givenergy-modbus/pull/312)) now explicitly warn that the value is not monotonically increasing and that consumers using `state_class: TOTAL_INCREASING` must apply their own monotonic clamp (e.g. HA's `monotonic=True` sensor path) rather than consuming the raw value directly. The hass side has already committed to this approach.

PR #312, opened 2026-06-25
feat: add e_self_consumption_today/total computed fields

dewet22:
## Summary

- Adds `e_self_consumption_today` and `e_self_consumption_total` as `@computed_field` properties on `SinglePhaseInverter`, following the existing `e_consumption_today` pattern
- Definition (GivTCP parity): `max(0, PV generation − grid export)` — clamped at zero to handle battery-to-grid export where `grid_out` can temporarily exceed `pv_generation`
- Docstrings explicitly document both caveats: the battery-to-grid under-count approximation, and that the raw difference is NOT monotonically increasing. Consumers using `state_class: TOTAL_INCREASING` (e.g. HA energy dashboard) MUST apply a monotonic clamp (e.g. `monotonic=True` sensor path) rather than consuming the raw value directly
- Placement on `SinglePhaseInverter` only — same as `e_consumption_today`; three-phase units are not verified to have all four input registers decoded

Closes #223 (hass-side wiring tracked in givenergy-hass#223)

## Test plan

- [ ] New `test_e_self_consumption_computed_formula` — parametrized: normal positive case, clamp-to-zero (battery-to-grid), and exact-zero (export = PV)
- [ ] New `test_e_self_consumption_none_when_any_input_missing` — any absent input register → `None`
- [ ] Existing snapshot dicts (null inverter, Hybrid fixture, AC fixture, `test_from_actual`) updated with the new fields
- [ ] `uv run pytest tests/model/test_inverter.py` → 73 passed
- [ ] `uv run tox` → py311–py314 + format + lint + build all OK

### DIFF
--- a/givenergy_modbus/model/inverter.py
+++ b/givenergy_modbus/model/inverter.py
@@ -949,6 +949,51 @@ class SinglePhaseInverter(  # type: ignore[valid-type,misc]
             return None
         return pv + grid_in - grid_out - ac_charge
 
+    @computed_field  # type: ignore[prop-decorator]
+    @property
+    def e_self_consumption_today(self) -> float | None:
+        """Self-consumption energy today (kWh): PV generation used on-site (GivTCP parity).
+
+        DERIVED: PV generation today − grid export today, clamped at ≥ 0.
+
+        Two limitations to be aware of:
+        - Battery-to-grid export is counted in grid_out, so this slightly under-counts
+          true self-consumption when the battery discharges to grid. GivTCP accepts this
+          approximation; we match it here.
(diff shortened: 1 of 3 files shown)


dewet22-codex:

dewet22-codex (approved):
I’ve reviewed the cleanly scoped replacement and have no blocking concerns. The arithmetic matches the stated GivTCP parity formula, missing inputs remain explicit as None, and the non-monotonic battery-export behaviour is now part of the field contract rather than being presented as a safe total-increasing counter. The downstream hass integration has a separate monotonic wrapper available for the sensor surface. The focused model suite (213 passed), Ruff, and CI are clean.

Issue #293, opened 2026-06-23
AIO (Model.ALL_IN_ONE): e_pv_generation_* (IR44/45-46) mislabelled; e_discharge_year (IR29) uncharacterised


dewet22:
Surfaced from field testing on givenergy-hass#95 (tester AberDino — a Gen-1 Gateway + 2× All-in-One in parallel, battery-only / AC-coupled, no PV). Part of the #48 register audit; this is the All-in-One branch of the #174 re-identification.

## Confirmed: IR44 / IR45-46 are not PV generation on an AIO

`e_pv_generation_today` (IR44) and `e_pv_generation_total` (IR45/46) were re-identified as PV generation in #174 by sentinel cross-correlation — but that was on solar single-phase inverters. The AIO routes to the same shared `SinglePhaseInverterRegisterGetter` LUT, so it inherits those names, and on a unit with no PV they still populate.

Checked against the committed AIO capture fixture (`tests/fixtures/captures/aio_a/aio_arm612_5min.log`, DTC `8001`):

```
e_pv_generation_today  (IR44)    = 1.8 kWh
e_pv_generation_total  (IR45/46) = 7575.7 kWh
```

Non-zero (7.5 MWh lifetime) on a battery-only AIO with no solar input — so these cannot be PV generation. The name is wrong on `Model.ALL_IN_ONE`.

## Not yet confirmed: what they actually measure

AberDino cross-checked against GivTCP and the Gateway's "Parallel Total Charge Energy" and reads IR44/45-46 as AC charge/discharge throughput (charge tracking battery-charge with no conversion loss; discharge differing by losses). I can't confirm that specifically from the fixture, and there's a likely conflation to rule out:

- In the fixture, `e_pv_generation_today` (1.8) ≠ `e_battery_charge_today` (4.8).
- The register that *does* read identical to battery-charge is `e_ac_charge_today` (IR35 = 4.8) — the already-correctly-named AC-charge register, whose no-conversion-loss quirk #174 documented.

So the "identical to battery-charge" behaviour belongs to IR35, not IR44, in the data I have. IR44/45-46 are not-PV, but their precise identity (AC throughput? discharge? something else) is open.

## IR29 `e_discharge_year`

[...]

PR #313, opened 2026-06-26
feat: add e_pv_direct_today computed field

dewet22:
## Summary

Adds `e_pv_direct_today` — PV energy used **directly by load** today (kWh): solar that bypassed both the battery and the grid. The #223 "bonus" that was deferred when `e_self_consumption_*` landed (#312).

```
e_pv_direct_today = max(0, pv_generation − grid_export − battery_charge + ac_charge)
```

### Why the `+ ac_charge` term

`e_battery_charge` is a single counter that lumps PV-sourced and grid-sourced (AC) charging together. The PV-direct figure must only subtract the *PV* that went to battery, so the AC-charge portion is added back:

```
pv_to_load = pv_gen − export − (pv_to_batt + ac_charge) + ac_charge
           = pv_gen − export − battery_charge + ac_charge
```

Worked example — Octopus Go user, 4 kWh charged from grid overnight, daytime PV 10 kWh (6→load, 3→batt, 1→export): naive `10−1−7 = 2` (under by the 4 kWh AC charge); corrected `10−1−7+4 = 6` ✓. It's a **no-op for pure-solar systems** (`ac_charge = 0`).

### Scope decisions

- **DC-coupled only.** On AC-coupled and All-in-One units the PV-generation registers (IR44/45-46) are mislabelled — they read non-zero on battery-only hardware (#293) — so the field returns `None` there via an explicit `is_ac_coupled`/AIO gate (a plain None-guard wouldn't catch AIO, which *does* resolve `e_battery_charge_today`).
- **Effectively HYBRID_GEN1 today.** `e_battery_charge_today` only routes on GEN1 among DC models (`_BATTERY_ENERGY_SOURCE`), so other DC models return `None` until that map widens (#184). Tested on maintainer's GEN1 hardware shape.
- **Daily only.** The lifetime equivalent needs `e_ac_charge_total`, which single-phase firmware doesn't expose (three-phase only, IR1378/9), and `e_battery_charge_total` only resolves on GEN1.
- **Non-monotonic** intraday — documented; consumers using `TOTAL_INCREASING` must apply their own monotonic clamp (same contract as `e_self_consumption_*`).
- Result rounded to the inputs' native 0.1 kWh resolution (a four-term deci-scaled sum accumulates float noise).

Part of #223 (hass-side wiring tracked in givenergy-hass#223).

## Test plan

- [ ] `test_e_pv_direct_today_formula` — parametrized: AC-charge correction, pure-solar no-op, export-clamp, all-to-load
- [ ] `test_e_pv_direct_today_none_on_ineligible_models` — AC-coupled + AIO gated to None despite all inputs present
- [ ] `test_e_pv_direct_today_none_when_battery_charge_unavailable` — GEN3 (no routing) and missing ac_charge → None
- [ ] Existing `model_dump` snapshots updated (null inverter, two inverter fixtures, plant fixture)
- [ ] `uv run tox` → py311–py314 + format + lint + build all OK (1188 passed)

### DIFF
--- a/givenergy_modbus/model/inverter.py
+++ b/givenergy_modbus/model/inverter.py
@@ -999,11 +999,20 @@ class SinglePhaseInverter(  # type: ignore[valid-type,misc]
     def e_pv_direct_today(self) -> float | None:
         """PV energy used directly by load today (kWh): solar bypassing battery and grid.
 
-        DERIVED: pv_generation − grid_export − battery_charge + ac_charge, clamped ≥ 0.
-        The + ac_charge term nets out grid-sourced (AC) battery charging — e_battery_charge
-        lumps PV-sourced and AC-sourced charge into one counter, so without it the figure
-        under-counts by the day's AC-charge energy on time-of-use tariffs (Octopus Go /
-        Predbat). For pure-solar systems ac_charge is 0 and the term is a no-op.
+        DERIVED: (pv_generation − grid_export) − pv_to_battery, clamped to [0, pv − grid_export],
+        where pv_to_battery = max(0, battery_charge − ac_charge).
+
+        e_battery_charge lumps PV-sourced and AC-sourced charge into one counter, so the
+        PV portion that went to the battery is battery_charge − ac_charge — subtracting it
+        leaves the PV that reached load directly. Without the ac_charge term the figure
+        would under-count by the day's AC-charge energy on time-of-use tariffs (Octopus Go
+        / Predbat); for pure-solar systems ac_charge is 0 and it's a no-op.
+
+        Both bounds matter. The lower clamp (≥ 0) handles export exceeding PV. The upper
+        clamp (≤ pv − grid_export, via flooring pv_to_battery at 0) handles AC→DC
+        conversion loss / counter skew making ac_charge momentarily exceed battery_charge:
+        without it, direct PV could exceed total on-site PV self-consumption, which is
+        nonsensical (a subset can't beat its superset) — Codex review on #313.
 
         Restricted to DC-coupled solar hybrids: on AC-coupled and All-in-One units the
         PV-generation registers (IR44/45-46) are mislabelled — they read non-zero on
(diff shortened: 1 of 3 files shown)


dewet22-codex:

givenergy_modbus/model/inverter.py
```diff
@@ -994,6 +994,45 @@ def e_self_consumption_total(self) -> float | None:
...
+        pv = self.e_pv_generation_today  # type: ignore[attr-defined]
+        grid_out = self.e_grid_out_day  # type: ignore[attr-defined]
+        battery_charge = self.e_battery_charge_today
+        ac_charge = self.e_ac_charge_today  # type: ignore[attr-defined]
+        if None in (pv, grid_out, battery_charge, ac_charge):
+            return None
+        # Round to the inputs' native 0.1 kWh resolution: a four-term sum of deci-scaled
+        # floats accumulates representation noise (e.g. 4.000000000000001) that 1dp removes
+        # losslessly, since the true result is always a multiple of 0.1.
+        return round(max(0.0, pv - grid_out - battery_charge + ac_charge), 1)
```
dewet22-codex: [R] I think this needs an upper clamp as well as the zero clamp. `e_pv_direct_today` is meant to be the part of PV self-consumption that bypassed both battery and grid, so it cannot be greater than `pv - grid_out`. The current formula can exceed that whenever `ac_charge > battery_charge` because the `+ ac_charge` correction overcompensates; the updated snapshots already show that shape (`e_self_consumption_today` is 8.1 while `e_pv_direct_today` becomes 8.4). I would clamp the result to `max(0, pv - grid_out)` after applying the AC-charge correction, so conversion/counter skew can turn “PV to battery” into zero but not into a negative quantity that inflates direct PV beyond total PV used on-site.

dewet22-codex (changes requested):
I left one blocking comment on the new PV-direct calculation. The topology gates, None propagation, and focused tests otherwise look sensible to me, but the current formula can make direct PV exceed total PV used on-site when the AC-charge correction is larger than the battery-charge counter. I think the derived field needs that upper bound before it is safe to expose.

givenergy_modbus/model/inverter.py
dewet22: Good catch — and it'\''s a real physical effect, not just counter noise: AC→DC conversion loss means the AC-charge counter (grid energy in) can legitimately exceed battery_charge (energy stored), making battery_charge − ac_charge negative. Fixed in f2555e2 by flooring pv_to_battery at 0, i.e. result = max(0, (pv − grid_out) − max(0, battery_charge − ac_charge)) — mathematically identical to the upper clamp you suggested, but framed as "PV-to-battery can'\''t be negative". Both inverter snapshots now collapse to e_self_consumption_today (all charge was AC, so no PV reached the battery), and I added an overcompensation regression case plus a direct ≤ self-consumption invariant.

dewet22-codex:
I’m starting a re-review pass on this now.

dewet22-codex (approved):
I’ve re-reviewed the latest head and the fix addresses my previous concern. The PV-direct calculation now floors PV-to-battery at zero before subtracting it, so the value no longer exceeds total on-site PV self-consumption in the overcompensation case. The added regression/invariant coverage is in the right place, focused local model tests passed, and CI is green.

PR #327, opened 2026-06-28
feat: rename HR63-83 grid-protection fields (_1/_2/_3 → _trip/_reconnect/_grid) + installer-tier setters

dewet22:
## Summary

The GivEnergy installer app confirms the three-level band structure for single-phase grid protection (HR63-82):
- **trip** (HR63-70) — faster/tighter disconnect thresholds
- **reconnect** (HR71-78) — outer band before reconnecting to grid
- **grid** (HR79-82) — steady-state grid band

The library's opaque `_1`/`_2`/`_3` suffixes (and the parallel three-phase `v_grid_*_1/_2/_3` at HR1018-1041) are renamed to `_trip`/`_reconnect`/`_grid`. Old names remain as deprecation-warning `@property` aliases for one release — the same mechanism used for `enable_standard_self_consumption_logic`, `battery_power_cutoff`, and others.

Twenty-one bounds-validating setter helpers are added for the single-phase registers (`set_v_ac_low_limit_trip`, …, `set_v_ac_10min_protect`), each gated at `installer=True` and `confirm=True` with docstrings citing G98/G99/G100 compliance responsibility.

## Changes

- **`inverter.py`**: 20 field renames in `HOLDING_REGISTERS`; 20 deprecated `@property` aliases added to `SinglePhaseInverter`
- **`inverter_threephase.py`**: 20 field renames in `THREE_PHASE_HOLDING_REGISTERS`; 20 deprecated `@property` aliases added to `ThreePhaseInverter`
- **`commands.py`**: 21 `RegisterMap` constants (HR63-83) + 21 grid-protection setter functions; `_as_int` annotation corrected from `int` to `int | float` (matching documented behaviour)
- **`writable-registers.md`**: updated field names, band terminology, and setter reference
- **Tests**: 16 + 20 parametrised alias tests (single + three-phase); 3 parametrised setter groups covering the confirm gate, bounds validation, installer flag, and register address

## Test plan

- [x] `uv run pytest --cov` — 1319 tests pass (57 new)
- [x] `uv run tox` — all py311–py314 + lint + build pass
- [x] Deprecated aliases warn exactly once with the new name in the message, and are absent from `model_dump()`
- [x] Setters: raise without `confirm=True`, raise on out-of-range, return a single installer-flagged `WriteHoldingRegisterRequest` at the correct register address

## Summary by CodeRabbit

* **New Features**
  * Added clearer AC grid protection settings with named trip, reconnect, and steady-state bands.
  * Added new installer write helpers for setting voltage, frequency, timing, and 10-minute protection thresholds.

* **Bug Fixes**
  * Improved handling of numeric inputs so values are validated more reliably and booleans are rejected.
  * Updated inverter field names in outputs for clearer, more consistent register labels.

* **Documentation**
  * Expanded reference docs to explain grid protection behaviour and the new setter requirements.


dewet22-codex:
I’ve started a review of this PR.

tests/client/test_client_write_safety.py
```diff
@@ -614,6 +614,102 @@ async def test_installer_command_rejects_installer_register_without_flag():
...
+    # confirm=False (default) raises
+    with pytest.raises(ValueError, match="confirm=True"):
+        fn(good_voltage)
+
+    # Mid-range value accepted, installer flag set, correct register
+    reqs = fn(good_voltage, confirm=True)
+    assert len(reqs) == 1
+    assert reqs[0].installer
+    assert reqs[0].register == hr
+    assert reqs[0].value == int(good_voltage * 10)
```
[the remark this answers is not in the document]
dewet22: Done — `assert reqs[0].encode()` added to all three parametrised test bodies in the follow-up commit. HR63–83 are already present in `INSTALLER_WRITE_REGISTERS` (write_registers.py).

tests/client/test_client_write_safety.py
```diff
@@ -614,6 +614,102 @@ async def test_installer_command_rejects_installer_register_without_flag():
...
+    fn = getattr(importlib.import_module("givenergy_modbus.client.commands"), fn_name)
+
+    with pytest.raises(ValueError, match="confirm=True"):
+        fn(good_freq)
+
+    reqs = fn(good_freq, confirm=True)
+    assert len(reqs) == 1
+    assert reqs[0].installer
+    assert reqs[0].register == hr
+    assert reqs[0].value == int(good_freq * 100)
```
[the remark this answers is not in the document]
dewet22: Done — `assert reqs[0].encode()` added to all three parametrised test bodies in the follow-up commit. HR63–83 are already present in `INSTALLER_WRITE_REGISTERS` (write_registers.py).

tests/client/test_client_write_safety.py
```diff
@@ -614,6 +614,102 @@ async def test_installer_command_rejects_installer_register_without_flag():
...
+    fn = getattr(importlib.import_module("givenergy_modbus.client.commands"), fn_name)
+
+    with pytest.raises(ValueError, match="confirm=True"):
+        fn(1.0)
+
+    reqs = fn(2.5, confirm=True)
+    assert len(reqs) == 1
+    assert reqs[0].installer
+    assert reqs[0].register == hr
+    assert reqs[0].value == 250  # 2.5s × 100
```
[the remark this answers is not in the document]
dewet22: Done — `assert reqs[0].encode()` added to all three parametrised test bodies in the follow-up commit. HR63–83 are already present in `INSTALLER_WRITE_REGISTERS` (write_registers.py).

givenergy_modbus/client/commands.py
```diff
@@ -981,6 +1003,294 @@ def set_peak_shaving_enabled(enabled: bool) -> list[WriteHoldingRegisterRequest]
...
+
+
+def set_v_ac_low_limit_trip(voltage: float, *, confirm: bool = False) -> list[WriteHoldingRegisterRequest]:
+    """Set AC under-voltage trip threshold (HR63). Installer-tier. Range 0.0–500.0 V.
+
+    Part of the G98/G99/G100 trip band. Pass confirm=True after verifying grid-code compliance.
+    Value is written as deci (int(voltage × 10)).
+    """
+    _grid_confirm("set_v_ac_low_limit_trip", confirm)
+    val = _as_int(voltage * 10, "voltage × 10")
```
dewet22-codex: These setters multiply before using `_as_int`, which loses two of the protections the existing command validators rely on. `True` is accepted because `True * 10` becomes `10`, and valid centi-scaled values can be rejected due binary float artefacts (`set_t_ac_high_freq_trip(0.28, confirm=True)` raises on `28.000000000000004`; `0.29` raises on `28.999999999999996`). The time setters also only check `val < 0`, so `set_t_ac_low_voltage_trip(700.0, confirm=True)` returns a request with value `70000` that later fails in `encode()` instead of being rejected by the helper. I think these public write helpers need a scaled-value validator that checks the original argument type before scaling, handles decimal user input without float artefacts, and enforces the uint16 ceiling for centisecond time registers.

dewet22-codex (changes requested):
I found one blocking validation issue in the new grid-protection setters and left it inline. The read-side rename/alias work looks consistent, and the installer allowlist/encode coverage is in place, but the scaled setter validation currently accepts bools after multiplication, rejects ordinary centisecond values because of binary float artefacts, and lets over-uint16 time values escape until request encoding.

Checked locally: `uv run --group test pytest tests/client/test_client_write_safety.py tests/model/test_inverter.py tests/model/test_plant.py` (366 passed). I also reproduced the edge cases directly: `set_t_ac_high_freq_trip(0.28, confirm=True)` and `0.29` raise due float artefacts, `set_v_ac_low_limit_trip(True, confirm=True)` returns value `10`, and `set_t_ac_low_voltage_trip(700.0, confirm=True)` builds value `70000` before `encode()` rejects it.

dewet22-codex:
I’ve started a re-review of the latest changes.

givenergy_modbus/client/commands.py
dewet22: Addressed in the follow-up commit via `_as_scaled_int()`: type-checks before scaling (bools caught pre-multiplication), uses `round()` to absorb IEEE 754 residuals, and the 8 time setters now have an explicit `not (0 <= val <= 65535)` upper-bound check.

givenergy_modbus/client/commands.py
```diff
@@ -158,6 +180,19 @@ def _as_int(val: int, name: str) -> int:
...
+    """Type-check val before scaling; use round() to absorb IEEE 754 artefacts.
+
+    Unlike _as_int, checks for bool before multiplication so True/False are
+    always rejected (True * 10 == 10 slips past a post-scale bool check).
+    round() ensures decimal inputs like 0.28 s (27.999... × 100) are accepted
+    without raising on the fractional residual.
+    """
+    if isinstance(val, bool) or not isinstance(val, int | float):
+        raise ValueError(f"{name} must be a number, not {type(val).__name__}")
+    return round(val * scale)
```
dewet22-codex: This fixes the bool and IEEE-754 cases I called out, but `round()` now silently coerces values that are genuinely outside the register precision. For example, `set_t_ac_high_freq_trip(0.281, confirm=True)` writes `28` (0.28 s), `0.285` also writes `28` via Python’s tie-to-even rounding, and `set_f_ac_low_limit_trip(47.456, confirm=True)` writes `4746`. For these grid-protection setters I think it is important to keep the existing command-helper invariant of failing loud on caller values that cannot be represented exactly at the documented scale, while still tolerating binary float residue for values such as `0.28`. A Decimal/string-based scale check, or an integer-nearness check with a very small tolerance, would avoid both the original artefact failure and this silent rounding.

dewet22-codex (changes requested):
I re-reviewed the latest head (`ab8d0ed9f3082212a1158f5c731769e0333f01a0`). The original cases I raised are fixed: `0.28`/`0.29` now encode, bool inputs are rejected before scaling, and centisecond time setters enforce the uint16 ceiling.

I left one follow-up inline on `_as_scaled_int`: the new `round()` path also silently rounds genuinely over-precision inputs, so values such as `0.281`, `0.285`, or `47.456` write a nearby register value instead of failing loud. For grid-protection setters I still think that needs tightening before merge.

Checked locally: `uv run --group test pytest tests/client/test_client_write_safety.py tests/model/test_inverter.py tests/model/test_plant.py` (366 passed).

givenergy_modbus/client/commands.py
dewet22: Agreed — replaced bare `round()` with a nearness check (tolerance 1e-9): IEEE 754 residuals (0.28 × 100 = 27.999…, distance ≈ 3e-15) are accepted; genuine sub-precision values (0.281 × 100 = 28.099…, distance ≈ 0.1) raise `ValueError("cannot be represented exactly at scale ×100")`. Test added for both cases.

dewet22-codex:
I’ve started another re-review of the latest changes.

dewet22-codex (approved):
I’ve re-reviewed the latest head (`3569ddef16347343e4540d2f5f84b633a188e01a`) and the scaled setter validation now addresses my concerns. Representable decimal values such as `0.28`/`0.29` encode cleanly, bool inputs are rejected before scaling, centisecond time values enforce the uint16 ceiling, and genuinely over-precision inputs such as `0.281`, `0.285`, and `47.456` now fail loud instead of being silently rounded.

Checked locally: `uv run --group test pytest tests/client/test_client_write_safety.py tests/model/test_inverter.py tests/model/test_plant.py` (366 passed), plus direct reproduction of the edge cases above. No remaining blocking concerns from me.

PR #331, opened 2026-06-28
feat: decode HR84–156 protection/adjustment/grid-compliance zones + device-address map

dewet22:
## Summary

Cross-checked against the GivEnergy Installer app v1.154.3, this decodes the
three previously-skipped HR zones in `inverter.py` as raw `uint16` read-back
(same posture as the HR300–359 block: scales app-unconfirmed, so raw is
correct for baseline-read regardless of the true scale).

### Decoded zones

- **HR84–93 — ISO/GFCI/DCI protection** (`iso_protection_1/2`,
  `gfci_protection_value/time_1/2`, `dci_protection_value/time_1/2`). GE-named
  insulation, residual-current and DC-injection protection thresholds and trip
  times.

- **HR99–107 — string/grid voltage + power adjustments**
  (`string_1/2_voltage_adjustment`, `grid_r/s/t_voltage_adjustment`,
  `grid_power_adjustment`, `string_1/2_power_adjustment`). HR105
  `battery_voltage_adjust` already had a confirmed centi scale from prior work
  — that entry is unchanged.

- **HR129–156 — PF-curve / CEI 0-21 / LVFRT grid-compliance**
  (`power_factor_cmd_memory_state`, four PF curve points, `cei021_*`,
  `lvfrt_reactive_rate`, `lvfrt_low/high_fault_value/time_1–4`). PF values
  may use GE's EE encoding — noted in the section comment.

All three zones were already being polled (HR0–179 in `load_config`); the
registers were fetched but discarded. No poll-window changes needed.

### Deliberately not decoded

- **HR157–161** — absent from the installer map (no name) → left unmapped.
- **HR162 `RESET_USER_INFORMATION`** — momentary write-trigger, not telemetry
  → comment only.
- **HR115** — the installer names it `IS_LAN`, contradicting the old
  `island_check_continue` guess. Neither reading decoded until confirmed.
- **HR166 `enable_rtc`** — no installer entry and no RTC/Real-Time-Control enum
  in the extract; the rename was considered and found unconfirmed. Left as-is
  with a comment documenting the investigation.

### Device-address map

Added the authoritative device-address map (installer app `device_address_map`)
as a comment on `plant._getter_for_device_address`, covering both routed and
currently-unmodelled bands (BMS cluster 0x20+, HV BAMS 0x90+).

Reconciliation baseline regenerated: 390 → 394 matched.

## Test plan

- [ ] `uv run pytest` — 1325 passing
- [ ] `tox` clean across py311–py314 (+ lint, mypy, build)
- [ ] `mkdocs build --strict` clean

## Summary by CodeRabbit

* **New Features**
  * Expanded inverter register decoding to show more configuration and protection values, including isolation, GFCI/DCI, voltage/power adjustment, power-factor, and grid-compliance fields.

* **Documentation**
  * Updated reference notes to reflect the newly decoded register ranges and clarify which areas are still reported as raw values.

* **Tests**
  * Refreshed model expectations to cover the newly exposed inverter and plant fields.


dewet22-codex:

givenergy_modbus/model/inverter.py
```diff
@@ -536,22 +536,44 @@ class SinglePhaseInverterRegisterGetter(RegisterGetter):
...
         "enable_charge": Def(C.bool, None, HR(96)),
         "battery_low_voltage_protection_limit": Def(C.uint16, C.centi, HR(97)),
         "battery_high_voltage_protection_limit": Def(C.uint16, C.centi, HR(98)),
-        # skip voltage adjustment settings 99-104
+        # HR(99-107): string / grid voltage + power calibration adjustments. GE-named via
+        # the installer app; scales unconfirmed, so raw read-back. HR(105) keeps its
+        # established centi-scaled `battery_voltage_adjust` (the installer name confirms it).
+        "string_1_voltage_adjustment": Def(C.uint16, None, HR(99)),
+        "string_2_voltage_adjustment": Def(C.uint16, None, HR(100)),
+        "grid_r_voltage_adjustment": Def(C.uint16, None, HR(101)),
```
dewet22-codex: This maps HR101–104 as voltage/power adjustments, but the library already exposes those same holding-register addresses with different public meanings: HR101/102 are the installer grid-import limit helpers (`GRID_IMPORT_LIMIT` / `_ENABLED`), and HR104 is write-safe as `ENABLE_BATTERY_SELF_HEATING`. The consumer-app reconciliation still reports HR103/104 as LoRa/self-heating, and the docs below still keep the HR101/102 grid-import note. As it stands, consumers get two incompatible labels for one address, and read-after-write of the existing helpers would surface under unrelated fields. I think the conflicted HRs need to stay unmapped, or use deliberately neutral/caveated names, until the app-version/device-context split is resolved.

dewet22-codex (changes requested):
I’ve left one blocking inline concern around the HR101–104 read/write naming collision. The focused tests are green locally, but I think that address conflict needs resolving before these fields become public API.

givenergy_modbus/model/inverter.py
dewet22: Good catch — this turned out to be a genuine 1ph/3ph address reuse rather than a label clash. HR101-104 are one shared register block that the installer app names with the three-phase R/S/T-phase meaning (`GRID_R/S/T_VOLTAGE_ADJUSTMENT` + `GRID_POWER_ADJUSTMENT`), while single-phase firmware repurposes the same addresses — and there both the consumer app v4.0.7 and our own write surface agree on `grid_import_limit` / `_enabled` / `enable_lora` / `enable_battery_self_heating`.

So `SinglePhaseInverter` now carries the 1ph names (which also makes the read fields match `GRID_IMPORT_LIMIT` / `_ENABLED` on the write surface — read-after-write is consistent) and `ThreePhaseInverter` overrides the four addresses back to the R/S/T names, with a regression test for the split both ways. I also tried writing HR101/102 on a Gen1 AC hybrid to confirm the magnitudes: it rejected the FC06 writes with an error, so the 1ph names stay documentary (consumer-app + write-surface) rather than write-confirmed — noted in the field comment.

dewet22-codex:
I’ve started a re-review pass on this PR.

dewet22-codex (approved):
I’ve re-reviewed the latest head and have no blocking concerns remaining. The HR101–104 issue I raised is resolved by using the single-phase names where they align with the existing write surface, dropping those fields from the three-phase merge, and adding the three-phase R/S/T adjustment names back as explicit overrides.

Checked locally with:

`uv run --group test pytest tests/model/test_inverter.py tests/model/test_inverter_threephase.py tests/model/test_plant.py`

Issue #265, opened 2026-06-17
HV BMU per-module cell/temp decode unvalidated; module-0 stride overlaps BCU block (IR60-105)


dewet22:
## Summary

HV per-module (BMU) cell/temperature decode is unvalidated against real hardware and looks wrong. The first real three-phase HV capture (givenergy-hass#174, lamztib, GIV-3HY-11, 6 packs) shows the **BCU cluster level decodes correctly**, but **per-module BMU cell/temperature data decodes as garbage**, and the offset model has a structural overlap that likely explains it.

## What validated ✅ (BCU cluster level)

Decoding the capture's device 0x70 cache through `Bcu` (`model/hv_bcu.py`) gives self-consistent, sane values:

```
number_of_modules=6  battery_soc_max=71  battery_soc_min=68  battery_soh=95
battery_voltage=470.3 V  battery_current(IR76)=-0.1 A  cluster_cell_temperature(IR68)=72
```

`soc_max`/`soc_min` match the inverter's `battery_soc` (71%); voltage matches `v_battery_bms`. So `plant.hv_stacks` → `Bcu` is the working path for per-stack SOC and cluster temperature. (This is almost certainly what the hass integration already reads for per-pack SOC.)

## What's broken ⚠️ (BMU per-module)

`Bmu.from_register_cache(cache, bmu_index)` decodes module data at `base = 120 * bmu_index` (`model/hv_bcu.py`, `decode_cells_temps_serial`): cells IR(60+base), temps IR(90+base), serial IR(114+base). For **module 0** that's `base=0`, i.e. cells IR60–83 and temps IR90–113 — which **overlaps the BCU's own register block (IR60–105)** in the same device cache. So module 0 reads the BCU's cluster registers as if they were cell data.

Against the capture, `Bmu(cache, 0)` produces nonsense per-cell temps (381 °C, 522 °C, 1000 °C). The structural overlap is a strong candidate cause: module-0 cell data cannot share the IR60–105 space the BCU getter already claims.

## Why I'm not proposing a fix yet

I can't determine the *correct* per-module layout from this capture alone:
- It's the integration's **redacted** form (serials zeroed) and only 180 s / partial — device 0x90 and 0xa0 (likely further modules/stacks) decoded as empty/garbage too, so they may simply not have been populated in this window.
[...]

PR #333, opened 2026-06-28
docs: mark HV BMU per-cell decode wire-confirmed (#265)

dewet22:
Flips the "not yet wire-confirmed" caveat on the per-cell HV read recipe to confirmed.

A contributor with a 6-module HV stack (givenergy-hass#174) ran `FC04 / 0x50 / start 60 / count 60` and the same at `0x51`. Both decode cleanly through `BmuRegisterGetter`:

- 24 cells ~3.30 V (tight LiFePO4 balance)
- 24 temperatures ~36.5 °C
- distinct per-module serials at each address — proving each module is read at its own device address rather than a single echoed cache

That settles the design assumption behind the #265/#326 per-module rework against real hardware. Docs-only.

## Summary by CodeRabbit

* **Documentation**
  * Clarified the HV battery per-module addressing notes with updated verification details.
  * Added confirmed probe results for two device addresses, including observed cell voltages, temperatures, and serial differences.
  * Updated the wording to reflect that the decoded data is now wire-confirmed.

### DIFF
--- a/docs/reference/registers/installer-app-reference.md
+++ b/docs/reference/registers/installer-app-reference.md
@@ -59,8 +59,10 @@ sweeping the BCU band (`0x70+`) when the per-cell array lives one band lower on 
   `CELL_COUNT_PER_MODULE` / `TEMPERATURE_SENSOR_COUNT`.
 
 The library's old `Bmu(base=120·k)` decode was addressing the wrong *device* (a stride within
-the BCU cache), not the wrong offset. **Not yet wire-confirmed** — no capture has addressed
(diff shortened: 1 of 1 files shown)


dewet22-codex:

dewet22-codex (approved):
I’ve reviewed this and don’t have any blocking concerns. The reference wording matches the reported probe shape, and the docs-only diff is clean.

One non-blocking follow-up I noticed: a few internal comments still say the HV BMU separate-address probing is not yet wire-confirmed (hv_bcu.py, client.py, plant.py). I don’t think that needs to hold this PR, but I’d fold those caveat updates into whichever change next touches this area.

PR #335, opened 2026-06-29
fix: split single-phase fault_code into inverter_fault_code + inverter_warning_code (#330)

### LINKED ISSUE #330: Single-phase fault_code reads IR(39)+IR(40) as one uint32, but they are two separate words
`SinglePhaseInverterRegisterGetter` decodes `fault_code` as a uint32 spanning IR(39) and IR(40):

```python
"fault_code": Def(C.uint32, (C.hex, 8), IR(39), IR(40)),
```

The GivEnergy Installer app v1.154.3 reference names these as two distinct 16-bit registers:

- IR(39) `INVERTER_FAULT_CODE`
- IR(40) `INVERTER_WARNING_CODE`

So the current combined uint32 likely conflates a fault word and a warning word. Worth verifying against a real single-phase capture before changing anything — if confirmed, IR(39) and IR(40) should be split (fault vs warning), most likely via deprecated aliases to keep the existing `fault_code` accessor working.

Separately worth checking against the same capture: `inverter_fault_messages` currently decodes HR(223)/HR(224), whereas the installer reference points the inverter fault word at IR(39) — these may be two different sources for the same information, or one may be stale.

Surfaced while decoding the adjacent BMS/HV diagnostics in #328.

dewet22:
## Summary

Closes #330. `SinglePhaseInverterRegisterGetter` decoded IR(39)/IR(40) as a single `uint32`
`fault_code`. The GivEnergy installer app (v1.154.3) names them as two distinct 16-bit words:

- IR(39) `INVERTER_FAULT_CODE`
- IR(40) `INVERTER_WARNING_CODE`

so the combined value conflated a fault word and a warning word.

## Change

- Split into `inverter_fault_code` (IR39) and `inverter_warning_code` (IR40), each a raw 4-hex word.
- `fault_code` is kept as a **deprecated alias** on `SinglePhaseInverter` that warns and recombines
  the two words into the original 8-hex form (so existing consumers keep working for a release).

Left deliberately out of scope:
- **No message decode.** IR(39)'s bitfield enum isn't confirmed, so the new fields stay raw hex
  rather than risk a wrong decode.
- **HR(223)/HR(224) untouched.** `inverter_fault_messages` decodes HR(223)/HR(224); the installer
  reference points the inverter fault word at IR(39). Whether those are two sources for the same
  thing or one is stale is unresolved — noted in #330, not guessed at here.

## Tests

- New `test_fault_code_split_and_deprecated_alias`: the two words decode independently, appear in
  `model_dump()` (and `fault_code` does not), and the deprecated `fault_code` recombines to the old
  8-hex form with exactly one `DeprecationWarning` (None when either word is absent).
- Updated the four model-dump snapshots (three in `test_inverter.py`, one in `test_plant.py`).

Read-path only (IR reads); no register writes.

## Test plan
- [x] `uv run pytest` — 1329 passing

### DIFF
--- a/givenergy_modbus/model/inverter.py
+++ b/givenergy_modbus/model/inverter.py
@@ -905,7 +905,14 @@ class SinglePhaseInverterRegisterGetter(RegisterGetter):
         "e_battery_charge_today_alt1": Def(C.deci, None, IR(36)),
         "e_battery_discharge_today_alt1": Def(C.deci, None, IR(37)),
         "countdown": Def(C.uint16, None, IR(38)),
-        "fault_code": Def(C.uint32, (C.hex, 8), IR(39), IR(40)),
+        # IR(39)/IR(40) are two distinct 16-bit words per the GivEnergy installer app
+        # v1.154.3 (IR39 INVERTER_FAULT_CODE, IR40 INVERTER_WARNING_CODE), not one uint32.
+        # The old combined "fault_code" (uint32 over both) is kept as a deprecated alias on
+        # SinglePhaseInverter that recombines them. Left as raw hex words — IR(39)'s bitfield
+        # enum isn't confirmed, and the separate HR(223/224) inverter_fault_messages source
+        # is unreconciled (see #330), so no message decode here.
+        "inverter_fault_code": Def(C.uint16, (C.hex, 4), IR(39)),
+        "inverter_warning_code": Def(C.uint16, (C.hex, 4), IR(40)),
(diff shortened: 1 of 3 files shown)


dewet22-codex:

dewet22-codex (approved):
I’ve reviewed the latest head and do not have any blocking concerns. The IR39/IR40 split preserves the old `fault_code` read path as a deprecated property while keeping the dumped model on the corrected fields, and the tests cover the split, recombination, and `None` behaviour.

Checked locally with: uv run --group test pytest tests/model/test_inverter.py tests/model/test_plant.py

PR #336, opened 2026-06-29
fix: mock-server fails loud on a capture that parses to zero frames (#322)

### LINKED ISSUE #322: mock-server silently serves an empty plant when a capture yields zero frames
`MockPlant.from_capture` → `_iter_capture_frames` only parses lines shaped `<ts> rx <hex>`. If a capture matches nothing, it yields zero frames and `mock-server` still starts up — logging "N devices seeded" (a default empty device) and answering "absent" for every read, with no sign the capture didn't load.

Hit this with `tests/fixtures/captures/three_phase_hv_a/giv3hy11_hass174_180s.txt`, the one fixture in a `tx: <hex>` / `rx: <hex>` form (no leading timestamp) rather than the canonical `<ts> rx <hex>` the other seven use. Every line is skipped → 0 frames → empty mock, while it reports "1 device seeded."

**Proposed:** warn or error when a capture parses to zero frames / zero registers, so a format mismatch fails loudly. Separately, reconcile that fixture — teach the loader to tolerate the `tx:/rx:` form, or ship a normalized `<ts> rx` copy (the frames are valid; only the line prefix differs).

Found while using `mock-server` to drive the real GE app against synthetic HV state (#265).

dewet22:
## Summary

Closes #322. `MockPlant.from_capture` → `_iter_capture_frames` only parsed the canonical
`<ts> rx <hex>` line shape. The one HV fixture committed in the integration `rx: <hex>` / `tx: <hex>`
form (no leading timestamp, colon-suffixed direction) matched nothing → zero frames → an empty mock,
while `mock-server` happily started and logged "0 devices seeded" and answered "absent" for every
read. A format mismatch looked like a healthy-but-empty plant.

## Change

- **`_iter_capture_frames` tolerates the `rx:/tx: <hex>` form** (only `rx` kept), so that fixture
  now loads through the same path `mock-server` uses. The canonical `<ts> rx <hex>` shape is
  unchanged.
- **`plant_from_capture` raises on zero frames**, naming the path(s) and the accepted line shapes —
  so a format mismatch fails loudly rather than silently serving an empty plant. (Kept out of
  `_iter_capture_frames` itself, which legitimately returns `[]` for "no rx lines" — that contract
  is covered by existing unit tests.)
- **Consolidated `test_three_phase_from_capture.py`** onto the shared `plant_from_capture` loader,
  removing the local `rx:/tx:` workaround it had been carrying. Its existing value assertions
  (battery decode, HV BCU, grid block) now double as a regression test that the fixture loads
  end-to-end through the shared loader.

## Tests

- New: the `rx:/tx:` colon form parses (rx only); a capture that parses to zero frames raises
  `ValueError`.
- The three-phase decode tests pass unchanged through the shared loader — proving the loader change
  produces byte-identical frames.

Tooling only; no library runtime change.

## Test plan
- [x] `uv run pytest` — 1330 passing

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved capture-file parsing to handle more line formats, including `rx:`/`tx:` entries without timestamps and mixed direction tokens.
  * Ensured only received data is used when rebuilding frames, with entries processed in the right order.
  * Loading from an empty capture now fails clearly instead of creating an unusable empty plant, helping surface bad input sooner.

* **Tests**
  * Added coverage for colon-separated capture lines and empty-capture failure handling.

### DIFF
--- a/givenergy_modbus/testing/mock_plant.py
+++ b/givenergy_modbus/testing/mock_plant.py
@@ -69,19 +69,32 @@ _FALLBACK_SERIAL = "MK0000G000"
 def _iter_capture_frames(*paths: str | Path) -> list[bytes]:
     """Return complete rx frames from one or more capture ``.log`` files, ts-sorted.
 
-    Each line is ``<ts> rx <hex>``; the hex is a socket chunk, so frames may span lines.
-    We concatenate the rx stream in timestamp order and split on the frame marker using
-    the MBAP length field — the same framing the wire uses.
+    Two line shapes are accepted: the canonical ``<ts> rx <hex>`` and the integration
+    capture's ``rx: <hex>`` / ``tx: <hex>`` (no timestamp, colon-suffixed direction). Only
+    rx lines are kept; the hex is a socket chunk, so frames may span lines, so we concatenate
+    the rx stream in timestamp order and split on the frame marker using the MBAP length
+    field — the same framing the wire uses. Timestamp-less lines sort stably in file order.
     """
     entries: list[tuple[str, bytes]] = []
     for path in paths:
         for line in Path(path).read_text(encoding="utf-8").splitlines():
             parts = line.split()
-            if len(parts) >= 3 and parts[1] == "rx":
-                try:
-                    entries.append((parts[0], bytes.fromhex(parts[-1])))
-                except ValueError:
-                    continue
+            if len(parts) < 2:
+                continue
+            # Direction token is either parts[0] ("rx:"/"tx:", no timestamp) or parts[1]
+            # ("rx"/"tx", canonical with timestamp). Tolerate a trailing colon either way.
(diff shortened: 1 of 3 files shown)


dewet22-codex:

dewet22-codex (approved):
I’ve reviewed the latest head and do not have any blocking code concerns. The shared capture loader now handles both timestamped and `rx:` forms while still failing loudly for a capture that seeds no frames, and the three-phase fixture now exercises that same path. Codecov is still reporting a narrow patch-coverage miss in the parser tolerance branch; I did not treat that as blocking because the behaviour change itself is covered by the new colon-format and zero-frame tests.

Checked locally with: uv run --group test pytest tests/client/test_mock_plant_integration.py tests/model/test_three_phase_from_capture.py

Issue #332, opened 2026-06-28
Verify set_grid_import_limit_enabled (HR102) is valid on single-phase inverters


dewet22:
PR #331 named HR101/102 `grid_import_limit` / `grid_import_limit_enabled` on the read path, matching the consumer app v4.0.7 and the existing write-surface `RegisterMap.GRID_IMPORT_LIMIT` / `_ENABLED`.

While confirming the magnitudes, I tried writing HR101/102 on a Gen1 AC hybrid (single-phase). The inverter **rejected the FC06 writes with an error response**, repeatably:

```
2:6/WriteHoldingRegisterResponse(ERROR 102 -> 1/0x0001)
```

The library ships `set_grid_import_limit_enabled()` (HR102, installer-tier) as though it's writable. On this hardware it would error. Possible explanations:
- the feature is genuinely live only on certain models (three-phase has its own import limit at HR1130/1131);
- the write needs an installer-auth/precondition handshake the app performs and we don't;
- the register is read-only on single-phase.

Worth verifying before relying on the setter for single-phase. Captured here so it isn't lost — I haven't tested the three-phase or EMS write paths. Read-path naming (PR #331) is unaffected.

dewet22:
Closing with what cross-checking the GE apps could settle and what it couldn't.

Settled: the read-path naming (HR101 `grid_import_limit` / HR102 `grid_import_limit_enabled`, PR #331) is right for single-phase — the consumer app v4.0.7 and the library's own write surface agree, and HR101-104 read 0 on an unconfigured single-phase capture.

Not settled: whether the *setter* is live. On my Gen1 AC hybrid, FC06 writes to HR101/102 are rejected repeatably with an error response (`2:6/WriteHoldingRegisterResponse(ERROR 102 -> 1/0x0001)`). From one unit I can't tell whether that's model-specific (three-phase has its own import limit at HR1130/1131), needs an installer-auth handshake the app performs and we don't, or the register is simply read-only on single-phase — and I've no second single-phase model to compare against.

Practical upshot: `set_grid_import_limit_enabled()` fails loud (the inverter returns the error, nothing silent), and the caveat is now documented inline at the register definition. That's as far as I can take it without more hardware. If anyone hits this on a different model — or would prefer it gated behind a capability rather than failing at the wire — reopen and I'll pick it up.

PR #337, opened 2026-06-29
fix: HR1063 is an export-rate %, not a power — rename + correct bound + add setter (#263)

### LINKED ISSUE #263: Fix p_export_limit (HR1063) max bound and add missing setter for 3-phase export power rate
## Background

`p_export_limit` on the three-phase inverter model (`givenergy_modbus/model/inverter_threephase.py:268`) is defined as:

```python
"p_export_limit": Def(C.deci, None, HR(1063), max=6500),
```

The GE v4.1.6 register reference (the hybrid inverter spec — the closest doc we have) labels HR(1063) `BackflowPowerRateSet` with unit `0.1%` and range `0~1000`, i.e. 0–100% of rated power. Two issues:

1. **`max=6500` is wrong.** The doc's range tops out at 1000 (= 100.0% in 0.1% units). Our `max=6500` has no documented basis and would allow values well outside the register's valid range.

2. **No setter command.** The doc marks HR(1063) R/W, but `client/commands.py` has no `set_export_limit` or equivalent for the three-phase model — only the EMS model has `set_ems_export_power_limit` (HR(2071), in watts). If installers or hass want to write a 3-phase export power rate cap, there's no named command to do it.

## Caveats

- The v4.1.6 spec covers hybrid inverters, not three-phase. HR(1063) only appears in `_THREE_PHASE_LUT`, not the single-phase LUT. Whether the hybrid doc's unit/range characterisation holds for three-phase hardware hasn't been confirmed against a real capture — a `BackflowPowerRateSet` name in the hybrid doc being used on a three-phase model is circumstantial.
- A setter command should therefore be gated on confirmation of the unit/range from three-phase hardware before shipping.

## Proposed fix

1. Correct `max=6500` → `max=1000` to match the documented range.
2. Add a `set_3ph_export_power_rate(rate_pct_deci: int)` command (or similar name) writing HR(1063), bounded 0–1000, once the unit/range is verified on three-phase hardware.

dewet22:
## Summary

Closes #263. The three-phase `p_export_limit` (HR1063) was defined as `Def(C.deci, None, HR(1063),
max=6500)` — the same shape as the watt-valued `p_load_ac*` fields, i.e. modelled as a power. It is
actually the **export power rate cap as a percentage** of rated power:

- v4.1.6 doc: `BackflowPowerRateSet`, unit `0.1%`, range `0~1000` (R/W)
- installer app: `EXPORT_LIMIT_POWER_SET`, on the three-phase register map (real 3ph provenance)

## Changes

- **Rename** `p_export_limit` → `export_power_rate`, `Def(C.deci, None, HR(1063), min=0.0, max=100.0)`.
  Note `max=` bounds the **post-conversion** value: with `C.deci`, the old `max=6500` allowed a raw
  register value up to 65000 — far outside the documented `0~1000` range. The corrected bound caps at
  100.0%.
- **Hard-fail the old name.** `p_export_limit`'s contract (power in watts, `max=6500`) was always
  wrong, so it now raises `AttributeError` pointing at `export_power_rate` rather than silently
  returning a value that changed both scale and meaning. (This is deliberately stronger than the usual
  soft-deprecation alias — there's no correct old value to return.)
- **Add `set_export_power_rate(rate)`** — installer-tier write of HR1063 in native 0.1% units
  (`raw = round(rate * 10)`), bounded `[0, 100]%`. HR1063 added to `INSTALLER_WRITE_REGISTERS`
  (alongside its 1102/1103 export-limit siblings).
- Regenerate the app-reconciliation baseline (HR1063 now appears under `installer_write_not_in_app`,
  as expected — the consumer app doesn't expose this three-phase installer register).

## Caveat

HR1063 is **confirmed present** on three-phase hardware via the installer app, but I have no
three-phase unit to validate the *write* against — so the setter ships with that caveat in its
docstring and in #263, mirroring the HR101/102 write-rejection lesson (#332). The two-gate installer
safety model means it only fires via `Client.installer_command()`.

## Tests

- `export_power_rate` decodes raw 1000 → 100.0%; an over-range raw (>100% post-deci) is dropped by the
  bound; the field appears in `model_dump()` and `p_export_limit` does not; accessing `p_export_limit`
  raises `AttributeError`.
- `set_export_power_rate(50)` → `WriteHoldingRegisterRequest(1063, 500, installer=True)`; `33.3` → raw
  `333` (0.1% resolution); out-of-range raises; HR1063 ∈ `INSTALLER_WRITE_REGISTERS`.

## Test plan
- [x] `uv run pytest` — 1331 passing


dewet22-codex:

dewet22-codex (approved):
I’ve reviewed the latest head and do not have any blocking concerns. The HR1063 rename, 0-100 percent bounds, installer-tier setter, and PDU installer allowlist are aligned; I also checked that the normal one-shot write path stays separated from installer writes.

Checked locally with: uv run --group test pytest tests/client/test_client_write_safety.py tests/model/test_inverter_threephase.py

PR #338, opened 2026-06-29
fix: gate HR311/317 (export priority, EPS enable) on has_ac_config_block (#297)

### LINKED ISSUE #297: Fold HR311/317 (AC-config-block writes) into the has_ac_config_block client gate
Follow-up to #296. That PR introduced capability-based write gating: HR313/314 (`battery_*_limit_ac`) are admitted by `Client.one_shot_command` only when `capabilities.has_ac_config_block and not is_three_phase` (via `_AC_CONFIG_WRITE_SAFE_REGISTERS`), so a DC-coupled hybrid or an undetected client correctly rejects them.

**HR311 (`EXPORT_PRIORITY`) and HR317 (`ENABLE_EPS`) are the same kind of register** — both in the HR(300-359) AC-output config block (`has_ac_config_block`: "export priority, EPS enable, AC charge/discharge limits and pause mode … absent/times out on DC-coupled/hybrid models"). But they predate #296 and sit in the universal `_InverterCommands.WRITE_SAFE_REGISTERS`, so they're currently accepted for DC hybrids and undetected clients too — the imprecision #296 fixed for 313/314.

## Proposed

Fold 311/317 into `_AC_CONFIG_WRITE_SAFE_REGISTERS` and remove them from the universal set, so the whole HR(300-359) writable block is gated consistently on `has_ac_config_block`.

This is a **behaviour change for existing registers** (311/317 would start rejecting on DC hybrids / undetected / three-phase), so it wants its own reviewed PR rather than riding in #296 — and a check on whether any three-phase path legitimately writes 311/317 (EPS/export priority on AC_3PH) before excluding them there.

Refs: #296, #268 (capability gating), #162 (`has_ac_config_block`).

dewet22:
## Summary

Closes #297. HR311 (`EXPORT_PRIORITY`) and HR317 (`ENABLE_EPS`) are HR(300-359) AC-output
config-block registers — the same family as HR313/314, which #296 moved behind the
`has_ac_config_block` capability gate. 311/317 predated that gate and stayed in the universal
`_InverterCommands.WRITE_SAFE_REGISTERS`, so they were accepted for **DC-coupled hybrids, three-phase
units, and undetected clients** — exactly the imprecision #296 fixed for 313/314.

## Change

- Move `EXPORT_PRIORITY` (311) and `ENABLE_EPS` (317) from the universal
  `_InverterCommands.WRITE_SAFE_REGISTERS` into `_AC_CONFIG_WRITE_SAFE_REGISTERS`.
- `one_shot_command` (and the installer path) now admit them only when
  `capabilities.has_ac_config_block and not is_three_phase` — i.e. Model.AC / All-in-One — exactly
  like 313/314. The whole HR(300-359) writable block is now gated consistently.
- The PDU-level `WRITE_SAFE_REGISTERS` (Gate 2) keeps 311/317, mirroring how 313/314 are handled
  there; the model-awareness is the client-side gate (Gate 1), not the PDU.

## Behaviour change

DC-coupled hybrids, three-phase, and undetected clients now **reject** HR311/317 writes (previously
accepted). AC/AIO single-phase is unchanged. Three-phase EPS / export-priority, if it exists, lives
at three-phase addresses — not the single-phase HR311/317 — so nothing legitimate on 3ph regresses.

## Tests

Mirror the existing 313/314 gating tests for 311/317:
- 311/317 ∈ `_AC_CONFIG_WRITE_SAFE_REGISTERS`, ∉ universal and ∉ three-phase allowlists;
- AC / All-in-One accept the writes;
- HYBRID_GEN1, AC_3PH, HYBRID_3PH, and undetected all reject with `InvalidPduState`.

## Test plan
- [x] `uv run pytest` — 1335 passing

### DIFF
--- a/givenergy_modbus/client/commands.py
+++ b/givenergy_modbus/client/commands.py
@@ -1362,13 +1362,16 @@ def restore_factory_defaults(*, confirm: bool = False) -> list[WriteHoldingRegis
 
 # HR(300-359) AC-output config-block writes that are gated on capabilities.has_ac_config_block
 # rather than the model-class allowlist — one_shot_command unions these into the safe set only when
-# the detected model exposes the block (Model.AC / All-in-One), never for a DC-coupled hybrid or an
-# undetected client (#295/#296 review). Currently just the AC charge/discharge limits; HR311/317
-# (also HR300-359) predate the gate and stay in the universal set pending a separate cleanup.
+# the detected model exposes the block (Model.AC / All-in-One), never for a DC-coupled hybrid, a
+# three-phase unit, or an undetected client (#295/#296 review). The AC charge/discharge limits plus
+# export priority (HR311) and EPS enable (HR317) — the whole HR(300-359) writable block is now gated
+# here (#297); 311/317 were previously in the universal set, accepted on DC hybrids and undetected.
 _AC_CONFIG_WRITE_SAFE_REGISTERS: frozenset[int] = frozenset(
     {
+        RegisterMap.EXPORT_PRIORITY,  # 311
         RegisterMap.BATTERY_CHARGE_LIMIT_AC,  # 313
         RegisterMap.BATTERY_DISCHARGE_LIMIT_AC,  # 314
+        RegisterMap.ENABLE_EPS,  # 317
     }
 )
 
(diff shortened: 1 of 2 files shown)


dewet22-codex:

dewet22-codex (approved):
I’ve reviewed the latest head and do not have any blocking concerns. Moving HR311/HR317 behind the same `has_ac_config_block` gate as HR313/HR314 matches the capability model and closes the undetected/DC-hybrid/three-phase write paths covered by the new tests.

Checked locally with: uv run --group test pytest tests/client/test_client_write_safety.py

PR #339, opened 2026-06-29
fix: scale e_inverter_export_total as 0.1 kWh (#182)

### LINKED ISSUE #182: fix: inverter energy totals scaled 10× (HR4141/2, HR4107/8)
Surfaced by auditing the v4.1.6 hybrid MODBUS protocol doc (2024-10-30) against the current register map.

The doc gives `0.1 kWh` units for several 32-bit energy registers that the library currently returns as raw `uint32` (no scaling):

| Reg | Attr | Doc unit | Code converter | Effect |
|-----|------|----------|----------------|--------|
| HR4141/4142 | `e_inverter_export_total` | 0.1 kWh | `uint32` (raw) | reported 10× too high |
| HR4107/4108 | `pv_power_setting` | 0.1 kWh | `uint32` (raw) | 10× — but this is a writable *setting*, so confirm intent first |

The sibling energy totals already scale correctly via `deci` (IR6/7 `e_battery_throughput`, IR11/12 `e_pv_total`), so the inconsistency looks like a genuine miss rather than a deliberate raw passthrough.

`e_inverter_export_total` is the high-confidence one — it's a reported cumulative total and should match the 0.1 kWh scaling the doc applies to every other energy register. `pv_power_setting` is writable ("PV total set" in the doc) so I'd like to understand the write path before changing its scale.

Fix: add `C.deci` post-conversion (consistent with the other totals) and a regression test. I haven't confirmed against hardware yet — the scaling is inferred from the doc unit column, which has been wrong before (see the IR10 current-unit correction in the doc's own revision history).

dewet22:
## Summary

Closes #182. `e_inverter_export_total` (HR4141/4142) was decoded as a raw `uint32`, but the v4.1.6
doc gives it `0.1 kWh` units — like the sibling cumulative totals `e_pv_total` (IR11/12) and
`e_battery_throughput` (IR6/7), which already use `C.deci`. Raw decoding over-reported by 10×.

## Change

Add `C.deci`. That's the established convention for these 0.1 kWh cumulative totals, and the
adjacent `_alt` energy block already uses it.

## Scope notes

- **Not live today.** The register sits above the library's poll path (max base HR240), so this
  corrects any *direct* read rather than a value the library currently surfaces.
- **Doc-driven.** No wire capture confirms the magnitude (the register is never populated in any
  fixture), and it's absent from both reverse-engineered app inventories — only the v4.1.6 doc covers
  it. The fix rests on the doc unit + sibling-total consistency.
- **HR4107/4108 `pv_power_setting` deliberately left raw.** Same doc unit, but it's a writable seed
  of undocumented effect (not a reported total), with no setter — its scale stays unresolved pending
  intent rather than changed blind. (Per the issue's own caution.)

## Tests

New `test_e_inverter_export_total_deci_scale`: raw uint32 12345 → 1234.5 kWh. Existing `None`
snapshots are unaffected (`deci(None)` is `None`).

## Test plan
- [x] `uv run pytest` — 1342 passing

### DIFF
--- a/givenergy_modbus/model/inverter.py
+++ b/givenergy_modbus/model/inverter.py
@@ -841,7 +841,12 @@ class SinglePhaseInverterRegisterGetter(RegisterGetter):
         #
         # Holding Registers, block 4140-4199
         #
-        "e_inverter_export_total": Def(C.uint32, None, HR(4141), HR(4142)),
+        # 0.1 kWh per the v4.1.6 doc, matching the sibling cumulative totals (e_pv_total,
+        # e_battery_throughput) which already use C.deci; raw uint32 over-reported 10×. Not
+        # currently in the library poll path (max base HR240), so this corrects any direct read
+        # (#182). Doc-driven — no wire capture confirms the magnitude, as for the adjacent _alt
+        # energy block.
+        "e_inverter_export_total": Def(C.uint32, C.deci, HR(4141), HR(4142)),
         #
         # Input Registers, block 0-59
         #
--- a/tests/model/test_inverter.py
+++ b/tests/model/test_inverter.py
@@ -1549,6 +1549,16 @@ def test_fault_code_split_and_deprecated_alias():
         assert empty.fault_code is None
 
 
+def test_e_inverter_export_total_deci_scale():
+    """HR4141/4142 e_inverter_export_total is 0.1 kWh (C.deci), not raw uint32 (#182)."""
+    from givenergy_modbus.model.register import HR
+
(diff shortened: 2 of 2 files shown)


dewet22-codex:

dewet22-codex (approved):
I’ve reviewed the latest head and do not have any blocking concerns. The `e_inverter_export_total` change applies the documented 0.1 kWh scaling through the existing converter path, with a focused regression test covering the direct-read case.

Checked locally with: uv run --group test pytest tests/model/test_inverter.py

PR #340, opened 2026-06-29
feat: extend _DTC_RATED_POWER from the app v4.0.7 model-code table (#320)

### LINKED ISSUE #320: resolve_model: backstop the Gen1/2/3 firmware heuristic + complete the rated-power tables with the app 4.0.7 vendor model-code table
The v4.0.7 app embeds a 65-entry HR(0) model-code → model-name table, extracted this session and committed at `docs/reference/registers/app_4.0.7_inventory.json` (`model_codes`). It's the manufacturer's own authoritative code map, and it lines up with the weak spots in our current model resolution.

**Today** (`model/inverter.py`): `resolve_model()` slices the 2-char DTC prefix and looks it up in `_DTC_PREFIX_TO_MODEL` (6 entries), except the 0x20xx Hybrid family, which it disambiguates by *firmware century* (`_HYBRID_FW_CENTURY_TO_GEN`) rather than the code. Rated power comes from `_DTC_RATED_POWER` (~37 codes, commented "likely not exhaustive") and `_DTC_BATPOWER` (8 codes). There's no full-code → `Model` table.

**Where the vendor table helps:**
- The 0x20xx Gen1/2/3 split currently rests on a firmware heuristic; the table gives per-code identities to cross-check against.
- The "EA-prefix Gen2, first DTC digit unknown" guess (the comment near the `Model` enum) is exactly the kind of thing vendor data resolves.
- It can complete/verify `_DTC_RATED_POWER` — the table encodes per-code kW ratings (e.g. 0x8001 = All-in-One 6kW).

**Possible shape** (not committing to it): a full-code lookup consulted first in `resolve_model()`, falling through to the existing prefix/firmware logic for codes the table doesn't cover. This path is load-bearing — `detect()` is consumed across cli/hass/modbus and hardware-validated — so the bar is bit-identical behaviour on known codes, with the table only adding resolution where we currently guess. Characterisation tests first.

**Caveat:** the table's values are marketing names ("All in One 6KW"), not our `Model` enum members, so the name → `Model` mapping needs care, and the high-nibble → family pattern is inferred rather than stated. Worth validating a few codes against real units before trusting it wholesale. Relates to #48 (register audit) for the rated-power completion.

dewet22:
## Summary

Closes #320 (the rated-power half; see "Out of scope" below). The GivEnergy app v4.0.7 embeds an
authoritative HR(0) model-code → model-name table. All **36 codes that overlapped** the library's
existing `_DTC_RATED_POWER` agreed **exactly** (zero mismatches), so this adds the **27 new codes**
the library previously couldn't rate:

- ALPS-CUBE-HY5.0 (`2401-2406`), POL-HY-8.0-GL (`2501-2506`)
- GIV-3HY-20-HV (`4005/4006`), GIV-PM commercial (`4007/4101/4103`)
- AcCouple (`6001-6006`), AIO-HY 11kW (`8205`), Battery-Ready (`8301-8303`)

**Purely additive.** Codes that returned `None` now return the vendor watt rating; no known code's
rating changes. The two Gateway codes (`7001/7002`) carry no kW in the table and are left as-is
(`7001` keeps its existing `12000`).

## Out of scope (stays hardware-blocked)

`resolve_model()` is deliberately untouched. The table **lumps** the 0x20xx Gen1/2/3 family under one
code — which *confirms* the generation split lives in firmware, not HR(0) — offers no EA-prefix code,
and its marketing names (e.g. "GIV-PM-30/50") don't map cleanly to `Model` enum members. So it can't
safely improve model *identity*, only ratings. The Gen-split and EA-prefix gaps remain blocked on a
field report, exactly as before.

## Tests

Adds a characterisation test driving the real `inverter_max_power` property via HR(0) across the
**full 64-code table** — the pre-existing codes had **no test at all**, so this both pins them
(catching any future accidental edit) and asserts the 27 new codes return the vendor watts. Plus an
unknown-code (`0x7002`) → `None` case. Expected watts are written explicitly in the test, independent
of the source dict.

## Test plan
- [x] `uv run pytest` — 1406 passing

### DIFF
--- a/givenergy_modbus/model/inverter.py
+++ b/givenergy_modbus/model/inverter.py
@@ -96,7 +96,10 @@ _DTC_PREFIX_TO_MODEL: dict[str, Model] = {
 AC_COUPLED_MODELS: frozenset[Model] = frozenset({Model.AC, Model.AC_3PH})
 
 # Rated AC output power in watts, keyed by 4-char hex device type code.
-# Sourced from britkat1980/giv_tcp:dev3; likely not exhaustive for all variants.
+# Original entries from britkat1980/giv_tcp:dev3. Entries marked "(app v4.0.7)" were added from
+# the GivEnergy app's authoritative HR(0) model-code table (#320): all 36 codes that overlapped the
+# existing table agreed exactly, so the additions are purely new codes (the table only resolves codes
+# that previously returned None). The two Gateway codes (7001/7002) carry no kW rating in the table.
 _DTC_RATED_POWER: dict[str, int] = {
     "2001": 5000,
     "2002": 4600,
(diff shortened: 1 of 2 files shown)


dewet22-codex:

dewet22-codex (approved):
I’ve reviewed the latest head and do not have any blocking concerns. The rated-power table expansion is additive, the existing `resolve_model()` behaviour is left alone, and the new parametrised test drives the real `inverter_max_power` lookup across the full mapped table plus an unknown-code case.

Checked locally with: uv run --group test pytest tests/model/test_inverter.py

Issue #264, opened 2026-06-17
3-phase i_battery (IR1140) wrong scale: C.deci should be C.centi (10x too large)


dewet22:
## Summary

`i_battery` on the three-phase inverter model uses the wrong scale converter. It's defined with `C.deci` but field evidence shows it should be `C.centi` (matching the single-phase definition). Values currently decode 10× too large.

## Evidence

First real three-phase wire capture (givenergy-hass#174, lamztib, GIV-3HY-11 HV hybrid, 6 packs) decoded offline. Cross-checking the battery current against the V×I identity at the battery terminals:

| `v_battery_bms` | raw IR(1140) | `i_battery` as deci | `i_battery` as centi | `p_battery_charge` (deci, validated) |
|---|---|---|---|---|
| 470.3 V | −17 | −1.7 A → V·I = −799 W | −0.17 A → V·I = **−80 W** | **86.4 W** |
| 470.3 V | −16 | −1.6 A → −752 W | −0.16 A → **−75 W** | **76.1 W** |
| 470.3 V | −13 | −1.3 A → −611 W | −0.13 A → **−61 W** | **62.8 W** |
| 470.3 V | −5  | −0.5 A → −235 W | −0.05 A → **−24 W** | **24.7 W** |

With the **centi** scale, V·I tracks the independently-validated `p_battery_charge` (IR1138/1139, `C.deci`, watts) to within 1–8% across the capture window. The `deci` scale is off by 10×.

## Current definition

`givenergy_modbus/model/inverter_threephase.py:376`
```python
"i_battery": Def(C.int16, C.deci, IR(1140), min=-500.0, max=500.0),
```
The inline comment reads `# i_battery shadows single-phase IR(51); same converter, different scale` — but single-phase `i_battery` (`model/inverter.py:713`) is `C.centi`, and the field data agrees with `centi`, not the deliberate `deci` override.

## Proposed fix

Change the converter to `C.centi`:
```python
"i_battery": Def(C.int16, C.centi, IR(1140), min=-500.0, max=500.0),
```
Add a regression test asserting the decoded current against the captured raw value (e.g. raw −17 → −0.17 A), cross-referenced to this capture.

## Scope note

[...]

PR #341, opened 2026-06-29
fix: decode 3ph per-phase active power as signed watts; rename off p_load_ac*/p_out_ac* (#185)

dewet22:
## Summary

Partial fix for #185. The three-phase per-phase active-power registers IR(1083-1085) and
IR(1091-1093) were decoded as **unsigned `C.deci`** — reading 10× low and silently dropping negative
values (out-of-bounds → `None`). Both the v4.1.6 doc (`1W`) and the installer app (`signed`, no
`divide_by`) say they are **signed 16-bit watts**.

## Confirmed on real hardware

Decoding the lamztib GIV-3HY-11 capture, the three grid-meter phases read **-80 / -185 / +269 W**,
which sum to **+4 W** — exactly `p_meter_export` (4.0 W) on that (exporting) house. Under the old
`deci` they'd sum to 0.4 W. That per-phase ↔ aggregate identity is independent confirmation of the
signed-watt scale, beyond the doc/installer agreement. (The capture test now asserts this sum.)

## Change

- IR(1083-1085) `p_load_ac1/2/3` → **`p_meter_active_ac1/2/3`** — grid-meter active power per phase
  (doc `MeterActivePower_R/S/T`, installer `METER_ACTIVE_POWER_R/S/T`; the old "load" name was wrong).
- IR(1091-1093) `p_out_ac1/2/3` → **`p_inverter_active_ac1/2/3`** — inverter active power per phase
  (doc/installer `ActivePower_R/S/T`).
- `Def(C.deci, …, max=6500)` → `Def(C.int16)` — signed watts; the deci-era bound is dropped.
- The old names **hard-fail** (`AttributeError`) — their unsigned-deci, load-mislabelled contract was
  always wrong, so returning a value would be returning a wrong one. Same treatment as the
  `export_power_rate` rename in #263.
- The v4.1.6 audit inventory's cross-refs for these six registers are synced to the new names and the
  now-resolved 1:1 scale (a full regen needs the private doc source, which isn't in the repo).

## Still open in #185 (not this PR)

The holding-register protection-timing/slope scales (HR1034/1051/1052/1011/1067/…) remain blocked:
the installer's 3ph *holding*-register map carries names only, no scale, so it can't adjudicate those
— they still need three-phase hardware. (`i_battery` IR1140 is also left alone: its `centi` was
wire-validated in #264, and the installer's `deci` is treated as the lesser evidence.)

## Test plan
- [x] `uv run pytest` — 1342 passing (incl. the capture sum-identity assertion)

### DIFF
--- a/docs/reference/registers/v4.1.6_inventory.json
+++ b/docs/reference/registers/v4.1.6_inventory.json
@@ -11152,13 +11152,13 @@
         "doc_unit": "1W",
         "doc_scale": 1.0,
         "code_attrs": [
-          "inverter_3ph.p_load_ac1"
+          "inverter_3ph.p_meter_active_ac1"
         ],
         "code_convs": [
-          "deci"
+          "int16"
         ],
-        "code_scale": 0.1,
-        "factor_off": 0.1
+        "code_scale": 1.0,
+        "factor_off": 1.0
       },
       {
         "reg": "IR1084",
@@ -11166,13 +11166,13 @@
         "doc_unit": "1W",
         "doc_scale": 1.0,
         "code_attrs": [
-          "inverter_3ph.p_load_ac2"
+          "inverter_3ph.p_meter_active_ac2"
         ],
         "code_convs": [
-          "deci"
+          "int16"
         ],
-        "code_scale": 0.1,
-        "factor_off": 0.1
+        "code_scale": 1.0,
+        "factor_off": 1.0
       },
       {
         "reg": "IR1085",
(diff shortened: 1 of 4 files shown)


dewet22-codex:

dewet22-codex (approved):
I’ve reviewed the latest head and do not have any blocking concerns. The six per-phase active-power fields now use signed-watt decoding, the old mis-scaled names fail loudly rather than returning misleading values, and the capture test’s phase-sum assertion gives this a good real-wire backstop.

Checked locally with: uv run --group test pytest tests/model/test_inverter_threephase.py tests/model/test_three_phase_from_capture.py

Issue #268, opened 2026-06-17
Epic: unify detect()/refresh() under a strategise()/probe() loop — derive capabilities, drive cadence, reconstruct offline from cache state


dewet22:
## Motivation

`detect()` couples discovery *policy* (what to read next) with the wire *I/O* that reads it, and `PlantCapabilities` can only be built by a live client talking to hardware. Two consequences:

1. **No offline capabilities.** A `Plant` rebuilt from a `probe`/`export` dump (e.g. givenergy-cli's new offline `shell`, cli PR #70) carries `register_caches` but `capabilities=None`, so the typed views (`.inverter`/`.ems`/`.meters`/`.hv_stacks`) come back empty or read the wrong device address. Consumers can't get a typed plant from a dump without reimplementing `resolve_model`/`PlantCapabilities` — i.e. coupling to our internals, which the CLI rightly refuses to do.
2. **Topology discovery isn't unit-testable** without MockPlant, because policy and I/O are entangled.

## Proposed shape

Separate policy from mechanism:

```
strategise(state) -> next register ranges to read   # pure, no I/O
probe(ranges)     -> populate caches                # the only wire I/O
# loop until strategise() proposes nothing
```

- **Offline falls out for free:** inject a complete cache as initial state; `strategise()` finds nothing left; capabilities derive from what's present. One entrypoint (`Plant.from_caches()` / `detect(offline=True)`) yields a fully-typed plant from a dump, no internals leaked.
- Subsumes the existing `prior=` hint (a prior is pre-seeded strategise state) and the `capabilities_cache` warm-start.
- `strategise` becomes a pure function over fixture caches → topology discovery unit-testable without MockPlant.

## Cache tri-state (negative knowledge)

`RegisterCache` holds only positive data, so `strategise()` can't tell "not probed yet" from "confirmed absent". Add a per-`device:bank` **present / absent / unknown** marker:

- lets `strategise()` terminate (stop re-proposing reads that always time out);
- makes an injected dump unambiguous (absent-from-dump = not present — the CLI's convention, made explicit by the library);
- enables surgical re-probe / self-healing: invalidating a `device:bank` must **reset to unknown** (drop the positive registers *and* the negative marker), so the next `strategise()` tick re-probes it — no full re-detect. Fits this hardware's transient-timeout flakiness (per-device recovery, not all-or-nothing);
- generally useful — any consumer can distinguish empty-vs-absent.

## Capabilities as the identity layer

[...]

PR #346, opened 2026-06-29
feat: generalise refresh skip-if-fresh to all banks via max_age (#268 slice 4)

### LINKED ISSUE #207: refresh(): self-tuning skip-if-fresh for IR(0,60) (auto-adapt to consumer cadence)

dewet22:
## Summary

- Extracts `_refresh_banks()` and `_refresh_ranges()` from `refresh()` as testable pure functions, paralleling how `_derive_capabilities` was extracted from `detect()` in Slice 3
- `max_age=` parameter on `refresh()` generalises the existing `ir0_max_age=` skip-if-fresh gate from IR(0,60) alone to every IR bank, using the existing `Plant.block_age()` infrastructure (#196, #207)
- `ir0_max_age=` now emits `DeprecationWarning` and maps to `max_age=`; scheduled for removal in 3.0
- Default `refresh()` (no `max_age`) is bit-identical to today — enforced by a golden-sequence characterisation test

## Test plan
- [ ] `test_refresh_hybrid_with_peripherals_golden_sequence` — pins the default (no `max_age`) wire order for a multi-peripheral HYBRID topology
- [ ] 6 `_refresh_ranges` unit tests: no-max-age includes all banks; fresh IR(0,60)/battery/meter each excluded; stale and never-seen blocks always solicited
- [ ] `test_refresh_ir0_max_age_emits_deprecation_warning` — confirms old callers get the warning and the skip still works
- [ ] 1450 tests passing, tox py311–py314 green

Closes #207

## Summary by CodeRabbit

* **New Features**
  * Added an optional freshness threshold for refreshes, so recently updated data can be skipped instead of re-read.
  * Refreshes now follow a more consistent order across inverter, battery, meter, and related device data.

* **Bug Fixes**
  * Improved detection logging and cache handling when devices are probed.
  * Kept the previous full-refresh behaviour when no freshness threshold is set.

* **Chores**
  * Added coverage for refresh ordering and freshness-based read skipping.

### DIFF
--- a/givenergy_modbus/client/client.py
+++ b/givenergy_modbus/client/client.py
@@ -281,34 +281,48 @@ def _strategise(
     Calls the same candidate helpers as ``_derive_capabilities`` so candidate generation
     has one implementation.  No I/O.
     """
+    ranges: list[ProbeRange]
+
     if step == "aio_modules":
         num = caps.bcu_stacks[0][1] if caps.bcu_stacks else 0
         addrs: list[int] | range = (
             list(prior.aio_battery_module_addresses) if prior is not None else _aio_module_candidates(num)
         )
-        return [ProbeRange("IR", addr, 60, 60, "probe") for addr in addrs]
+        ranges = [ProbeRange("IR", addr, 60, 60, "probe") for addr in addrs]
 
-    if step == "hv_bmus":
+    elif step == "hv_bmus":
         if not (caps.is_hv and caps.device_type is not Model.ALL_IN_ONE and caps.bcu_stacks):
-            return []
-        addrs = (
-            list(prior.hv_bmu_addresses)
-            if prior is not None and prior.hv_bmu_addresses
-            else _hv_bmu_candidates(caps.bcu_stacks)
-        )
-        return [ProbeRange("IR", addr, 60, 60, "probe") for addr in addrs]
+            ranges = []
+        else:
+            addrs = (
+                list(prior.hv_bmu_addresses)
+                if prior is not None and prior.hv_bmu_addresses
+                else _hv_bmu_candidates(caps.bcu_stacks)
+            )
+            ranges = [ProbeRange("IR", addr, 60, 60, "probe") for addr in addrs]
 
-    if step == "meters":
+    elif step == "meters":
         addrs = prior.meter_addresses if prior is not None else _COLD_METER_RANGE
-        return [ProbeRange("IR", addr, 60, 30, "probe") for addr in addrs]
+        ranges = [ProbeRange("IR", addr, 60, 30, "probe") for addr in addrs]
 
-    if step == "lv_bcu":
+    elif step == "lv_bcu":
         addr = prior.lv_bcu_address if prior is not None else LV_BCU_ADDRESS
         if addr is None:
-            return []
-        return [ProbeRange("IR", addr, 60, 60, "probe")]
+            ranges = []
+        else:
+            ranges = [ProbeRange("IR", addr, 60, 60, "probe")]
 
-    raise ValueError(f"_strategise: unknown step {step!r}")
+    else:
+        raise ValueError(f"_strategise: unknown step {step!r}")
+
+    _logger.debug(
+        "_strategise(%s, prior=%s): %d range(s) → %s",
+        step,
+        "hinted" if prior is not None else "cold",
+        len(ranges),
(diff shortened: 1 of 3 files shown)


tests/client/test_refresh_resilience.py
```diff
@@ -337,3 +337,112 @@ async def test_refresh_default_always_solicits_ir0():
...
+
+
+def test_refresh_ranges_no_max_age_includes_all_banks():
+    """With max_age=None every bank for a HYBRID plant is included."""
+    from givenergy_modbus.client.client import _refresh_ranges
+
+    plant, caps = _client_with_caps(Model.HYBRID, lv_battery_addresses=[0x33]).plant, None
+    client = _client_with_caps(Model.HYBRID, lv_battery_addresses=[0x33])
+    plant = client.plant
+    caps = client.plant.capabilities
```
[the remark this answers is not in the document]
dewet22: Good catch — removed in 0e87c21.

dewet22-codex:

givenergy_modbus/client/client.py
```diff
@@ -1047,6 +1136,7 @@ async def refresh(
         timeout: float = 2.0,
         retries: int = 1,
         retry_delay: float = 0.5,
+        max_age: float | None = None,
```
dewet22-codex: [R] This changes the fourth positional argument from ir0_max_age to max_age: existing callers using await refresh(timeout, retries, retry_delay, 30) now skip every fresh IR bank and do not see the deprecation warning. Since Client.refresh() is consumed by hass/cli and ir0_max_age is only deprecated rather than removed, I think the legacy positional slot needs to remain ir0_max_age and the new max_age should be keyword-only or appended after it.

dewet22-codex (changes requested):
I found one blocking compatibility issue around the public Client.refresh() signature. Apart from that, the helper extraction and all-bank freshness gate looked consistent to me. I ran uv run --group test pytest tests/client/test_load_refresh.py tests/client/test_refresh_resilience.py locally: 54 passed.

givenergy_modbus/client/client.py
dewet22: Correct — fixed in 361973e: `ir0_max_age` stays as the 4th positional slot, `max_age` is now keyword-only via `*`.

dewet22-codex:
I’ve started a scoped re-review of the positional-argument compatibility fix.

dewet22-codex (approved):
I’ve re-reviewed the compatibility fix and the outstanding blocker is resolved: ir0_max_age remains in the legacy positional slot, with max_age keyword-only. Focused tests passed locally: uv run --group test pytest tests/client/test_load_refresh.py tests/client/test_refresh_resilience.py.

PR #347, opened 2026-06-29
feat: skip detect-confirmed-absent banks in refresh() (#268 slice 5)

dewet22:
## Summary

Final slice of the #268 epic. `_refresh_ranges` now reads the tri-state presence marker that Slice 1 (#344) introduced: a bank that `detect()` marked ABSENT (`Plant.block_present()` returns `False`) is skipped **before** the `max_age` freshness gate, unconditionally. The presence marker is a stronger, cheaper signal than a 2s timeout — a known-absent device should never be re-solicited every poll cycle. Re-discovery stays `detect()`'s job (or `invalidate_presence()` → UNKNOWN); refresh polls what's known-present.

This closes the loop on the marker: `mark_absent` now has both a writer (every detect probe-failure site, Slice 1) and a reader (the refresh poll path).

The `max_age is None` fast path folds into one unified loop — output is **bit-identical for real topologies**, since `caps` never lists an absent bank in the normal flow (absent peripherals aren't in caps). The skip only fires when a caps bank and a `False` presence marker coexist — the capabilities-as-target direction (optimistic/restored caps vs live reality).

## Test plan
- [ ] `test_refresh_ranges_skips_absent_bank_even_without_max_age` — absent bank dropped on the default path; present sibling still polled
- [ ] `test_refresh_ranges_skips_absent_bank_regardless_of_freshness` — absent skip applies with `max_age` set too (presence beats freshness)
- [ ] `test_refresh_ranges_includes_present_bank` — explicitly-PRESENT bank not over-skipped
- [ ] `test_refresh_does_not_solicit_absent_device` — end-to-end: `refresh()` never puts the absent device on the wire
- [ ] Golden-sequence characterisation tests unchanged (no banks marked absent in those fixtures)
- [ ] 1454 tests passing, tox py311–py314 green

Completes #268.

### DIFF
--- a/givenergy_modbus/client/client.py
+++ b/givenergy_modbus/client/client.py
@@ -362,31 +362,33 @@ def _refresh_ranges(
     *,
     now: datetime | None = None,
 ) -> list[TransparentRequest]:
-    """Return the TransparentRequests for one refresh cycle, skipping fresh banks.
+    """Return the TransparentRequests for one refresh cycle, skipping absent and fresh banks.
 
-    When ``max_age`` is None every bank is included (bit-identical to the pre-#268
-    behaviour). When set, any bank whose ``plant.block_age()`` is not None and ≤
-    ``max_age`` seconds is omitted. No I/O.
+    A bank that detect marked ABSENT (``plant.block_present()`` is False) is skipped
+    unconditionally — the presence marker is a stronger, cheaper signal than a timeout,
+    so a known-absent device is never re-solicited (call ``detect()`` or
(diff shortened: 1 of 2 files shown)


dewet22-codex:

dewet22-codex (approved):
I’ve reviewed this final slice and have no concerns. The absent-marker gate is scoped to exact refresh-bank keys, keeps the present/unknown paths polling normally, and the default sequence remains covered by the existing characterisation tests. I ran uv run --group test pytest tests/client/test_refresh_resilience.py tests/client/test_load_refresh.py tests/client/test_detect.py locally: 125 passed.

Issue #352, opened 2026-06-30
Retire the caps-absent 0x32→inverter legacy; cleanly separate inverter (0x11) from battery (0x32)


dewet22:
## Context

Surfaced while fixing #350. With `PlantCapabilities` present the addressing is clean — inverter at 0x11 (0x31 facade on AC/HYBRID_GEN1), LV battery packs at 0x32–0x37. But when `capabilities is None` there's residual #119/#189 legacy that still treats 0x32 as the inverter:

- `Plant._getter_for_device_address` maps `0x11/0x31/0x32 → SinglePhaseInverterRegisterGetter` when caps are None, so a 0x32 battery bank validates under inverter rules and skips the battery splice guard.
- `Plant.inverter` reads `register_caches[0x32]` directly when caps are None.
- `PlantCapabilities.inverter_address` defaults to `0x32` — a vestige (`_derive_inverter_address` rewrites it to 0x11, so the default never survives normalisation, but it shouldn't be 0x32).

This was the enabling condition for #350: a fresh `Plant` mid-reconnect read 0x32 (a battery) while caps were None, routed it to the inverter getter, and bypassed the splice guard. #350 closed the safety hole (an internally-impossible frame is now refused at `_commit_bank` regardless of getter), but the addressing muddle itself remains.

## Proposal

Make the separation unconditional:

- caps-None: route `0x32–0x37 → BatteryRegisterGetter`, keep `0x11/0x31 → inverter`.
- `Plant.inverter` (caps-None): read the inverter address, not 0x32 — or return an empty-cache model, matching the `.ems` / `.gateway` accessors.
- Default `PlantCapabilities.inverter_address` to 0x11 (or make it required), dropping the 0x32 vestige.

## Why this is separate from #350

It changes the `Plant.inverter` caps-absent contract and touches the addressing-from-captures / `from_caches` / replay tests, plus the pre-#189 capture-replay path (where 0x32 genuinely carried inverter data). That warrants its own review rather than riding the battery-safety patch. The only consumers of the legacy mapping are bare-Plant replay tooling and pre-#189 captures; live use always has capabilities after `detect()`.

## Acceptance

- A caps-None 0x32 read validates as a battery (splice-guarded, cold-start-held).
- `plant.inverter` no longer reads 0x32 without capabilities.
- pre-#189 capture-replay either still works or is explicitly dropped with a note in the changelog.

dewet22:
[...]

Issue #350, opened 2026-06-30
Splice-guard baseline poisoning on primary LV battery after reconnect — healthy frames permanently rejected (2.9.x regression)


dewet22:
## Summary
On 2.9.2, the **primary** LV battery (0x32) went permanently offline after the inverter's ~02:00 cloud-sync reconnect. The splice-guard "last-good" baseline for 0x32 was seeded from an internally-inconsistent frame — cell-voltage and cell-temp registers all **zero**, but firmware and capacity **correct** — so every subsequent healthy read is rejected as a physics-impossible delta against the zero baseline. The pack never recovers: good data can't overwrite the poisoned baseline. Not seen on 2.8.0 or earlier; appears to be a regression in the 2.9.x detect/reconnect rework (#268).

## Environment
- givenergy-modbus 2.9.2 (via givenergy-hass v1.3.32rc1), `Model.HYBRID_GEN1`
- 2× LV packs: **0x32 = `BG2134G…`** (fw 3007, **primary**), 0x33 = `DZ2228G…` (fw 3009, secondary)
- Trigger: the inverter's daily ~02:00 cloud reconnect (old-firmware behaviour) — forces a fresh detect/refresh cycle

## Evidence
Held baseline is internally impossible — zero cells/temps alongside correct fw + capacity:
```
Rejected battery bank for device 0x32 — sub-bus splice (>=2 physics-impossible deltas); keeping last-good.
Trips: IR(68) cell_mV 0->3269 … IR(75) 0->3272, IR(103) cell_temp_deci 0->240, IR(104) 11->231, IR(88) cap_centiAh 14987->13210.
```
```
Battery bank for device 0x32: constant register(s) IR(98) 3007->2048 disagree with baseline — holding last-good, will adopt only after 900 s of sustained agreement. Trips: IR(98) IMMUTABLE 3007->2048 …
```
Per-device counters since the 02:02 reconnect:
- `splice_guard_rejections`: 0x32 = 2682 (climbing every ~13 s), 0x33 = 19
- `cold_start_holds`: 0x33 = 3, **0x32 = 0**
- `splice_guard_holds`: 0x32 = 117, 0x33 = 26

0x32 went `unavailable` at 02:09:18 and has not recovered; 0x33 healthy throughout. Both packs were healthy on the fresh install/detect the day before — the poisoning is reconnect-specific.

## Leads (maintainer)
[...]

Issue #189, opened 2026-06-05
feat(model): retire 0x31 read-alias; unify inverter addressing on 0x11 for AC/HYBRID_GEN1


dewet22:
## Background

During investigation of #187 (write-register echo cache bug on HYBRID_GEN1) we gathered substantial live-hardware and app-protocol evidence that calls the `0x31` read-alias into question. This issue documents that evidence and tracks the cleanup when the remaining gap (live AC validation) can be filled.

## The claim that needs correcting

`inverter_address_for()` in `givenergy_modbus/model/inverter.py` contains this docstring:

> The `0x31` special-case is load-bearing, not a hedge: the `hybrid_2_bat_a` fixture (HYBRID_GEN1, ARM 449) **answers identity-only at `0x11`** and serves its full register banks at `0x31`, so the generic `0x11` default would silently lose most banks for these models.

This claim is **false**, and was inferred from passive dongle captures, not from actively querying `0x11`.

## Evidence that 0x11 is a full facade on HYBRID_GEN1

All tests run against a live HYBRID_GEN1 unit at ARM 449. All reads were done multiple times to account for contention.

**1. Direct HR read at 0x11:** `probe_read_address.py` (extended in #187) asked `0x11` for HR(0,60) and HR(60,60) — both returned 4/4 responses. Same result at `0x31`. The "identity-only" claim is therefore false: `0x11` serves the full config bank when asked.

**2. Value-equality:** After reading, the facade check compared HR(0)–HR(119) between the two addresses (excluding HR(35-40) = system_time RTC, which ticks between sequential reads). Result: **"0x11 mirrors 0x31 across all 114 shared HR registers — full facade."**

**3. Write round-trip:** Writing CHARGE_TARGET_SOC (HR 116) to `0x31` mutated the hardware (29 → 30 confirmed by a fresh `load_config` read). Writing to `0x11` also mutated it. The hardware accepts writes at both addresses.

**4. Official app behaviour (HYBRID_GEN1, two firmware versions):**
- Writes: always to `0x11` — registers 20 and 27 captured at firmware 449 and 451 via MockPlant. Firmware-invariant.
- Reads: primarily `0x11` — HR(116) read at `0x11` on firmware 449 (single-register read `HR 116..116`) and at `0x11` on firmware 451 (block read `HR 60..119`). One initial read on firmware 451 went to `0x31` first, suggesting the app may do a coarse bulk read at `0x31` before switching to targeted `0x11` reads.

[...]

PR #353, opened 2026-06-30
test: address inverter-bank tests at 0x11, not 0x32 (#352 prep)

dewet22:
Preparatory test-suite change for #352 (retiring the caps-absent `0x32 → inverter` addressing legacy). No production code changes.

## Why
The test suite has long used device address `0x32` as a generic "inverter device" — a habit from the pre-#189 era when the inverter answered there. In the current model `0x32` is LV battery pack #1 and the inverter's canonical address is `0x11`. #352 makes that separation unconditional (a caps-absent `0x32` read routes to the battery getter, closing the #350 seeding window), which breaks ~34 tests that feed inverter-shaped banks at `0x32` and expect an immediate commit. Splitting the mechanical churn out first keeps the #352 change small and reviewable.

## What
Behaviour-neutral on `main`: both `0x11` and `0x32` currently route to the inverter getter, so these moves change nothing today.
- **`test_plant.py`** — 32 cache-mechanics tests now feed and assert at `0x11`. Helper defaults are untouched; only call sites gain an explicit `device_address=0x11`.
- **`conftest.py`** — the two `CLIENT_MESSAGES` inverter-bank PDUs regenerated to `device_address=0x11`, re-encoded from the library so the frame bytes and CRC are authoritative rather than hand-edited.

## Verification
- Full suite green as-is (1461 passed) — confirms behaviour-neutrality.
- With #352's production diff applied on top, failures drop from 34 to exactly 3 (the two bare-Plant `.inverter` golden tests and the non-battery-device splice test), all of which change meaning under #352 and travel with it.
- `tox` (py311–py314 + lint) green.

## Summary by CodeRabbit

* **Tests**
  * Updated register and freshness test coverage to use the inverter-facing device address consistently.
  * Refined checks for cache updates, stale-content tracking, presence handling, CRC-failure behaviour, and timestamp normalisation.
  * Expanded scenarios to confirm register-age calculations and block status remain correct across overlapping windows, repeated values, and timezone-naive datetimes.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -313,10 +313,10 @@ _server_messages: PduTestCases = [
 # Messages a client should be expected to process (or, typical messages a server would send)
 _client_messages: PduTestCases = [
     (
-        "2:4/ReadInputRegistersResponse(device_address=0x32 base_register=0)",
+        "2:4/ReadInputRegistersResponse(device_address=0x11 base_register=0)",
         ReadInputRegistersResponse,
         {
-            "check": 0x172D,
+            "check": 0x3250,
             "inverter_serial_number": "SA1234G567",
             "base_register": 0x0000,
             "register_count": 0x003C,
@@ -386,7 +386,7 @@ _client_messages: PduTestCases = [
             # fmt: on
             "data_adapter_serial_number": "WF1234G567",
             "padding": 0x8A,
-            "device_address": 0x32,
+            "device_address": 0x11,
             "error": False,
             "crc_failed": False,
         },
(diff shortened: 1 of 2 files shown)


dewet22-codex:

dewet22-codex (approved):
I’ve reviewed this with #352 in mind and have no blocking concerns. The changed call sites move the inverter-bank/cache-mechanics coverage onto 0x11, and the regenerated client-message fixtures decode cleanly with the new device address and CRCs. I checked the remaining 0x32 uses in test_plant.py and they look intentional for battery, legacy, or address-specific cases rather than missed inverter-bank fixtures.

I ran uv run --group test pytest locally: 1461 passed, with the existing deprecation warnings.

Issue #50, opened 2026-05-11
Reader/writer tasks log CRITICAL on intentional client.close()


dewet22:
## Summary

`client.py:148` logs `"network_consumer reader at EOF, cannot continue"` at CRITICAL when the reader task ends, and `client.py:160` logs `"network_producer writer is closing, cannot continue"` at CRITICAL when the writer task ends.

Both fire on the normal exit path triggered by `client.close()`, so any short-lived usage (one-shot polls, scripted register dumps, the TUI exiting cleanly) emits CRITICAL log lines immediately after a successful operation. CRITICAL should signal something that needs attention; "I closed cleanly" doesn't.

## Repro

```python
client = Client(host="…", port=8899)
await client.connect()
await client.refresh_plant(full_refresh=True)
await client.close()
# CRITICAL  givenergy_modbus.client.client  network_consumer reader at EOF, cannot continue
```

## Suggested fix

Track whether shutdown was initiated by `close()` (e.g. a `self._shutting_down` flag set before cancelling/closing the tasks). On the expected path, downgrade the message to DEBUG (or drop it entirely). Reserve CRITICAL for the surprise-EOF / writer-died case where the inverter actually goes away mid-flight.

dewet22:
Fixed in #52 (merged 0c1a69f). Reader and writer tasks now log at DEBUG when exiting via an intentional `client.close()`. CRITICAL is preserved only for the genuine surprise-EOF / writer-died case.

PR #357, opened 2026-07-02
fix: graceful connection teardown + typed ConnectionLost on unexpected drops (#356)

### LINKED ISSUE #356: Graceful producer/consumer teardown on connection loss — clean self-teardown + distinguish "connection lost" from a genuinely-stuck producer
## Summary

The client's producer/consumer task teardown on an *unexpected* connection drop is untidy, and one error message conflates two unrelated failure modes. Surfaced by the givenergy-hass 2.9.4 soak (a routine nightly peer-initiated reconnect logging loudly) and related to the field-reported producer stall on comms-noisy EMS plants (givenergy-hass#233).

The **severity** of the drop logs was just addressed (`bc12802`: the peer-drop teardown logs went CRITICAL → WARNING). This issue is the **structural** follow-up.

## Current lifecycle (for reference)

- **Producer** (`_task_network_producer`): `while writer and not writer.is_closing()` → pull `(frame, frame_sent, response_future)` from `tx_queue` → `writer.write` + `await writer.drain()` → signal `frame_sent`. On loop exit: WARNING + `connected=False` (or DEBUG if `_shutting_down`).
- **Consumer** (`_task_network_consumer`): `while reader and not reader.at_eof()` → `reader.read(300)` → decode → `plant.update` → resolve futures. Same exit branch.
- **Send** (`send_request_and_await_response`): `tx_queue.put` (5 s cap → `"TX queue full — producer likely died"`) → await `frame_sent` (scaled cap → `"Producer task is stuck — frame not sent"`) → await `response_future` (retry/give up).
- **`close()`**: sets `_shutting_down`, cancels queued `frame_sent` futures, cancels both tasks, closes writer/reader.
- **Reconnect policy lives in the consumer** (the hass coordinator): it watches `connected` and calls `connect()` (idempotent, tears down prior state). The lib does **not** self-reconnect.

## The core problem: two conditions, one error

`"Producer task is stuck — frame not sent"` (and `"TX queue full — producer likely died"`) fire for two unrelated situations:

1. **Connection lost** — the producer loop *exited cleanly* because `writer.is_closing()`, and a frame queued afterward has no one to send it. Routine (the nightly dongle cloud-sync TCP drop). The producer isn't "stuck" — it's *gone*; the caller gets a scary "stuck" error for a benign disconnect.
2. **Genuinely wedged** — the producer is blocked in `await writer.drain()` on a half-open socket that hasn't fully closed. `drain()` has **no write-side timeout**, so the task stays wedged until `close()` cancels it. This is the givenergy-hass#233 field symptom on comms-noisy EMS plants.

## Gaps

1. **No clean self-teardown on drop.** A task exiting from a peer-drop sets `connected=False` but leaves a *zombie half-state* — dead tasks, writer/reader still attached, frames possibly stranded in `tx_queue` — until the coordinator happens to call `connect()`.
2. **Misleading error taxonomy.** "stuck" / "queue full — producer died" usually mean *connection lost*. There's no typed `ConnectionLost`-style signal to separate "reconnect me" from "this is a real bug."
3. **Unbounded `writer.drain()`** — the one place a producer can genuinely wedge, with no write-side timeout.
4. **Stranded queued frames** — their `frame_sent` futures are only cancelled by `close()`; on a pure drop they hang for the full `frame_sent_timeout`.
5. **Asymmetric teardown** — a dying writer doesn't signal the reader (or vice versa); on a half-open socket they can diverge, one side blocking on `read(300)` while the other's already gone.

## Proposed direction (design latitude)

- **One clean teardown path** on any unexpected drop: cancel both tasks, fail stranded `tx_queue` frames fast, flip `connected`.
- **A typed `ConnectionLost`-style signal** so `send_request_and_await_response` reports "connection lost" instead of "stuck" — letting callers distinguish a reconnect-me condition from a genuine bug.
- **Bound `writer.drain()`** (write-side timeout) so the genuinely-wedged case (Mode 2 / #233) self-resolves rather than hanging until `close()`.
- **Symmetric teardown** so one side dying tears down the other.

## Non-goal

The library does **not** gain auto-reconnect — reconnect policy stays the coordinator's (it owns the failure-count / backoff logic). This is about *clean teardown + clear signalling* only.

**Caveat:** if the teardown/signalling changes here *directly* alter reconnect semantics — e.g. how or when the coordinator must call `connect()`, or `connect()`'s idempotent prior-teardown contract — keeping that consistent is in-scope. The line is: don't *add* a reconnect policy; do keep the existing one working.

## Boundaries / relationships

- **`bc12802` (shipped):** CRITICAL → WARNING on the drop logs — fixed the *loudness*; this fixes the *structure*.
- **givenergy-hass#233:** the field symptom of Mode 2 (producer stall on a comms-noisy EMS plant).
- **#274 (connect()+detect() atomicity):** must not regress — a fresh `connect()` tears down prior state; the new teardown path has to compose cleanly with that.

## Acceptance sketch

- A peer-initiated drop yields a clean teardown and a `ConnectionLost`-typed error from any in-flight send (not "stuck"); the coordinator reconnects exactly as it does today.
- A wedged `drain()` self-resolves within a bounded time — no producer task left hung until `close()`.
- No regression to connect()+detect() atomicity (#274) or the intentional-shutdown DEBUG path.

dewet22:
Closes #356. Graceful producer/consumer teardown on unexpected connection loss, plus a typed `ConnectionLost` signal — the structural follow-up to the log-severity trim (`bc12802`).

## Problem
On an unexpected drop (e.g. the dongle's nightly ~02:00 cloud-sync reconnect) the network tasks each noticed their own side dying and exited, leaving a zombie half-state until the next `connect()`: the sibling task running against dead state, queued frames stranded for the full scaled timeout, and in-flight senders failing slowly with misleading errors that conflated *connection lost* (routine — reconnect) with *genuinely wedged producer* (a real bug). The one true wedge — `writer.drain()` on a half-open socket — could hang unboundedly (the field symptom on the comms-noisy EMS plant, givenergy-hass#233).

## Changes
- **`ConnectionLost(CommunicationError, TimeoutError)`** — the dual base is a deliberate compatibility contract: consumers catching bare `TimeoutError` (the historical error type for a dead connection) see no behaviour change; catching `ConnectionLost` first opts into immediate-reconnect policy.
- **`Client._abort_connection()`** — one idempotent teardown path run by whichever task notices death: fails all in-flight and queued futures with `ConnectionLost` (waiters unblock in ms, not after ~10s+ timeouts), cancels the sibling task, flips `connected`. Deliberately leaves reader/writer/task attributes to `close()`/`connect()` (#274 atomicity preserved). Intentional shutdown early-returns — the #50 DEBUG paths are untouched.
- **Bounded `writer.drain()`** (`_DRAIN_TIMEOUT = 10 s`) — a stall means the peer stopped ACKing; treated as connection loss instead of wedging until `close()`.
- **Send path** — fail-fast `ConnectionLost` when the connection is known-dead (a dead-window refresh batch fails in ms); `except ConnectionLost` arms ordered before `except TimeoutError` (it *is* a `TimeoutError` — order is load-bearing, regression-tested); the `frame_sent` timeout becomes an honest "producer wedged = real bug" signal that now also triggers teardown so the system recovers.
- **No auto-reconnect** — reconnect policy stays with the consumer/coordinator; `connect()` remains the reconnect primitive, unchanged.

Design spec was maintainer-reviewed before implementation; the two planning-time refinements (guard keyed on a `_connection_lost` flag rather than `connected`, so never-connected clients are unaffected; the drain-stall handler failing the already-dequeued frame's futures itself) are documented in the plan.

## Tests
12 new tests: exception typing/compat, abort semantics (idempotence, shutdown early-return, sibling-only cancellation), drain-stall teardown, consumer-EOF fast unblock, fail-fast guard, retry-ordering, stuck-producer teardown, and a dedicated replacement for the #225 identity-guard regression (its old test was legitimately repurposed for the new stuck-producer semantics). Three pre-existing tests that used EOF as inert scaffolding now set `_shutting_down` (EOF is no longer a no-op by design); the unexpected-EOF path keeps independent coverage.

## Verification
Full suite 1477 passed; tox matrix (py311–py314 + format + lint + build) green.

Consumer note: after release I'll drop a note to the hass integration — catching `ConnectionLost` ahead of `TimeoutError` would let it reset/reconnect on the first drop rather than after the 3-strike timeout tolerance; that's its policy call.

### DIFF
--- a/givenergy_modbus/client/client.py
+++ b/givenergy_modbus/client/client.py
@@ -1572,10 +1572,10 @@ class Client:
                 if frame_sent and not frame_sent.done():
                     frame_sent.set_result(True)
                 continue
-            self.writer.write(message)
-            if self._capture_sink is not None and self._capture_redactor_tx is not None:
-                self._emit_to_sink("tx", self._capture_redactor_tx.feed(message))
             try:
+                self.writer.write(message)
+                if self._capture_sink is not None and self._capture_redactor_tx is not None:
+                    self._emit_to_sink("tx", self._capture_redactor_tx.feed(message))
                 await asyncio.wait_for(self.writer.drain(), timeout=_DRAIN_TIMEOUT)
             except TimeoutError:
                 _logger.warning(
@@ -1583,6 +1583,19 @@ class Client:
                     _DRAIN_TIMEOUT,
                 )
                 exc = ConnectionLost("writer drain stalled — connection lost")
+                # This frame is already dequeued, so the teardown's queue-drain
+                # can't reach it — fail its futures here.
+                for fut in (frame_sent, response_future):
+                    if fut is not None and not fut.done():
+                        fut.set_exception(exc)
+                self.tx_queue.task_done()
+                self._abort_connection(exc)
+                return
+            except OSError as e:
+                _logger.warning(
+                    "network_producer: socket error during write/drain (%s) — treating connection as lost", e
+                )
+                exc = ConnectionLost(f"socket error during write/drain — connection lost: {e}")
                 # This frame is already dequeued, so the teardown's queue-drain
                 # can't reach it — fail its futures here.
                 for fut in (frame_sent, response_future):
@@ -1684,6 +1697,13 @@ class Client:
             except TimeoutError as exc:
                 _discard(response_future)
                 raise TimeoutError("TX queue full — producer task has likely died") from exc
+            if self._connection_lost:
+                # Lost the race with _abort_connection's queue-drain: our frame was
+                # enqueued after the drain and will never be sent. _discard cancels the
+                # response future, so a post-reconnect producer skips the stale frame
+                # (the queue-front skip-if-resolved check).
+                _discard(response_future)
+                raise ConnectionLost("connection lost while enqueueing — reconnect before sending")
             # Safety-net wait for the producer to actually send this frame. Worst case the
             # frame sits behind a full queue, and the producer sleeps tx_message_wait + up to
             # tx_jitter (plus a drain) between sends — so scale the bound by the full queue
--- a/givenergy_modbus/client/client.py
+++ b/givenergy_modbus/client/client.py
@@ -442,9 +442,12 @@ class Client:
     - Take a per-client lock around ``refresh_plant()`` so successive polls don't
       overlap. Writes don't need the same lock — they're free to land between
       polls.
-    - Connection loss is surfaced via ``self.connected`` flipping to ``False``;
-      the reader/writer tasks log a WARNING when this happens (a peer-initiated
-      drop is recoverable via reconnect). ``connect()`` is
+    - Connection loss is surfaced three ways: ``self.connected`` flips to
+      ``False``, the noticing task logs a WARNING, and every in-flight or
+      subsequently attempted request raises ``ConnectionLost`` (a
+      ``CommunicationError`` that is also a ``TimeoutError``, so legacy
+      ``except TimeoutError`` handling keeps working — catch ``ConnectionLost``
+      first to distinguish reconnect-me from a genuine stall). ``connect()`` is
       idempotent and tears down the previous connection on its own, so it can
       be called directly as a reconnect primitive.
     """
--- a/givenergy_modbus/client/client.py
+++ b/givenergy_modbus/client/client.py
@@ -1545,6 +1545,7 @@ class Client:
         else:
             self.connected = False
             _logger.warning("network_consumer: connection lost (reader at EOF)")
+            self._abort_connection(ConnectionLost("reader at EOF — connection lost"))
 
     async def _task_network_producer(self):
         """Producer loop to transmit queued frames with an appropriate delay.
--- a/givenergy_modbus/client/client.py
+++ b/givenergy_modbus/client/client.py
@@ -129,6 +129,12 @@ _FRAME_MARKER = bytes.fromhex("59590001")
 # it sane when the queue is idle. Module-level so tests can shrink it.
 _FRAME_SENT_MIN_TIMEOUT = 5.0
 
+# Upper bound on writer.drain() inside the producer loop (#356). On a healthy link
+# drain completes near-instantly (it's local socket-buffer backpressure); a stall
+# means the peer stopped ACKing — a half-open connection — so it's treated as
+# connection loss rather than left to wedge the producer until close() (hass#233).
+_DRAIN_TIMEOUT = 10.0
+
 
 class FrameRedactor:
     """Frame-aware stateful redactor for a captured GivEnergy byte stream.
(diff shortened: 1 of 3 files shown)


dewet22-codex:
I’ve started a review pass on this PR, including issue #356 as the framing context, and I’ll watch it for follow-up changes.

givenergy_modbus/client/client.py
```diff
@@ -1519,7 +1575,22 @@ async def _task_network_producer(self):
             self.writer.write(message)
             if self._capture_sink is not None and self._capture_redactor_tx is not None:
                 self._emit_to_sink("tx", self._capture_redactor_tx.feed(message))
-            await self.writer.drain()
+            try:
+                await asyncio.wait_for(self.writer.drain(), timeout=_DRAIN_TIMEOUT)
```
dewet22-codex, line 1579: R: This catches the bounded-drain timeout, but a socket error from write/drain still bypasses the shared teardown. I reproduced this with writer.drain raising ConnectionResetError: _task_network_producer raises the raw exception, does not call task_done for the dequeued frame, and leaves frame_sent/response_future unresolved until callers hit the old frame_sent timeout path. That seems to miss the #356 contract for a peer reset, so I think the write/drain OSError/ConnectionError path needs to fail the current frame futures and call _abort_connection just like the timeout branch.

givenergy_modbus/client/client.py
```diff
@@ -1580,6 +1652,9 @@ async def send_request_and_await_response(
         original "retry immediately" behaviour (e.g. fast probes, latency-
         sensitive interactive commands) should pass ``retry_delay=0``.
         """
+        if self._connection_lost:
```
dewet22-codex, line 1668: R: This guard covers sends that start after the abort, but there is still a race when a sender is already blocked in tx_queue.put() at the moment another task calls _abort_connection(). The abort drains the queue and marks the connection lost, which wakes this putter; it then enqueues a fresh frame after the drain has finished, waits on frame_sent, and eventually raises "Producer task is stuck" instead of ConnectionLost. My temporary repro used maxsize=1, filled the queue, started a send, then called _abort_connection(); it took the stuck path and left a dead frame in the queue. I think the send path needs to re-check _connection_lost after the put resumes, or otherwise make the queue-drain/put race consume that frame with ConnectionLost.

dewet22-codex (changes requested):
Thanks, I read this against #356 and the main shape looks aligned with the intended no-auto-reconnect contract. I found two blocking edge cases in the teardown boundary, both around paths that can still leave callers on the old slow or raw-error behaviour instead of the typed ConnectionLost signal. Focused checks passed locally: uv run --group test pytest tests/client/test_client.py tests/client/test_detect.py (118 passed). The two temporary regressions I used fail against this head.

givenergy_modbus/client/client.py
```diff
@@ -1580,6 +1652,9 @@ async def send_request_and_await_response(
...
+            raise ConnectionLost("connection lost — reconnect before sending")
```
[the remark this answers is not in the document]
dewet22-claude: Fixed in 0c7cf56 — good catch. The write + capture-emit + bounded drain now sit in one `try:` with an `except OSError` sibling to the stall handler: it fails the dequeued frame's futures with `ConnectionLost`, calls `task_done()`, and runs `_abort_connection` instead of propagating raw. Regression test `test_producer_socket_error_during_drain_aborts_connection` reproduces your `ConnectionResetError` case and was red before the fix.

givenergy_modbus/client/client.py
dewet22-claude: Fixed in 0c7cf56. The send path re-checks `_connection_lost` immediately after `tx_queue.put()` resumes: a putter woken by the abort's queue-drain now discards its response future and raises `ConnectionLost` instead of riding the stuck-producer path. The stale frame is inert — `_discard` cancels the response future, so a post-reconnect producer skips it via the queue-front skip-if-resolved check. Regression test `test_send_blocked_in_full_queue_raises_connection_lost_on_abort` uses your maxsize=1 repro shape and was red before the fix.

dewet22-codex:
I’ve picked up the new head and am doing the scoped re-review now.

dewet22-codex (approved):
I’ve re-reviewed the new head (`0c7cf56`) and the two teardown blockers I raised are resolved. The producer now treats socket errors during write/drain as connection loss, failing the current frame futures and running `_abort_connection`, and the send path now re-checks `_connection_lost` after a blocked `tx_queue.put()` resumes so that race raises `ConnectionLost` instead of the stuck-producer timeout. The new regressions cover both shapes, and the focused local run passed: `uv run --group test pytest tests/client/test_client.py::test_producer_socket_error_during_drain_aborts_connection tests/client/test_client.py::test_send_blocked_in_full_queue_raises_connection_lost_on_abort tests/client/test_client.py tests/client/test_detect.py` (120 passed). I have no blocking concerns remaining from this scope.

PR #359, opened 2026-07-02
fix: tolerate silence at 0x32 in the detect LV preamble (#358)

### LINKED ISSUE #358: detect() aborts on gateway-class plants — the mandatory 0x32 LV-battery preamble is fatal when nothing answers
First live Gateway config-flow attempt (AberDino's Gen1 Gateway fronting 2× AIO, via givenergy-hass#95): `detect()` resolves identity fine (`HR(0)=0x7001 → Model.GATEWAY`) and then hard-aborts in the LV battery step. Reproduced twice in the HA debug log.

The mechanism: `_detect_lv_batteries` opens with an unconditional `ReadInputRegisters(device_address=0x32, base_register=60)` sent via `send_request_and_await_response` with the full known-tier treatment (timeout + retries), not the tolerant `_probe` path used for `0x33–0x37`. Behind a Gateway nothing answers at `0x32` — the batteries are register-embedded per-AIO in the gateway rollup (`aio1/2_soc`, per-AIO charge/discharge counters), not separately-addressed LV packs — so the read exhausts its retries and the resulting `TimeoutError` propagates out of `detect()`. The consumer sees "cannot connect" despite a healthy connection and successful identification.

This is the last survivor of the mandatory-`0x32` assumption family that #352 retired on the routing side: the preamble still *assumes* an LV battery exists at `0x32` on any plant that reaches the LV step. The assumption isn't even safe on hybrids — a hybrid with no battery pack attached is a perfectly valid AC/PV-only plant, and it would hit the same abort.

Proposed direction (some latitude): route the `0x32` preamble through the tolerant probe tier and `mark_absent` on silence, mirroring how `0x33–0x37` already behave — an empty `lv_battery_addresses` is a perfectly valid outcome. Gating the LV step off entirely when identity resolved a gateway/HV-class model would be a reasonable adjunct, but the tolerant-probe fix is the general one (it covers the battery-less hybrid too).

Supporting evidence: a ~10-minute passive wire capture from the same Gateway (GivTCP polling it every 30 s) validates the GatewayV1 register map against live silicon for the first time — the decode yields coherent values across 46 fields including the per-AIO split (per-AIO charge todays sum exactly to the battery total), so once `detect()` completes, a Gateway entry has real data behind it. Capture and HA debug log are linked from givenergy-hass#95; a daylight capture is expected to follow, which I'll fold into the fixture corpus.

dewet22:
Closes #358. `detect()` no longer aborts on plants where nothing answers at `0x32` — gateways (battery data register-embedded in the rollup) and battery-less hybrids (a valid AC/PV-only configuration).

## Problem
`_detect_lv_batteries` opened with an unconditional known-tier read of `IR(60,60)@0x32`; total silence exhausted the retries and the bare `TimeoutError` propagated out of `detect()`, so the consumer saw "cannot connect" despite a healthy connection and successful identification. Surfaced by the first live Gateway config-flow attempt (givenergy-hass#95, reproduced twice in the HA debug log). This was the last survivor of the mandatory-`0x32` assumption family that #352 retired on the routing side.

## Changes
- `batt_candidates` is computed first and the `0x32` preamble only runs when `0x32` is a candidate — a hinted gateway reconnect (prior `lv_battery_addresses=[]`) skips the dead read entirely instead of stalling the full known-tier timeout on every reconnect.
- Silence at `0x32` is a valid outcome: `except TimeoutError` → `mark_absent(0x32)` + drop the pre-seeded cache; an empty `lv_battery_addresses` then derives naturally. `except ConnectionLost: raise` is ordered above it — a dead connection is not an absent battery (the #356 dual-base ordering trap, regression-tested).
- The preamble keeps its patient known-tier treatment on battery plants (deliberate — the #350 family); `warn_timeout=False` so expected absence doesn't pollute the per-device retry counters, matching the `0x33+` probe semantics.

## Evidence & validation
The accompanying wire capture from the same Gateway (GivTCP polling it every 30 s) gave the GatewayV1 register map its first live-traffic validation: `HR(0)=0x7001 → Model.GATEWAY`, 46 fields decoding to coherent physics including an exact per-AIO energy split (per-AIO charge todays sum precisely to the battery total). A daylight capture is expected and will be folded into the fixture corpus separately.

## Tests
Field-abort repro (identity `0x7001` → `detect()` completes, `0x32` ABSENT, `lv_battery_addresses=[]`), `ConnectionLost` propagation through the new except arm, hinted-skip (no `0x32` request issued at all). Cold-mode wire sequence unchanged — detect characterisation goldens untouched. Full suite 1483 passed; tox matrix green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved low-voltage battery detection so missing or unresponsive primary-pack checks are handled more gracefully.
  * Detection now continues normally for supported plant types when the primary battery address does not respond.
  * Connection loss during battery detection is now surfaced correctly instead of being treated as a simple timeout.
  * When prior information shows no primary battery, detection skips that check entirely.

### DIFF
--- a/givenergy_modbus/client/client.py
+++ b/givenergy_modbus/client/client.py
@@ -901,24 +901,43 @@ class Client:
         non-contiguous and a transient BMS timeout on pack N must not drop pack N+1 onward.
         is_valid() gating and caps population are handled by _derive_capabilities.
         """
-        await self.send_request_and_await_response(
-            ReadInputRegistersRequest(base_register=60, register_count=60, device_address=0x32),
-            timeout=timeout,
-            retries=retries,
-        )
-        # #352/#289: since #352 a caps-absent 0x32 read routes through the battery getter, so its first
-        # preamble frame is held by the cold-start splice guard (the cache stays empty pending a
-        # corroborating re-read) exactly like 0x33+. detect() reads each address once, so without this
-        # confirming read the primary pack is dropped at the _derive_capabilities gate below and
-        # refresh() never re-polls it. One healthy re-read corroborates and commits; a flapping/spliced
-        # bank fails to corroborate and correctly stays out (#289 anti-poison intact).
-        if not self.plant.register_caches.get(0x32):
-            await self.send_request_and_await_response(
-                ReadInputRegistersRequest(base_register=60, register_count=60, device_address=0x32),
-                timeout=timeout,
-                retries=retries,
-            )
         batt_candidates = prior.lv_battery_addresses if prior is not None else _COLD_LV_BATTERY_RANGE
+        # The 0x32 preamble keeps the patient known-tier treatment (the primary pack matters — the
+        # #350 family), but silence is NOT fatal (#358): gateways keep battery data register-embedded
+        # in the rollup, and a hybrid with no pack attached is a valid AC/PV-only plant. Skipped
+        # entirely when prior capabilities carry no 0x32 (a hinted gateway reconnect would otherwise
+        # stall the full timeout on a read that can never answer, every reconnect).
+        if 0x32 in batt_candidates:
+            try:
+                await self.send_request_and_await_response(
+                    ReadInputRegistersRequest(base_register=60, register_count=60, device_address=0x32),
+                    timeout=timeout,
+                    retries=retries,
+                    warn_timeout=False,  # expected-absence probe semantics: don't pollute retry counters
+                )
+                # #352/#289: since #352 a caps-absent 0x32 read routes through the battery getter, so
+                # its first preamble frame is held by the cold-start splice guard (the cache stays empty
+                # pending a corroborating re-read) exactly like 0x33+. detect() reads each address once,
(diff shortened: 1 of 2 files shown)


dewet22-codex:

dewet22-codex (approved):
I have reviewed this against #358 and have no blocking concerns. The 0x32 preamble now only runs when 0x32 is in the candidate set, plain TimeoutError marks the pack absent and evicts the cache, and ConnectionLost still propagates before the TimeoutError arm. I also checked the cold-start hold boundary behind the CodeRabbit extra-test suggestion; the existing branch handles the second-read timeout via the same cache-evicting absence path, so I am not treating the missing extra regression as blocking. Local checks passed: uv run --group test pytest tests/client/test_detect.py (71 passed), uv run --group test pytest tests/client/test_detect.py tests/model/test_plant.py (230 passed), and uv run --group test pytest (1483 passed, 4 existing deprecation warnings). GitHub checks were green on the reviewed head.

PR #361, opened 2026-07-03
fix: default select_gateway to the V1 word order; first Gateway fixtures (#360)

### LINKED ISSUE #360: select_gateway() mis-selects the V2 word order on GAAA0014 — energy totals decode 10⁴–10⁵ too large
First live-data check of the Gateway variant selector, from AberDino's daylight capture (givenergy-hass#95): the energy `*_total` fields decode to impossible values (e.g. `e_grid_import_total` = 403,518,259.3 kWh) because `select_gateway()` picks the V2 (swapped-word) register layout on a unit whose true layout is V1.

The selector reads raw `IR(1603)` and compares it against 10 ("GA000009 and earlier < 10"). On this unit `IR(1600–1603)` = `0x4741 0x4141 0x0000 0x0104` (rendering as firmware `GAAA0014`), so the raw test sees 260 → V2. But the capture is decisive that the unit uses **V1 order**: assembling first-register-as-high yields sane, internally-consistent lifetime totals across all ten uint32 pairs (grid import 12,710.8 kWh, PV 5,087.0, export 2,192.7, load 12,006.4, battery charge/discharge 8,855.9/8,998.8, per-AIO 7,871.8/984.1), while the V2 order reproduces the observed garbage byte-for-byte. Notably this unit falsifies the version-split claim as stated, not just the raw-value comparison: `GAAA0014` is "version 14" under any reading, and it is V1-ordered. The V1/V2 split also moves the per-AIO serial register bases, so a mis-selection shifts those decodes too.

Both layout tables predate any live gateway traffic (phase 1.5), and GivTCP's decode of the same unit shows the identical inflated totals — the swap assumption appears to share a common, never-live-verified lineage.

Proposed fix: `select_gateway()` returns `GatewayV1` unconditionally. Every live observation to date is V1-ordered; `GatewayV2` stays in the codebase for explicit use, pending the first real sighting — which would announce itself as a unit whose totals decode insanely *under V1*. If anyone reading this has a Gateway on a `GA0000xx`-series firmware, a short capture would settle whether the swapped layout exists in the wild at all.

The daylight capture also rides into the fixture corpus with this fix (first committed Gateway fixture, solar registers non-zero) — it validates the rest of the GatewayV1 map nicely: today-counters are exact (per-AIO charge 9.8+9.3 = battery 19.1), and the AC-side vs battery-DC energy block semantics show conversion losses in the correct direction both ways.

dewet22:
Closes #360. `select_gateway()` now defaults to the V1 word order, and the repo gains its first committed Gateway fixtures.

## Problem
The first live gateway (AberDino's, via givenergy-hass#95) falsified the V1/V2 selector on both levels: firmware `GAAA0014` carries raw `IR(1603)=0x0104` (260, so the `>= 10` raw-value test mis-fired into the swapped-word V2 layout), and it is "version 14" by any reading yet is V1-ordered on the wire. Result: every uint32 energy total decoded 10⁴–10⁵ too large (`e_grid_import_total` = 403,518,259.3 kWh), and the per-AIO serial bases shifted. Both layout tables predate any live traffic; GivTCP's decode of the same unit shows identical inflation, so the swap assumption shares a common never-live-verified lineage.

## Fix
`select_gateway()` returns `GatewayV1` unconditionally — every live observation to date is V1-ordered. `GatewayV2` stays in the codebase for explicit use; a genuine V2 unit would announce itself as a gateway whose totals decode absurdly large *under V1*, and selection can be reinstated on that evidence (the docstring records the falsification for whoever finds one).

## Fixtures
`gateway_2aio_a/` — daylight (solar non-zero, three clients sharing the bus) and deep-night (GivTCP-only: the cleanest single-client poll-cycle record; `p_pv=0` guards legitimately-zero decode paths). Redaction verified — only `GW2412G000`/`WK2412G000` placeholder forms present. Golden-master tests pin `Model.GATEWAY` classification, V1 selection, the V1-order totals on real values (the regression trap: V2 order inflates them ~10⁴×), and the per-AIO-sum == battery-total map consistency on both captures.

## Verification
Full suite 1486 passed; tox matrix green. The V1-order decode is internally consistent across all ten uint32 pairs and both captures, and matches the site's plausible history (12,710.8 kWh lifetime import, 5,087.0 PV, 8,855.9/8,998.8 battery).


dewet22-codex:

dewet22-codex (changes requested):
I found a blocker while decoding the new fixture logs. The envelope serials are redacted, but the gateway payload contains a second AIO serial at IR1846-1850 with non-zero unit digits. The current GatewayV1 and GatewayV2 serial LUTs do not cover that live slot, so FrameRedactor missed it and select_gateway currently decodes shifted AIO serial fields from the fixture. This needs fixing before merge because the committed capture would publish an unredacted device identifier and the new golden tests do not pin the live AIO serial layout. I would update the Gateway serial map and redaction coverage from the capture, rewrite the fixture bytes with the unit digits zeroed, and add fixture assertions for the redacted AIO serials. Local focused check otherwise passed: uv run --group test pytest tests/model/test_gateway.py tests/model/test_fixture_golden_master.py (27 passed).

dewet22-claude:
Excellent catch — fixed in ee7f413 (amended + force-pushed so the branch tip carries no unredacted serial).

The live layout is a contiguous 5-register stride from IR(1841): aio1 1841-1845, aio2 1846-1850 — neither pre-live serial table had the aio2 slot, which is exactly why `FrameRedactor` (LUT-derived groups) missed it. Changes:

- `_GATEWAY_V1_SERIALS` corrected to the live stride (aio3 at 1851-1855 by extension); redaction coverage follows automatically, and the `test_serial_groups_pinned_floor` inventory is updated to the live addresses.
- Both fixtures re-redacted (`CH2542G066` → `CH2542G000`): 54 frames patched via the library's own encoder so CRCs stay real (all affected frames originally carried valid CRCs — the wire=0x0000 frames in the captures never contained the exposed serial).
- Golden tests now pin the live serial layout on the daylight capture (`aio1 == "CH2414G000"`, `aio2 == "CH2542G000"`, aio3 empty) — a wrong layout can no longer pass silently.

Full suite 1486 passed, tox matrix green.

PR #364, opened 2026-07-07
fix: coalesce identical splice-reject WARNINGs during sustained bursts (#355)

### LINKED ISSUE #355: Coalesce repeated splice-rejection WARNINGs during sustained sub-bus noise bursts
### Field observation (motivation)

On 2026-07-01 ~10:46–10:48 a ~2¼-minute correlated sub-bus noise burst hit a HYBRID_GEN1 (LV, two packs). The guards handled it correctly:

- **`0x32`** (primary) — ~13× `sub-bus splice (>=2 physics-impossible deltas); keeping last-good`, trips `IR76 229→56`, `IR78 225→0`, `IR81 245→0` (cell/mosfet temps collapsing toward zero — the temp-zero corruption cohort). Last-good held throughout; the mosfet-temp sensor stayed at 24.5 °C, never showed 0 °C, and recovered to 24.7 °C ~17 s after the burst ended. No poison adopted, and the entity never went `unavailable`.
- **`0x33`** (secondary) — scalar-immutable hold (#286): `IR98` bms_fw `3009→2816`, holding last-good pending 900 s sustained agreement.
- **`0x11`** — CRC guard (#255) skipped a frame (`wire=0xf4c6 computed=0x5398`), plus retries/timeouts.

The simultaneity across `0x11`/`0x32`/`0x33` points to a transport-layer noise burst, not a battery fault — the valid-CRC windowed-splice case the guard exists for. The guard did its job; this is a "working as intended" capture.

### The ergonomics problem

Each rejected poll logs one WARNING, so a sustained burst emits many identical lines (~13 here over 2¼ min; an hour-long burst would be ~1,400), which buries genuinely-actionable logs. The raw count is already captured programmatically by `splice_reject_count` (#284), so the per-line WARNINGs aren't the only record of how many rejections occurred.

### Proposed strategy (coalesce, don't drop)

- **Onset** — first rejection of a new burst (or a changed trip signature): WARNING with full trip detail, as today.
- **Repeats** — same device + same trip signature: suppress at WARNING (optionally DEBUG); `splice_reject_count` already tallies them, so nothing quantitative is lost.
- **Resolution** — when a commit finally succeeds (or the signature changes): one WARNING, e.g. `0x32 splice burst cleared after N rejections over Ms — last-good held throughout`.
- **Escalation safety** — if a burst persists past a threshold (~5 min) without clearing, re-emit a WARNING so a genuinely-stuck/persistent condition doesn't go silent.

This keeps onset + resolution + count visible, drops the repetitive middle, and still surfaces a burst that never recovers. Design latitude on the exact mechanism (per-device last-signature + count + first-timestamp tracking, threshold values). No urgency — the current behaviour is correct, just noisy.

dewet22:
Closes #355. Sustained same-signature reject bursts now log onset (full detail, unchanged wording) → DEBUG repeats (`splice_reject_count` keeps the tally) → a 300 s escalation re-warn with the running count → one clearing WARNING with stats when a commit finally succeeds (singletons skip it).

The signature is the **set of tripped (register, rule) pairs** — value-agnostic, so jittering corrupt values can't defeat the coalescing. Both hard-reject sites (serial-block change, ≥2-physics) route through it; the scalar-immutable and escrow paths were already INFO-level holds and are untouched.

On the 2026-07-01 field burst this turns ~13 identical WARNINGs into 1 (plus a clearing line); an hour-long burst goes from ~1,400 lines to onset + ~11 escalations + resolution.

Tests: 6 new (onset/repeat, clear-with-stats, singleton-no-clear, signature-change re-warns, escalation, per-device isolation). Two sustained-burst regression tests updated — their substantive protections are unchanged; they now assert the counter tally + the coalesced onset+escalation pattern (timing-deterministic: exactly 2 WARNINGs each).

Full suite 1492 passed; tox matrix green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced repeated warning spam during sustained battery splice-reject bursts.
  * Repeated identical reject patterns are now logged more quietly, with re-warnings only after a configurable interval, when the pattern changes, or when the burst escalates.
  * Recovery now emits a single consolidated “splice burst cleared” summary including rejection counts.
  * Logging coalescing is applied per device, and short reads no longer clear an active burst.
* **Tests**
  * Updated and extended burst scenario coverage to verify warning coalescing, escalation, and clearing behaviour.

### DIFF
--- a/givenergy_modbus/model/plant.py
+++ b/givenergy_modbus/model/plant.py
@@ -1163,11 +1163,16 @@ class Plant(GivEnergyBaseModel):
         if getter_cls is BatteryRegisterGetter:
             if not self._splice_guard(device_address, incoming, register_count, now=received_at):
                 return False
-            # A committing battery bank closes out any coalesced reject burst (#355).
-            commit_ts = received_at if received_at is not None else datetime.now(UTC)
-            if commit_ts.tzinfo is None:
-                commit_ts = commit_ts.replace(tzinfo=UTC)
-            self._splice_burst_note_commit(device_address, commit_ts)
+            # A committing FULL battery bank closes out any coalesced reject burst (#355).
+            # Short reads (register_count < 60) pass the guard unevaluated — mirroring the
+            # guard's own gate — so a count=1 fan-out commit between corrupt full banks
+            # must not pop the burst (it would log a misleading 'cleared' and re-spam
+            # fresh onsets).
+            if register_count >= 60:
+                commit_ts = received_at if received_at is not None else datetime.now(UTC)
+                if commit_ts.tzinfo is None:
+                    commit_ts = commit_ts.replace(tzinfo=UTC)
+                self._splice_burst_note_commit(device_address, commit_ts)
         self.register_caches[device_address].update(incoming)
         return True
 
--- a/givenergy_modbus/model/plant.py
+++ b/givenergy_modbus/model/plant.py
@@ -1336,12 +1336,20 @@ class Plant(GivEnergyBaseModel):
             self._splice_reject_streak.pop(device_address, None)
             self._bump(self.splice_reject_count, device_address)
             warn_now, suffix = self._splice_burst_note_reject(device_address, phys, now_ts)
+            # "Please report if seen" solicits UNcharacterised shapes only (#355): the corpus's
+            # sole corruption signature is the temperature family (cell-mass IR76-79/103-104 +
+            # mosfet IR81 — the temp-zero cohort and its partial variants), characterised to
+            # death, so reporting it adds nothing. Any trip OUTSIDE the temp family is a novel
+            # shape — exactly the evidence the corpus model and the #299 heal question still want.
+            _TEMP_FAMILY = ("cell_temp_deci", "mosfet_temp_deci")
+            plea = "" if all(name in _TEMP_FAMILY for _ir, name, _o, _n in phys) else " Please report if seen."
             (_logger.warning if warn_now else _logger.debug)(
                 "Rejected battery bank for device 0x%02x — sub-bus splice (>=2 physics-impossible "
-                "deltas); keeping last-good. Trips: %s%s. Please report if seen.",
+                "deltas); keeping last-good. Trips: %s%s.%s",
                 device_address,
                 self._format_splice_trips(phys),
                 suffix,
+                plea,
             )
             return False
 
--- a/givenergy_modbus/model/plant.py
+++ b/givenergy_modbus/model/plant.py
@@ -16,6 +16,7 @@ from givenergy_modbus.model.battery_splice import (
     IMMUTABLE_SERIAL,
     SCALAR_IMMUT_HEAL_POLLS,
     SPLICE_REJECT_HEAL_POLLS,
+    SPLICE_REJECT_REWARN_SECONDS,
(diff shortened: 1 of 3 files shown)


dewet22-codex:

givenergy_modbus/model/plant.py
```diff
@@ -1152,10 +1160,14 @@ def _commit_bank(
...
-        ):
-            return False
+        if getter_cls is BatteryRegisterGetter:
+            if not self._splice_guard(device_address, incoming, register_count, now=received_at):
+                return False
+            # A committing battery bank closes out any coalesced reject burst (#355).
+            commit_ts = received_at if received_at is not None else datetime.now(UTC)
+            if commit_ts.tzinfo is None:
+                commit_ts = commit_ts.replace(tzinfo=UTC)
+            self._splice_burst_note_commit(device_address, commit_ts)
```
dewet22-codex: RAG: Red. This closes the burst after any accepted battery read, including the short-read path where _splice_guard returns before treating the frame as a full observation. A count=1 battery read between identical corrupt full banks will pop the burst and, if there were already multiple rejects, log splice burst cleared even though the very next full bank is still rejected with the same signature. It also resets the coalescing so the next corrupt poll warns as a fresh onset. I think the clearing call needs to be gated to an accepted full battery-bank observation, for example register_count >= 60, rather than every accepted battery cache update.

dewet22-codex (changes requested):
I found one blocking edge in the coalescing state.

The main sustained full-bank path behaves as intended and the focused plant suite is green, but a short battery read accepted between hard-rejected full banks currently closes the burst. In a local reproduction, two identical corrupt full banks followed by a count=1 battery read logged splice burst cleared, then the next identical corrupt full bank warned as a fresh onset. That makes the clear line misleading and can reintroduce repeated onset warnings if short reads interleave with a sustained burst. I left the detail inline; I think the clear should be tied to an accepted full battery-bank observation rather than every accepted battery cache update.

Focused check passed locally: uv run pytest tests/model/test_plant.py.

givenergy_modbus/model/plant.py
dewet22-claude: Fixed in the follow-up commit — good catch on the state edge. The clear is now gated `register_count >= 60`, mirroring the guard's own short-read gate, so only an accepted full-bank observation closes a burst. Regression test drives your exact interleave (two corrupt full banks → count=1 commit → corrupt full bank): no 'cleared' line, no fresh onset, and the eventual full-bank recovery still closes out with the complete tally.

dewet22-codex:
I have started a re-review pass on this PR, focused on the short-read burst-clear blocker from the previous review.

dewet22-codex (approved):
I have re-reviewed the latest head 88e963a against the short-read burst-clear blocker and have no remaining blocking concerns.

The clear path is now gated to accepted full battery banks, so short battery reads no longer pop the coalescing state. The new regression covers the exact interleave from my review: corrupt full bank repeats, an accepted count=1 read, another corrupt full bank, and then a healthy full-bank recovery with the full reject tally.

Focused checks passed locally: uv run pytest tests/model/test_plant.py; uv run pytest tests/model/test_plant.py::test_splice_burst_short_read_does_not_clear -q.

PR #368, opened 2026-07-08
Share the BCU-count/module-count register decode

dewet22:
## Summary

Small follow-up from investigating a planned "Slice C2" for the #293 manifest series (unifying `client.py`'s live `detect()` with `plant.py`'s offline reconstruction path) — that unification turned out to already be complete, done by the earlier #268 epic before the #293 series started. `_derive_capabilities()`/`_derive_hv_topology()` are a single shared function (not two copies) called from both the live and offline paths, and every other piece of candidate-address logic in that area (`_hv_bmu_candidates`, `_aio_module_candidates`, `_COLD_METER_RANGE`, `_COLD_LV_BATTERY_RANGE`) is already shared too.

One small leftover did surface: `_detect_bcu_stacks()` independently re-decoded the raw `IR(61)` BCU-count and `IR(64)` module-count registers in three places (its own hinted path, its own cold path, and again inside `_derive_hv_topology`) instead of going through a shared helper — the one piece of logic in this area that wasn't already DRY.

Extracted `_bms_bcu_count()`/`_bcu_module_count()` in `plant.py` and pointed all four call sites at them. Zero behaviour change — the decode is byte-identical, just written once.

## Test plan

- [x] Full suite (1534 tests) passes unmodified
- [x] mypy, ruff, tox all green

### DIFF
--- a/givenergy_modbus/client/client.py
+++ b/givenergy_modbus/client/client.py
@@ -38,6 +38,8 @@ from givenergy_modbus.model.plant import (
     Plant,
     PlantCapabilities,
     _aio_module_candidates,
+    _bcu_module_count,
+    _bms_bcu_count,
     _derive_capabilities,
     _hv_bmu_candidates,
 )
@@ -725,7 +727,7 @@ class Client:
                     retries=probe_retries,
                 ):
                     bcu_cache = self.plant.register_caches.get(0x70 + offset, RegisterCache())
-                    actual_modules = bcu_cache.get(IR(64)) or 0
+                    actual_modules = _bcu_module_count(bcu_cache)
                     caps.bcu_stacks.append((offset, actual_modules))
                 else:
                     self.plant.mark_absent(0x70 + offset, "IR", 60, 5)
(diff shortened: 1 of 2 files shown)


dewet22-codex:

dewet22-codex (approved):
Reviewed 0b79195. I don’t see any blocking concerns: the new helpers preserve the old IR(61)/IR(64) fallback semantics, and the live hinted/cold detect paths plus offline topology derivation now share that decode without changing absent-cache or zero-count behaviour.

Focused local checks passed: ============================= test session starts ==============================
platform darwin -- Python 3.11.15, pytest-9.0.3, pluggy-1.6.0
rootdir: /Users/dewet/git/givenergy-modbus
configfile: pyproject.toml
plugins: cov-7.1.0, timeout-2.4.0, asyncio-1.3.0
timeout: 1.0s
timeout method: signal
timeout func_only: False
asyncio: mode=Mode.AUTO, debug=False, asyncio_default_fixture_loop_scope=None, asyncio_default_test_loop_scope=function
collected 245 items

tests/model/test_plant.py .............................................. [ 18%]
........................................................................ [ 48%]
.....................................................                    [ 69%]
tests/client/test_detect.py ............................................ [ 87%]
..............................                                           [100%]

============================= 245 passed in 0.57s ============================== (245 passed), plus .

PR #369, opened 2026-07-08
Manifest Slice D: write surface (#293)

dewet22:
## Summary

Fourth and **final** manifest slice (A semantics-routing+renames [#365] → B fact-table consolidation [#366] → C1 polling-range manifest [#367] → C2 [investigated, already unified by #268; leftover in #368] → **D write surface**). With this merged, `manifest.py` owns every model/capability-keyed decision across the read AND write surface, and **2.10.0 can cut.**

This is the highest-risk surface in the codebase (register writes to real hardware), so it carries an extra verification layer — see the test plan.

Relocates the four write-safe register allowlists from `commands.py` into `manifest.py` and unifies the write-gating decision:

- **New:** `manifest.WRITE_SAFE_SINGLE_PHASE` / `WRITE_SAFE_THREE_PHASE` (derived from the single-phase set) / `WRITE_SAFE_EMS` / `WRITE_SAFE_AC_CONFIG` — every register annotated with its meaning for visual inspectability — plus `write_safe_registers(model, arm_fw=None)`, one accessor encoding the base-set selection, the AC-config union gated on `has_ac_config_block AND NOT is_three_phase` (AC_3PH has the block but remaps those controls, so it's correctly excluded), and the `model=None` → single-phase conservative fallback.
- `client.py`'s `one_shot_command()` and `installer_command()` each drop their ~7-line duplicated inline dispatch for one accessor call. The `installer_command()` `| INSTALLER_WRITE_REGISTERS` union stays at the client boundary — the installer/destructive tier is a universal danger classification, deliberately NOT a per-model fact.
- `commands.py`'s three mixin `WRITE_SAFE_REGISTERS` ClassVars + module-level `_AC_CONFIG_WRITE_SAFE_REGISTERS` are deleted; every test repointed to the manifest constants.

## Zero behaviour change

Pure relocation — no model gains or loses write permission for any register. `#297` (HR311/317 folded into `has_ac_config_block`) was already landed before this slice; nothing about the gate semantics changes here.

## Test plan

- [x] A temporary bridging test proved each relocated set byte-equal to its `commands.py` original, then was removed once its job was done
- [x] Hardcoded-literal regression pins for the single-phase and three-phase sets — non-tautological fences so a future edit to the collapsed range/comprehension expressions fails loud rather than silently shifting the allowlist (the three-phase pin double-covers the derived set)
- [x] The full accept/reject behavioural matrix in `test_client_write_safety.py` (per model, per tier, incl. the AC_3PH-no-union case and the undetected→single-phase fallback) passes with no assertion weakened
- [x] **`register-safety-reviewer` PASS** — independently expanded every range comprehension against the originals, verified the AC_3PH guard, the None fallback, and the installer-tier boundary
- [x] Full suite (1541), mypy, ruff, and tox all green

## Summary by CodeRabbit

* **New Features**
  * Added unified write-safe register handling based on device model, including support for single-phase, three-phase, EMS and AC-config register sets.
  * Improved register validation so writes are checked against the appropriate model-specific allowlist.

* **Bug Fixes**
  * Refined write permissions for certain devices so the correct registers are available for supported configurations.
  * Updated coverage to better protect against allowlist drift and unexpected register writes.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -71,8 +71,10 @@ Prefer real evidence (wire captures, GivTCP cross-reference) before adding a wri
 register. A new writable register must be added in **two** places or it fails at
 send time:
 
-1. the command-mixin `WRITE_SAFE_REGISTERS` in `client/commands.py`, and
-2. the canonical `WRITE_SAFE_REGISTERS` in `pdu/write_registers.py` — enforced by
+1. the model-aware `WRITE_SAFE_*` set in `model/manifest.py` (Gate 1 — the client-boundary
+   check via `write_safe_registers()`; add to the set matching the models that physically
+   accept the register: `WRITE_SAFE_SINGLE_PHASE` / `_THREE_PHASE` / `_EMS` / `_AC_CONFIG`), and
+2. the canonical `WRITE_SAFE_REGISTERS` in `pdu/write_registers.py` (Gate 2) — enforced by
    `WriteHoldingRegisterRequest.ensure_valid_state()` at `encode()` time. A register
    missing here builds a request fine but raises `InvalidPduState` on the wire.
 
--- a/givenergy_modbus/client/client.py
+++ b/givenergy_modbus/client/client.py
@@ -1373,6 +1373,21 @@ class Client:
                     retry_delay=retry_delay,
                 )
 
+    def _resolve_write_safe(self) -> frozenset[int]:
+        """Registers the detected model permits at the normal (non-installer) write tier.
+
+        The single source of the caps→manifest resolution shared by one_shot_command()
+        and installer_command() — write-gating logic that must stay identical across
+        both call sites. Composition (base selection, AC-config union with its
+        not-three-phase guard, undetected→single-phase fallback) lives in
+        manifest.write_safe_registers (#293 Slice D).
+        """
+        caps = self.plant.capabilities
+        return manifest.write_safe_registers(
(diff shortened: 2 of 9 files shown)


dewet22-codex:
I have started a careful review pass on this PR.

dewet22-codex (approved):
Reviewed 13e7c29 with extra attention on the write-gating invariants. I do not see any blocking concerns: the relocated manifest sets match the pre-PR command-layer sets, the AC-config union still lands only on AC and All-in-One, AC_3PH stays excluded from HR311/313/314/317, undetected clients keep the single-phase fallback, and installer-tier writes remain unioned only at Client.installer_command.

Local verification passed: imported base-vs-PR allowlists for direct comparison, ran an explicit accept/reject matrix for AC/AIO/DC-hybrid/AC_3PH/undetected/EMS cases, ran uv run --group test pytest tests/model/test_manifest.py tests/client/test_client_write_safety.py tests/client/test_inverter_commands.py tests/client/test_commands.py (366 passed), ran the full suite with uv run --group test pytest (1541 passed), plus git diff --check.

One non-blocking housekeeping note: AGENTS.md still tells future write-surface changes to edit the old command-mixin WRITE_SAFE_REGISTERS in client/commands.py. Since this PR deliberately moves that model-aware owner to manifest.WRITE_SAFE_*, I would align that guidance soon so the next agent lands in the right place.

Issue #370, opened 2026-07-08
Recapture a complete three_phase_hv_a fixture (with HR banks) for live full-cycle integration coverage


dewet22:
## Context

The live full-cycle integration test added alongside the #293 manifest series (`tests/client/test_mock_plant_integration.py::test_full_cycle_instantiates_plant`) connects a real `Client` to `MockPlant` seeded from each golden fixture and runs the complete `detect → load_config → refresh` cycle, asserting the fully-decoded typed `Plant`.

`three_phase_hv_a` had to be **excluded** from that test: its committed capture (`tests/fixtures/captures/three_phase_hv_a/…`) is **IR-only** — it contains no HR bank at device `0x11`. So `detect()`'s very first identity read, HR(0,60)@0x11, gets an error response → `TimeoutError` → detect tears down before a cycle can run. The other five topologies (EMS, hybrid, gateway, AIO) are covered live; three-phase stays covered only by the **offline** `tests/model/test_fixture_golden_master.py`, which replays response bytes and never exercises the detect/load_config/refresh routing.

## Ask

Recapture a three-phase HV plant with a capture that includes the **HR configuration banks** (at minimum HR(0,60) at 0x11 for identity, plus the HR(60,60)/HR(120,60)/HR(1000-1124) banks `load_config()` polls for three-phase), so a live full cycle can complete. Then add a `three_phase` case to the parametrized `test_full_cycle_instantiates_plant` (lifting the three-phase topology pins).

This closes the one remaining gap in live end-to-end coverage of the manifest-driven polling routing — three-phase is the only model class currently guarded offline-only.

Follows up on the #293 series (2.10.0).

PR #371, opened 2026-07-08
Live full-cycle integration test: MockPlant → detect/load_config/refresh → full Plant decode

dewet22:
## Summary

Closes the one coverage hole the #293 capability-manifest series (2.10.0) left behind. That series rewrote exactly which register banks `detect()`/`load_config()`/`refresh()` poll per model and how decoded values route into the typed `Plant` — but no single test exercised that whole path live:

- `test_mock_plant_integration.py` ran a live `detect()`+`refresh()` against the real `MockPlant` asyncio server, but asserted only capabilities/flags, never the decoded device objects, and never called `load_config()`.
- `test_fixture_golden_master.py` asserts the full decoded topology — but **offline**, replaying response bytes through `Plant.update()`, bypassing the client/detect/poll stack.
- `load_config()` was never run against the live server at all.

This adds a parametrized test that connects a real `Client` to `MockPlant` (seeded from each golden capture) and runs the complete `detect → load_config → refresh` cycle over the real socket/framer/PDU stack, then asserts the fully-decoded typed `Plant` — lifting the golden-master's per-fixture topology pins but reaching them through the live stack.

## Coverage

| Fixture | Cycle | Asserts |
|---|---|---|
| EMS (60s) | clean | 2 managed inverters via rollup, data_source, serials |
| EMS (30min) | clean | meters 0x01/0x03 pf + apparent (derived **live**, unlike the offline test which hand-sets meter addresses) |
| hybrid 0x11-poll | clean | HYBRID_GEN1, 2 LV packs, IR44 stays PV |
| gateway daylight | clean | GatewayV1, GAAA0014, lifetime totals, AIO serials |
| AIO | partial-tolerated | ALL_IN_ONE, HV BCU stack, IR44 = inverter-output (#293) |
| hybrid-passive | partial-tolerated | classification + LV-battery detection from a pre-#189 dongle capture |

`three_phase_hv_a` is excluded — its committed capture is IR-only (no HR bank at 0x11), so `detect()`'s identity read errors and a live cycle can't start. Tracked in #370 (recapture a complete three-phase fixture); three-phase stays covered offline meanwhile.

Partial captures raise `RefreshPartiallySucceeded`, which carries a `.plant` decoded from the banks that responded — the test tolerates it and asserts on that plant.

## Teeth (verified once, not committed)

- **Mutation:** dropping the gateway `REFRESH_IR_RANGES` entry turns the gateway case RED — a *polling-routing* regression the offline golden-master structurally cannot catch (it hand-feeds bytes and never polls). Removing `ALL_IN_ONE` from the IR44 identity set turns the AIO case RED. Both reverted.
- **Before/after (v2.9.8 vs v2.10.0):** the gateway decode is byte-identical across the series (the refactor was behaviour-preserving on the polling/decode path), while the AC/AIO PV value moved from `e_pv_generation_today` to `e_inverter_out_today` exactly as Slice A intended.

## Test plan

- [x] 6 new cases green; full suite (1547), mypy, ruff, tox all green
- [x] Test-only — no production code touched

## Summary by CodeRabbit

* **Tests**
  * Added broader end-to-end coverage for client connections using live fixture data.
  * Verified that multiple device topologies are instantiated and decoded correctly after a full detection and refresh cycle.
  * Improved resilience in test coverage for cases where a refresh completes only partially.

### DIFF
--- a/tests/client/test_mock_plant_integration.py
+++ b/tests/client/test_mock_plant_integration.py
@@ -136,10 +136,13 @@ async def _full_cycle(client):
     RefreshPartiallySucceeded — carrying a Plant fully decoded from the banks that DID
     respond. Clean captures raise nothing; these fixtures never RefreshFailed.
     """
-    await client.detect(**_DETECT)
+    # Fast, in-process timeouts throughout: MockPlant answers present banks instantly,
+    # so these only bound waits on genuinely-absent device addresses (detect's meter/
+    # peripheral probe sweeps) and absent banks. No network to be slow — don't raise them.
+    await client.detect(timeout=0.3, retries=0, probe_timeout=0.05, probe_retries=0)
     for step in (client.load_config, client.refresh):
         try:
-            await step(timeout=1.0, retries=0)
+            await step(timeout=0.3, retries=0)
         except RefreshPartiallySucceeded:
             pass
     return client.plant
--- a/tests/client/test_mock_plant_integration.py
+++ b/tests/client/test_mock_plant_integration.py
@@ -13,6 +13,7 @@ from pathlib import Path
 import pytest
 
 from givenergy_modbus.client.client import Client
+from givenergy_modbus.exceptions import RefreshPartiallySucceeded
 from givenergy_modbus.model.inverter import Model
 from givenergy_modbus.model.register import HR
 from givenergy_modbus.model.register_cache import RegisterCache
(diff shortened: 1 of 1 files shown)


dewet22-codex:

tests/client/test_mock_plant_integration.py
```diff
@@ -113,6 +114,145 @@ async def test_aio_errors_on_absent_three_phase_bank():
...
+    Some committed captures don't carry every HR/IR bank the cycle polls; MockPlant
+    error-responds to an absent bank, which the client surfaces as
+    RefreshPartiallySucceeded — carrying a Plant fully decoded from the banks that DID
+    respond. Clean captures raise nothing; these fixtures never RefreshFailed.
+    """
+    await client.detect(**_DETECT)
+    for step in (client.load_config, client.refresh):
+        try:
+            await step(timeout=1.0, retries=0)
+        except RefreshPartiallySucceeded:
```
dewet22-codex, line 146: I think this weakens the new regression fence enough to fix before merge. Because every RefreshPartiallySucceeded is swallowed and the final assertions only depend on detect plus refresh data, the new parametrized test still passes if load_config is skipped entirely across all six fixtures. That leaves the specific #369 follow-up gap partly open: a future load_config routing change that turns a clean fixture partial, or even a no-op load_config, can still go green as long as refresh/detect populate the asserted topology. I would make the tolerance explicit per case, for example by returning the partial failures and asserting the clean fixtures have none while aio/hybrid-passive have only the known missing banks, or by pinning at least one value/request that requires the load_config bank for the clean cases.

dewet22-codex (changes requested):
I have left one blocking comment. The new test does exercise the socket path and the focused test passes, but the current assertions still pass if load_config is skipped entirely, so I think the #369 coverage gap is not quite closed yet.

Local checks run: uv run --group test pytest tests/client/test_mock_plant_integration.py -k full_cycle -vv (6 passed), plus git diff --check. I also ran a detect-plus-refresh-only probe and all six new cases still passed, which is the basis for the inline comment.

tests/client/test_mock_plant_integration.py
dewet22: Good catch — confirmed and fixed in 63e1997 (landed just after this review). I verified the gap exactly as you describe: skipping load_config entirely left all six cases green. Closed it per your last suggestion, by pinning fields decoded *only* by load_config:

- hybrid + aio → the inverter's HR(60,60) config (`charge_target_soc`, `battery_charge_limit`, `enable_charge`) — `None` without load_config.
- EMS → the HR(2040,36) config (`export_power_limit == 9500`, `plant_enabled`) — the `is_ems` entry in the `LOAD_CONFIG_RANGES` manifest table.

Verified with teeth: skipping load_config now reddens hybrid+aio, and dropping the `is_ems` `LOAD_CONFIG_RANGES` entry reddens the EMS case — so both the inline HR reads and the table-driven routing are fenced.

One residual, fixture-data-limited: the `has_ac_config_block → HR(300,60)` entry still can't be pinned because no committed capture carries that bank (the AIO's HR300 read comes back absent). Folded into the recapture follow-up #370.

dewet22-codex:
I have started a re-review pass on this PR.

dewet22-codex (approved):
Reviewed 63e1997. I do not see any blocking concerns remaining: the new load_config-exclusive assertions now make the formerly-unverified leg load-bearing. Re-running the old detect-plus-refresh-only probe now fails the EMS rollup, hybrid 0x11, and AIO cases, which closes the blind spot from my earlier review.

Local checks passed: uv run --group test pytest tests/client/test_mock_plant_integration.py -k full_cycle -vv (6 passed), uv run --group test pytest tests/client/test_mock_plant_integration.py (29 passed), plus git diff --check. CI is green on the current head.

PR #372, opened 2026-07-08
fix(gateway): report AIO power in house convention (positive = discharge)

dewet22:
## What

The Gateway firmware reports its AIO power fields **positive-while-charging** — inverted relative to the single-phase inverter's `p_battery` (which is positive = discharge) and relative to GivTCP. This adds a small composable `Converter.negate` post-converter and applies it to all five affected fields so they land in the library's house convention (**positive = discharge/export**), letting consumers drop their own negations.

Fields flipped:
- `p_liberty` (IR1619)
- `p_aio_total` (IR1702)
- `p_aio1_inverter` / `p_aio2_inverter` / `p_aio3_inverter` (IR1816-1818)

## Evidence (fixture-confirmed, not just external comparison)

Decoded `gateway_2aio_a`:

| Field (raw) | Daylight (PV=1997W) | Night (PV=0) |
|---|---|---|
| `p_aio_total` | +154 | −797 |
| `p_liberty` | +212 | −786 |
| `p_aio1/2_inverter` | +76 / +85 | −386 / −411 |

The **night capture is unambiguous**: PV=0, grid ≈ −31W, load=817W → the house runs off the batteries (discharging ~797W), raw reads **−797**. Daylight: PV surplus into near-full batteries (94/96% SOC) → charging, raw reads **+154**. So raw is positive-while-charging; negating gives positive = discharge, consistent with `p_battery`.

## Scope note (2 → 5)

hass asked for `p_liberty` + `p_aio_total`. The fixtures show the per-AIO components carry the **same** convention and sum exactly to the total (night: −386 + −411 = −797), so flipping the total without them would ship an internally inconsistent model. All five flip together. `p_ac1` (grid) is a separate convention and is left untouched.

`p_liberty`'s exact semantics remain unconfirmed, but it tracks `p_aio_total`'s sign across both captures, so the same treatment applies.

## Tests

- `Converter.negate` unit test (sign flip + None passthrough) and precision.
- `test_gateway_power_readings` updated for the flipped expectations.
- `test_fixture_golden_master` gateway day/night cases now assert the sign convention against real captures (charge → negative, discharge → positive, per-AIO sum == total). Mutation-checked: dropping either `negate` turns both fixture cases red.
- Full suite + tox (py311-314, lint, format, build) green.

Coordination context: dewet22/givenergy-hass#95.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected sign handling for several gateway power readings, including Liberty, total AIO power, and per-inverter AIO values.
  * Improved displayed power values so they now follow the expected convention across different fixture states.

* **Tests**
  * Expanded coverage for gateway power readings and fixture checks.
  * Added unit tests for sign-flipping behaviour and numeric precision handling.

### DIFF
--- a/givenergy_modbus/model/gateway.py
+++ b/givenergy_modbus/model/gateway.py
@@ -74,7 +74,13 @@ _GATEWAY_COMMON_LUT = {
     "p_ac1": Def(C.int16, None, IR(1616)),
     "p_pv": Def(C.uint16, None, IR(1617), max=50000),
     "p_load": Def(C.uint16, None, IR(1618), max=50000),
-    "p_liberty": Def(C.int16, None, IR(1619)),
+    # p_liberty / p_aio_total / p_aio{1,2,3}_inverter: firmware reports positive
+    # while charging; C.negate brings them into the library's house convention
+    # (positive = discharge), matching the single-phase inverter's p_battery.
+    # Fixture-confirmed on gateway_2aio_a (night: battery discharge → raw −797;
+    # daylight: charge → raw +154). p_liberty's exact semantics are still open,
+    # but it tracks p_aio_total's sign, so the same treatment applies (#194/hass).
+    "p_liberty": Def(C.int16, C.negate, IR(1619)),
     "fault_protection": Def(C.uint32, None, IR(1620), IR(1621)),
     "gateway_fault_codes": Def(C.uint32, _gateway_fault_code, IR(1622), IR(1623)),
     "v_grid_relay": Def(C.deci, None, IR(1624), min=0.0, max=500.0),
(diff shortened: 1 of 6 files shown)


dewet22-codex:

dewet22-codex (approved):
Reviewed c3adbb3 and I do not see blocking concerns. The change keeps the Gateway fields on the existing int16 path and adds only a post-conversion sign flip, while leaving p_ac1 untouched. The fixture coverage now pins both daylight charging and night discharge behaviour from the live Gateway captures, and the audit scale handling treats negate as magnitude-neutral.

Local checks: uv run --group test pytest tests/model/test_gateway.py tests/model/test_register.py tests/model/test_fixture_golden_master.py tests/scripts/test_audit_app_reconciliation.py (90 passed). GitHub checks are green.

PR #376, opened 2026-07-08
fix(redactor): redact HV BMU per-module serials in captures (#375)

### LINKED ISSUE #375: FrameRedactor leaks HV BMU per-module serials (0x50+ stacks)
## Summary

`FrameRedactor` (and by extension `RegisterCache.redact_serials`) does **not** redact the per-module HV BMU serial numbers on multi-module HV stacks addressed at 0x50–0x55. The envelope/inverter serials are handled, but the trailing per-module BMU serial (≈IR114–118 within each 0x50-device FC04 IR(60,60) block) passes through intact.

## Evidence

Verified against a real three-phase HV capture (GIV-3HY-11, 6 HV modules) by feeding every rx/tx frame through `FrameRedactor` and extracting distinct serials before vs after:

```
distinct HV serials BEFORE redaction: HY####G6xx  (6 distinct, real)
distinct HV serials AFTER  redaction: HY####G6xx  (identical — unchanged)
```

The 6 real module serials survive redaction byte-for-byte.

## Why it matters

- Blocks committing any HV-stack wire capture as a golden fixture — would publish real serials, violating the share-safe-export guarantee (#212/#214).
- Same class of hazard as #361: a register layout whose serials aren't reached by the redactor's serial-group mapping leaks silently, with no error.

## Likely cause

`redact_serials` applies `Converter.serial`-tagged groups + "manually decoded" BMU groups, but those BMU groups don't appear to be applied to the 0x50–0x55 HV-stack device frames (they may only be mapped onto 0x11 / battery-device contexts). Needs root-cause confirmation.

## Acceptance

- The 6 module serials in the reference capture redact to a single zeroed placeholder (leak count 6 → 1), matching the `AAYYWWA000` scheme.
- Regression test pins the before/after leak count.
- Prerequisite for landing the three-phase HV golden fixture (#370-adjacent).

dewet22:
Closes #375.

## Problem

`FrameRedactor` (capture path) kept its **own** serial-register-group builder, `_build_serial_register_groups()`, which had diverged from the canonical `RegisterCache._get_serial_groups()` (export path). The copy omitted:

- the manually-decoded **BMU serial groups** (IR114+ per module — no LUT `Def` carries them),
- `identifier=True` fields (e.g. meter product identifier),
- dynamic model-module discovery (hardcoded `inverter, battery, ems, gateway` only).

So HV per-module serials on 0x50–0x55 stacks passed through shared captures **unredacted** — a silent share-safe-export leak (#212/#214), same class of hazard as #361 (a layout the redactor's mapping doesn't reach leaks with no error).

## Fix

1. Point `FrameRedactor` at the single canonical `_get_serial_groups()` and delete the divergent copy — one source of truth.
2. Mirror `redact_serials`' **fail-open guard**: skip a group when `redact_serial` returns `None` or an unchanged passthrough. Without it, unifying introduced a new corruption — the now-included overlapping battery serial group IR(110–114) would re-encode register 114 of a BMU serial living at 114–118, zero-padding and clobbering it. The guard matches the export path exactly.

## Verification

- **Real three-phase HV capture (6 HV modules):** distinct HV serials **6 → 0** after redaction.
- New regression test `test_frame_redactor_redacts_bmu_module_serial` pins a BMU-serial frame (IR114–118) redacting to the zeroed `HY####G000` placeholder; red before the fix, green after.
- Full suite (1550) + tox (py311–314, lint, format, build) green.

Prerequisite for landing the three-phase HV golden fixture (#370-adjacent).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved redaction of captured frames so serial-bearing values are handled consistently and less likely to leak in logs or captures.
  * Prevented unintended overwriting of nearby register values during redaction, helping preserve frame integrity.
* **Tests**
  * Added coverage for redaction of BMU module serial data in captured register frames.

### DIFF
--- a/givenergy_modbus/client/client.py
+++ b/givenergy_modbus/client/client.py
@@ -59,64 +59,12 @@ _GE_SERIAL_STR_PATTERN = re.compile(r"^[A-Z]{2}\d{4}[A-Z]\d{3}$")
 Direction = Literal["rx", "tx"]
 
 
-# ---------------------------------------------------------------------------
-# Serial-register lookup — which register addresses carry C.serial values
-# ---------------------------------------------------------------------------
-#
-# Built once at import time by walking every registered RegisterGetter LUT and
-# collecting all Defs whose pre_conv is Converter.serial.  The result maps
-# (register_type_name, address) → field_name. Register addresses are globally
-# unique per type across all models (HR_13 is always the inverter serial, etc.),
-# so no device-address filter is needed.
-#
-# Populated set (verified at B-3 implementation time):
-#   HR: 8-12 (battery serial), 13-17 (inverter serial)
-#   IR: 110-114 (battery), 1627-1631 (gateway first_inverter),
-#       1831-1835/1838-1842/1845-1849 (gateway v1 AIO 1-3),
-#       1841-1845/1848-1852/1855-1859 (gateway v2 AIO 1-3),
-#       2066-2085 (EMS rollup ×4)
-def _build_serial_register_groups() -> "list[tuple[str, int, int]]":
-    """Build a list of (reg_type, base_address, count) for every C.serial register group.
-
-    A serial field spans `count` consecutive registers starting at `base_address`.
-    Addresses are globally unique per type across all models (HR_13 is always the
-    inverter serial, etc.), so no device-address filter is needed.
-    """
-    from givenergy_modbus.model import battery, ems, gateway, inverter
-    from givenergy_modbus.model.register import Converter
-
-    seen: set[tuple[str, int, int]] = set()
-    groups: list[tuple[str, int, int]] = []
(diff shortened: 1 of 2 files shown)


dewet22-codex:
I have started a review pass.

dewet22-codex (approved):
Reviewed f64deca and I do not see blocking concerns. The capture redactor now uses the shared RegisterCache serial-group map, including the explicit BMU module groups, and keeps the same fail-open handling for unrecognised overlapping windows. That covers the issue 375 leak without changing unrelated frame handling.

Local checks passed:
- uv run --group test pytest tests/client/test_capture.py tests/model/test_register_cache.py -k "redact or serial" (29 passed)
- git diff --check 38ce07878f56e87c59decec9cd3d1168932e754e..HEAD

CI is green as well.

PR #377, opened 2026-07-08
feat(fixtures): land multi-module 3ph HV capture + BCU unit serial

dewet22:
Refs #265, #375, #370.

## What

Lands the second capture for `three_phase_hv_a`: a ~10 min givenergy-cli passive capture of the same GIV-3HY-11 plant (128 frames, **zero error responses**), supplied by the same reporter via the hass-side soak. It adds what the original HA capture lacked:

- **Per-module BMU reads at 0x50–0x55** — all 6 HV modules answering at their own device addresses with per-cell blocks. This is the multi-module probe #265 asked for: first fixture where per-module, per-cell voltages (~3.35 V LiFePO4) and temps decode against real silicon.
- The BCU's second `IR(120,60)` block, BAMS devices at 0x90/0xA0, and both meters.

Golden-master tests pin per-module per-cell decode, BCU cluster fields (6 modules, 52 Ah nominal — matching the reporter's stated hardware), and the 3ph inverter/meter blocks.

## Model change: BCU unit serial at IR(138–142)

The count-to-zero redaction sweep found a GE serial (`HB…`, same manufacture batch as the modules) on the BCU device at IR(138–142) — first wire evidence, previously unmodelled and therefore **unredacted** (the #375 leak class, one register block over). Modelled as a `C.serial` Def on `BcuRegisterGetter`, which both exposes `Bcu.serial_number` and auto-registers the group in the canonical serial groups so capture/export redaction covers it. FrameRedactor regression test included.

## Redaction provenance

Redacted through the library's `FrameRedactor` (date scheme — prefix + manufacture date kept, unit digits zeroed, CRCs re-encoded), then verified count-to-zero (full-serial scan AND the stricter prefixless tail scan: no non-placeholder serials remain).

Two genuine hardware quirks on module 0x55, documented in the README and follow-up issues:
- Its serial is stored **split** on the wire (`HY` at IR110, tail at IR115–118) — structurally invisible to fixed-group redaction, so its unit digits were hand-zeroed through the library encoder; the wire layout itself is preserved as captured.
- It populates only 12 of 24 temp sensors, and its serial prefix decodes as `t_cell_21 = 1852.1` (the imperative decode applies no bounds). Pinned as characterisation; the bounds gap is filed as a follow-up.

## Limitations

Neither capture in this fixture carries `HR(0,60)` at 0x11, so it cannot drive a live `detect()` cycle — three-phase live-full-cycle coverage remains #370 (a detect-inclusive capture ask is with the reporter).

## Verification

- New tests: device topology, 6× per-module per-cell decode, BCU cluster + serial, inverter + meters — 7 passing.
- Full suite (1555) + tox (py311–314, lint, format, build) green.


dewet22-codex:
I have started a review pass.

dewet22-codex (approved):
Reviewed 2c8f81f and I do not see blocking concerns. The BCU IR(138-142) serial is registered via C.serial, so the canonical serial-group scan covers capture/export redaction, and the new fixture pins the separate-address HV BMU decode from 0x50-0x55 against real frames. I also checked the split 0x55 serial path against Plant.update with HV BMU capabilities; it still commits rather than dropping the bank.

Local checks passed:
- uv run --group test pytest tests/client/test_capture.py tests/model/test_three_phase_from_capture.py
- uv run --group test pytest tests/model/test_hv_bcu.py tests/model/test_register_cache.py tests/client/test_load_refresh.py -k "bcu or bmu or serial or hv_bmu or redact"
- uv run --group test pytest tests/client/test_mock_plant_integration.py tests/model/test_plant.py -k "hv_bmu or full_cycle or distinct_module_serials or serials"
- uv run --group test pytest tests/model/test_fixture_golden_master.py tests/client/test_detect_characterisation.py tests/client/test_offline_from_caches.py -k "three_phase or hv or capture"
- git diff --check 2a72a45c8d2d3d778a1135dae66d5a5903030250..HEAD

One small provenance caveat: the capture description says 128 frames, while the committed file is 128 lines and the fixture loader parses 118 complete RX frames (119 rx lines, 3 tx lines, 6 comments). That looks documentation-only and does not affect the fixture coverage or redaction result.

Issue #318, opened 2026-06-26
EMS remaining-battery-energy appears to under-count multi-battery managed inverters


dewet22:
Investigating a discrepancy reported downstream in dewet22/givenergy-hass#221, from the EMS tester whose plant is our `ems_2_inv_3_bat_a` golden fixture. The numbers point at the EMS controller's remaining-energy figure not aggregating all packs on a multi-battery managed inverter — but I'd like the library side to confirm against the fixture's actual registers before I tell the user it's a real bug.

## Plant (from `tests/fixtures/captures/ems_2_inv_3_bat_a/README.md`)

- Inverter #1 — 2× Giv-Bat 5.2 (primary/secondary chain) → **10.4 kWh nominal**
- Inverter #2 — 1× Giv-Bat 5.12 Gen3 → **5.12 kWh nominal**
- **Total nominal: 15.52 kWh**

## Live observation (HA side)

- EMS "Remaining Battery Energy" = **7,535 Wh** at plant SOC 70% (inv #1 SOC 57%, inv #2 SOC 84%).
- That implies ~**10.8 kWh** usable — only **~70% of the 15.52 kWh nominal**, which is low for LFP.

## The tell — per-inverter isolation

From the tester's overnight history; chart-derived, so approximate:

- When inv #2 sat at ~0% SOC, total remaining ≈ 1.3 kWh at inv #1's ~24% SOC → inv #1 resolves to ~**5.4 kWh** capacity.
- Cross-check (inv #2 at 100%, inv #1 at 65%, total ≈ 8.9 kWh) → inv #2 ≈ **5.4 kWh**.
- So the two inverters resolve to *equal* capacity — yet inv #1's nameplate (10.4 kWh) is double inv #2's (5.12 kWh). The two-battery inverter is being counted as roughly **one battery's worth**.

## Hypothesis

The EMS remaining-energy register (or our decode/aggregation of it) omits the secondary pack on a primary/secondary multi-battery inverter. The per-inverter SOC looks fine; it's the energy/capacity total that's short.

## Ask

Decode `ems_2_inv_3_bat_a` and confirm at the register level —

1. The per-pack battery nominal-capacity registers (HR308–310 and equivalents): are all three packs present and correctly scaled?
2. The EMS remaining-energy register: does it represent all three packs, or does inv #1's secondary pack drop out?

[...]

PR #386, opened 2026-07-09
feat(model): add remaining battery energy fields

### LINKED ISSUE #374: Add Plant-level total-remaining-battery-energy field (EMS multi-battery undercount remedy, #318)
Split out from #318 so the agreed remedy does not get lost again.

#318 concluded the EMS remaining-energy figure (IR2091) is a **genuine firmware undercount** on primary/secondary multi-battery managed inverters — the secondary pack drops out — **not** a decode bug on our side. The agreed remedy (2026-06-26) was a **Plant-level `computed_field`** returning total remaining usable energy in **Wh**, summed over *all* packs including the dropped secondary, as a like-for-like swap for the existing EMS "Remaining Battery Energy" sensor.

- Per-pack basis: `cap_remaining` (Ah) × pack voltage (library's call on measured `v_battery` vs a nominal basis, and whether `cap_calibrated` should bound it).
- `IR2091` / the existing `Ems` field stays as-is for diagnostics and to document the quirk. The change is purely additive.
- Plant-level is the right home — the `Ems` object alone does not see the inverter sub-bus; the full `Plant` does.

Not blocking hass (they can drive the sensor off the per-pack `Battery.cap_*` sum today), but worth shipping so CLI and other consumers get it for free.

dewet22:
Closes #374.

## What

Adds a correct "remaining battery energy" figure that includes **every** pack — the remedy for the #318 EMS undercount — as decode-side model fields. No polling change; `IR2091` / `Ems.remaining_battery_wh` untouched.

- **`Battery.remaining_energy_nominal_wh`** — `cap_remaining (Ah) × num_cells × 3.2 V`. Nominal voltage is stable across load/SOC (the basis nameplate capacity is quoted on), so it's the figure to sum for a slowly-varying reading. If a pack doesn't report `num_cells`, falls back to measured `v_out` so it still contributes rather than silently dropping — the exact failure mode #318 is about.
- **`Battery.remaining_energy_measured_wh`** — `cap_remaining × v_out`, instantaneous (fluctuates ~2–3% with SOC / load).
- **`Plant.remaining_battery_energy_wh`** — sums the nominal field over the Plant's LV packs. On a primary/secondary chain this includes the secondary pack IR2091 drops. `None` (not `0`) when no pack decodes, so an EMS-controller dongle signals "sum the inverter dongles" rather than a misleading `0 Wh`.

Both Battery fields are whole **Wh** (`int`), matching the `remaining_battery_wh` sibling and keeping `model_dump()` stable.

## The topology boundary (why this shape)

Exploring `ems_2_inv_3_bat_a` (four separate dongle captures) confirmed the EMS controller and its managed inverters sit on **physically separate buses, each behind its own dongle**. A `Plant` is one dongle connection, and battery sub-bus addresses collide across inverters (each reuses `0x32–0x37`), so **no single Plant can hold all packs**. This PR fixes the **within-dongle** undercount (a primary/secondary chain has both packs on its dongle); **cross-inverter aggregation stays with the consumer** (hass), which holds the per-inverter connections. Documented in the `Plant` property docstring.

## Tests (TDD)

Every field written test-first. Battery: nominal, measured, `None`-cap, `num_cells`-absent → `v_out` fallback, both-absent → `None`. Plant: synthetic multi-pack sum, `None` on empty, skips undecodable packs, and a **fixture-backed** test decoding the real 2× Giv-Bat 5.2 capture → **5087 Wh** nominal (the sum IR2091 collapses). Full suite (1571) + tox (py311–314, lint, format, build) green.

## Consumer handoff (post-merge)

Boundary note to hass: repoint the "EMS Remaining Battery Energy" sensor off `Ems.remaining_battery_wh` (IR2091) to a cross-inverter sum of each managed inverter's `Plant.remaining_battery_energy_wh`. Library figure is nominal-basis Wh; IR2091 stays available for diagnostics.

## Summary by CodeRabbit

* **New Features**
  * Added battery remaining-energy values in Wh, calculated using both nominal and measured voltage bases.
  * Added a plant-level total for remaining battery energy across connected low-voltage packs.

* **Bug Fixes**
  * Improved handling of missing battery data, returning no value where key inputs are unavailable.
  * Updated plant totals to ignore undecodable packs instead of counting them as zero.

* **Tests**
  * Expanded coverage for battery energy calculations and plant-level energy totals.

### DIFF
--- a/givenergy_modbus/model/plant.py
+++ b/givenergy_modbus/model/plant.py
@@ -1894,6 +1894,24 @@ class Plant(GivEnergyBaseModel):
         return [Battery.from_register_cache(self.register_caches[i + 0x32]) for i in range(self.number_batteries)]
 
     @property
+    def remaining_battery_energy_wh(self) -> int | None:
+        """Total remaining battery energy (Wh) summed over this Plant's LV packs (#374).
+
+        Nominal-voltage basis (``Battery.remaining_energy_nominal_wh``). This sums every pack on
+        *this* dongle, so on an inverter with a primary/secondary chain it includes the secondary
+        pack the EMS controller's own figure (``Ems.remaining_battery_wh``, IR2091) drops.
+
+        Returns ``None`` (not ``0``) when no pack decodes — e.g. an EMS-controller dongle has no
+        LV battery sub-bus, so its Plant returns ``None`` rather than a misleading ``0 Wh``.
+
+        Scope: one Plant is one dongle, and battery sub-bus addresses collide across managed
+        inverters (each reuses 0x32-0x37), so this cannot span inverters. A consumer holding
+        several inverter connections sums their per-Plant values for the site total.
+        """
+        present = [e for e in (b.remaining_energy_nominal_wh for b in self.batteries) if e is not None]
+        return sum(present) if present else None
+
+    @property
     def hv_stacks(self) -> list[HvStack]:
         """Return HV battery stacks (BCU + per-module BMUs) for HV systems; [] for LV systems.
 
--- a/givenergy_modbus/model/battery.py
+++ b/givenergy_modbus/model/battery.py
@@ -1,7 +1,7 @@
 from enum import IntEnum
 from typing import ClassVar
 
-from pydantic import ConfigDict, create_model
+from pydantic import ConfigDict, computed_field, create_model
 
 from givenergy_modbus.model.register import (
     IR,
@@ -104,6 +104,10 @@ _BatteryBase = create_model(  # type: ignore[call-overload]
 )
 
 
+# LiFePO4 nominal cell voltage — 16 cells × 3.2 V = the GivEnergy "51.2 V" LV pack convention.
+_LIFEPO4_NOMINAL_CELL_V = 3.2
+
+
 class Battery(_BatteryBase, RegisterMetadataMixin):  # type: ignore[misc,valid-type]
     """GivEnergy battery data model."""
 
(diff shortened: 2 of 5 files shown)


dewet22-codex:
Starting a review now. I’ll focus on the Battery energy calculations, Plant aggregation boundary, and the #374/#318 EMS undercount handoff.

givenergy_modbus/model/plant.py
```diff
@@ -1893,6 +1893,24 @@ def batteries(self) -> list[Battery]:
             ]
         return [Battery.from_register_cache(self.register_caches[i + 0x32]) for i in range(self.number_batteries)]
 
+    @property
+    def remaining_battery_energy_wh(self) -> int | None:
```
dewet22-codex, line 1897: Amber/blocking for the #374 contract: this is a plain `@property`, so it is visible as `plant.remaining_battery_energy_wh` but absent from `Plant.model_dump()` / JSON. The tracking issue and investigation thread both framed the additive remedy as a Plant-level `computed_field`, matching the new Battery fields and the "CLI and other consumers get it for free" handoff. With the current form only attribute-based consumers see it. I think this needs `@computed_field` here as well, or the PR contract/body/tests need to narrow this to an attribute-only accessor.

dewet22-codex (changes requested):
Thanks for this. The pack-level calculations and the per-dongle aggregation shape look good, and the focused model tests pass locally.

I left one blocking comment on the Plant aggregate: it is implemented as a plain property, so it does not appear in `Plant.model_dump()` despite #374 framing this as a Plant-level computed field for downstream/CLI handoff.

Local check:

`uv run --group test pytest tests/model/test_battery.py tests/model/test_plant.py tests/model/test_plant_accessors.py` — 215 passed.

givenergy_modbus/model/plant.py
dewet22: Good flag, but keeping it a plain `@property` is deliberate — the premise that it should match the Battery computed fields conflates two different `model_dump` contracts:

- `Battery.model_dump()` is the **decoded-values** view, so `remaining_energy_*_wh` belong there as `@computed_field`.
- `Plant.model_dump()` is the **raw persistable state** — only `{capabilities, serials, register_caches}`. Every derived accessor (`inverter`, `batteries`, `ems`, `hv_stacks`) is a plain `@property` excluded from it; the decoded Battery objects (and their computed fields) don't appear in `Plant.model_dump()` either. A SOC-varying derivation doesn't belong in the plant's dumpable/persistable state.

So `remaining_battery_energy_wh` follows the established Plant convention: attribute-accessible (`plant.remaining_battery_energy_wh`, exactly as consumers read `plant.inverter` / `plant.batteries`), not in the raw-state dump. hass consumes it by attribute for the cross-inverter sum. I've narrowed the contract explicitly and pinned it with a test asserting it's accessible but absent from `model_dump()` (`test_remaining_battery_energy_wh_is_attribute_only_not_in_model_dump`, b3b236e). The change-detection hash is unaffected regardless — it hashes raw register blocks, not `model_dump()`.

PR #389, opened 2026-07-09
feat(model): placeholders for known-but-unreachable battery/meter addresses (#213)

### LINKED ISSUE #213: API: return a placeholder for known-but-unreachable addresses so a transiently-absent device stays present-but-unavailable
## Context

When `givenergy-hass` restarts and `detect(prior=…)` is called, a battery whose BMS is slow to answer is absent from `PlantTopologyMismatch.actual`. Part A of the HA resilience work (merged in givenergy-hass#140) handles this at the coordinator level: it retries the probe, and if the loss persists it keeps the full prior cached and fires a repair issue rather than baking in the reduced topology.

However, within a restart session a transiently-absent pack is entirely absent from `plant.batteries` (index-shifted or missing), so its entities read as absent/crash rather than *unavailable*. The coordinator works around this by keeping `client.plant.capabilities = exc.actual` for the live poll so data reads dispatch correctly for what responded — but the missing device's entities show as absent rather than unavailable until it is re-probed.

## Requested API change

When `detect(prior=…)` is called and an address listed in `prior` is not found in the detection sweep, instead of dropping it from `plant.batteries` / `plant.meters`, return an all-`None` placeholder entry at that index so:

- `plant.batteries` length matches the prior battery count (index alignment preserved)
- The placeholder's register values are all `None` (or a sentinel that renders as unavailable in HA)
- Subsequent `refresh()` calls attempt to read the address and populate it if it responds

This is the Part B handoff from the [givenergy-hass Part A plan](https://github.com/dewet22/givenergy-hass/pull/140). Part A neither needs nor regresses it — `sensor.py:~1172` (`_battery_index >= len(batteries)`) already renders a dropped-index entity unavailable. Part B would make the within-session experience cleaner: the device shows as present-but-unavailable rather than index-shifted or crashed.

## Proposed boundary

`client.plant.batteries[n]` returns a `Battery` (or `BatteryModule`) with all fields `None` when address `n` has a prior entry but no register cache — rather than raising `IndexError` or omitting the index.

I'm happy to discuss the implementation shape before writing the modbus-side PR.

dewet22:
Closes #213.

## What

`Plant.batteries` and `Plant.meters` used to silently drop any
capabilities-listed address whose register cache was empty. A battery that
stopped responding just vanished from the list — the consumer couldn't tell
"present but unreachable this poll" from "never existed".

Both accessors now emit an all-None placeholder — `from_register_cache(RegisterCache())` — for every capabilities-listed address, whether or not it responded. Index and key alignment is preserved. On a placeholder, `is_valid()` is `False` and every decoded field is `None`, so a consumer can distinguish the two cases.

## Scope

- **Batteries + meters only.** LV batteries are the demonstrated case of the disappearing-entry behaviour, and meters were named in the issue. No presence-marker gating — the placeholder is keyed purely off the capabilities address list.
- No production change beyond the two accessors; no new public types.

## Behaviour change (release-notes flag)

`len(plant.batteries)` now equals `len(capabilities.lv_battery_addresses)` rather than the count of *responding* packs, and `plant.meters` is keyed by every meter address, not just reachable ones. Anything iterating these needs to guard on `.is_valid()` (or a `None` field) rather than assuming every entry decodes. I'll flag this to the hass and cli agents once it lands.

## Tests

`tests/model/test_plant_accessors.py` — the two former "skips missing caches" tests are flipped to assert the placeholder contract (list length, `is_valid()` False, a representative field `None`). Full suite green (1587), tox green (py313/py314/format/lint/build).

### DIFF
--- a/givenergy_modbus/model/plant.py
+++ b/givenergy_modbus/model/plant.py
@@ -1884,12 +1884,18 @@ class Plant(GivEnergyBaseModel):
 
     @property
     def batteries(self) -> list[Battery]:
-        """Return Battery models for the Plant."""
+        """Return Battery models for the Plant.
+
+        A capabilities-listed address with no register cache yields an all-None placeholder
+        Battery at its index rather than being dropped (#213) — a transiently-absent pack
+        (e.g. a BMS slow to answer after a restart) stays present-but-unavailable and index-
+        aligned instead of shifting its siblings' entities. Subsequent refresh()es populate it
+        if it responds.
+        """
         if self.capabilities:
             return [
-                Battery.from_register_cache(self.register_caches[addr])
+                Battery.from_register_cache(self.register_caches.get(addr, RegisterCache()))
                 for addr in self.capabilities.lv_battery_addresses
-                if addr in self.register_caches
             ]
         return [Battery.from_register_cache(self.register_caches[i + 0x32]) for i in range(self.number_batteries)]
 
@@ -1976,13 +1982,17 @@ class Plant(GivEnergyBaseModel):
 
     @property
     def meters(self) -> dict[int, Meter]:
-        """Return Meter models keyed by device address."""
+        """Return Meter models keyed by device address.
+
+        A listed-but-uncached meter address yields an all-None placeholder Meter at its key
+        rather than being omitted (#213), so a transiently-absent meter reads as present-but-
+        unavailable and is repopulated on a later refresh() if it responds.
+        """
         if not self.capabilities or not self.capabilities.meter_addresses:
             return {}
         return {
-            addr: Meter.from_register_cache(self.register_caches[addr])
+            addr: Meter.from_register_cache(self.register_caches.get(addr, RegisterCache()))
             for addr in self.capabilities.meter_addresses
-            if addr in self.register_caches
         }
 
     @property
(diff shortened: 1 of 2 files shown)


dewet22-codex:
Starting a review now. I’ll focus on the behavioural diff and any invariants around the changed modbus path.

dewet22-codex (approved):
Reviewed the model-layer change and I don’t see blocking concerns. The empty-cache placeholder path stays inert for both `Battery` and `Meter`: the models validate, `is_valid()` is false, and `model_dump()` contains no non-`None` values. Index/key alignment also holds when an earlier listed battery address is missing and a later one is populated.

One handoff note I checked: HASS currently serves `exc.actual` on a persistent `PlantTopologyMismatch`, so it will not see these prior-listed placeholders until the planned consumer-side follow-up opts into keeping the prior topology for that tick. That looks like the expected sequencing rather than a modbus blocker, and the PR body already flags the downstream behaviour change.

Local checks:

`uv run --group test pytest tests/model/test_plant_accessors.py tests/model/test_plant_devices.py` — 40 passed
`uv run --group test pytest tests/model tests/client/test_detect.py` — 841 passed
`uv run --group test ruff check givenergy_modbus/model/plant.py tests/model/test_plant_accessors.py` — passed

PR #390, opened 2026-07-09
chore: structured issue forms for device validation + capture/probe intake

dewet22:
Adds the register-decode half of the cross-repo issue-template split (agreed with the hass side): this repo hosts device-validation and capture/probe intake; the hass integration keeps "something's wrong" and feature-request forms.

## What

Replaces the single free-text `.github/ISSUE_TEMPLATE.md` with structured GitHub issue-forms:

- **`new-device.yml`** — "New / unvalidated device". Free-text model, optional firmware, what's-unconfirmed, and a steer toward a ~10-minute `capture` recorded alongside the reporter's existing integration (GivTCP / the app). `capture` is the right first-contact tool because it reveals what a *working* setup actually polls — `export` presumes decode already works, which is the thing being validated.
- **`capture-probe.yml`** — "Submit a capture / probe". Context field foregrounded as the important one; instructions for both `capture` (passive wire recording) and a targeted `probe` (usually a directed follow-up).
- **`config.yml`** — disables blank issues; routes integration symptoms to the hass repo via a contact link.

## Notes

- Both forms tell the reporter that `capture`/`probe` **redact serials automatically**, so contributions arrive pre-sanitised — no redaction burden placed on the user, and consistent with the project's redaction discipline.
- The `givenergy-cli` invocations (`capture -o … -d 600`, `probe -t ir -b 0 -n 60 -d 0x11`) were taken from the current cli command definitions, not from memory.
- The hass contact-link URL resolves regardless of whether the hass chooser is live yet, so there's no merge-ordering dependency on the hass-side fast-follow.

No code or release impact — repository metadata only.

## Summary by CodeRabbit

* **New Features**
  * Added dedicated issue forms for submitting a new/unverified device and for capture/probe reports.
  * Improved issue reporting guidance with clearer prompts and automatic redaction notes.
  * Disabled blank issue creation and added a direct support path for Home Assistant integration problems.

* **Documentation**
  * Streamlined the standard issue template to focus submissions on the most useful details.


dewet22-codex:
Starting a review now. I’ll focus on the changed modbus behaviour, edge-case coverage, and any downstream HASS/CLI handoff implications.

.github/ISSUE_TEMPLATE/new-device.yml
```diff
@@ -0,0 +1,60 @@
...
+  - type: markdown
+    attributes:
+      value: |
+        Thanks for helping get your hardware supported. The single most useful thing
+        you can provide is a **capture** — a recording of the Modbus traffic between
+        your inverter and whatever already talks to it (GivTCP, the GivEnergy mobile
+        app, another integration). That lets us see exactly which registers a working
+        setup reads, which is how we work out the decode for a new device.
+
+        Serials are redacted automatically, so a capture is safe to share as-is.
```
dewet22-codex: [blocking] This over-promises the privacy guarantee for the exact case this form targets. `givenergy-cli capture` only redacts frames the library can decode; undecodable frames pass through unchanged, and `scripts/scan_capture_serials.py` exists because new/split serial locations can still leak. For new or unvalidated devices, that unknown-layout edge is especially relevant. I think this needs to mirror the CLI wording instead: serials are redacted in decodable frames, but reporters should skim before posting publicly.

.github/ISSUE_TEMPLATE/capture-probe.yml
```diff
@@ -0,0 +1,55 @@
...
+labels: ["to investigate"]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Use this to send in data: a **capture** (passive recording of the traffic from a
+        working integration) or a **probe** (a targeted read of specific register banks,
+        usually because we asked for one on another issue).
+
+        Serials are redacted automatically by both commands, so the output is safe to
```
dewet22-codex: [blocking] This also says `probe` output is automatically redacted, but the current CLI probe renderer prints register words directly. If the requested range includes serial-bearing registers, the output is not going through the capture redactor. I think this needs softer wording for both commands, or separate wording for `capture` and `probe`, before we steer users to share the attachment publicly.

dewet22-codex (changes requested):
Thanks, the issue-form split itself looks sensible and the command shapes match the CLI. I’m requesting changes for the privacy wording only: both forms currently describe capture/probe output as safe to share as-is, which overstates the actual redaction guarantee.

.github/ISSUE_TEMPLATE/new-device.yml
dewet22: Agreed it overstated the guarantee. Reworded: capture redacts the serials it recognises, and where a new device carries unrecognised ones we take on removing residual personal info when we process the report — with a private upload route via the repo's security-advisory reporting for anyone who'd rather not attach on a public issue. No asking reporters to vet hex frames.

.github/ISSUE_TEMPLATE/capture-probe.yml
dewet22: Right that probe bypasses the redactor. The intro now states that capture redacts recognised serials while probe prints raw, and that we strip any residual PII our side; the blanket "safe as-is" is gone, with the same private security-advisory route offered for anyone who'd rather not share publicly.

dewet22-codex:
Starting a re-review now. I’ll focus on the previous privacy/redaction wording blockers and check the updated issue forms in that scope.

dewet22-codex (approved):
I’ve re-reviewed the updated head and have no blocking concerns now.

The previous privacy blockers are resolved: the new wording no longer says capture/probe output is safe to share as-is, it scopes `capture` redaction to recognised serials, and it correctly states that `probe` output is raw. I also checked that private vulnerability reporting is enabled for the repo, so the private-upload route in the forms is available.

Local checks: issue-form YAML parses cleanly; `git diff --check` is clean.

PR #391, opened 2026-07-09
feat(testing): MockPlant.from_spec — declarative synthetic register state (#324)

### LINKED ISSUE #324: mock-server: synthetic-capture + sentinel mode for app/register interrogation
The #265 investigation produced a small helper (`capgen.py`) that: re-wraps non-canonical captures into the mock-readable format; synthesises valid `rx` frames for arbitrary device/register banks via the library's own response encoder; round-trips them through `plant.update` to confirm they survive the commit guards; and seeds sentinel banks (`raw == address`, the `register-identification.md` technique). Together that turns `mock-server` into a no-hardware way to drive the real GE app against arbitrary synthetic state.

**Proposal (low priority):** promote a minimal version into the testing helpers — e.g. `MockPlant.from_spec({device: {(type, base): values}})` and a `mock-server --sentinel` mode — so synthetic-capture construction is reproducible rather than a one-off script.

dewet22:
Closes #324.

## What

`MockPlant.from_spec({device: {(HR|IR, base): values}}, verify=True)` — build a servable mock from a pure register spec, no base capture required. Together with `start()` / `givenergy-cli mock-server`, this turns arbitrary synthetic state into something the real GE app (or this library's client) can be driven against.

Most of #324 turned out to have already shipped in `6631d43` (the register-identification work: `from_sentinels`, `identify()`, `sentinel_devices()`, `docs/register-identification.md`). This PR adds the remaining library half — declarative arbitrary-value seeding plus the one capgen piece with real teeth:

## The verify step

A synthetic bank can serve fine over the wire yet be **silently rejected on ingest** by the client-side commit guards — which surfaces as "the client sees nothing" mid-interrogation, a confusing dead-end. With `verify=True` (the default) every HR/IR bank is round-tripped through a scratch `Plant.update()` at construction:

- each bank is fed **twice**, mirroring detect()'s corroborating cold-start re-read (#289), so battery banks commit exactly as they would live;
- any register that fails to land raises `ValueError` naming the device and bank;
- non-HR/IR register classes are rejected under verify (they don't exist on the read wire) and allowed with `verify=False` for direct cache seeding.

The TDD run made the case for the feature by itself: the first "healthy" synthetic inverter bank in the test was silently discarded by a guard I hadn't accounted for (serial validation on HR(13-17)) — exactly the invisible-rejection mode verify exists to surface.

## Scope

- Library-side only. The `mock-server --spec` / `--sentinel` CLI flags belong to givenergy-cli; I'll offer that wiring via the usual coordination route once this lands.
- No production/client code touched — `givenergy_modbus/testing/` + tests.

## Tests

`tests/testing/test_from_spec.py` — construction, healthy-bank verification (incl. the cold-start double-feed), guard-rejection error naming, unsupported-class rejection, verify=False passthrough. Full suite 1592 green, tox green (py313/py314/format/lint/build).

## Summary by CodeRabbit

* **New Features**
  * Added a way to create a mock plant directly from an in-memory register specification.
  * Verification can be enabled to confirm supplied values are accepted and stored correctly.
* **Bug Fixes**
  * Improved validation so invalid or unsupported register data is rejected with clearer errors.
  * Added safeguards for inconsistent battery/register values during mock setup.
* **Tests**
  * Expanded coverage for verified and unverified mock plant creation, including supported and unsupported register types.

### DIFF
--- a/givenergy_modbus/testing/mock_plant.py
+++ b/givenergy_modbus/testing/mock_plant.py
@@ -227,7 +227,12 @@ def _make_device_serials_distinct(plant: Plant) -> None:
         )
 
 
-def _verify_spec_commits(spec: dict[int, dict[tuple[type[Register], int], Sequence[int]]]) -> None:
+def _verify_spec_commits(
+    spec: dict[int, dict[tuple[type[Register], int], Sequence[int]]],
+    *,
+    inverter_serial: str = _FALLBACK_SERIAL,
+    adapter_serial: str = _FALLBACK_SERIAL,
+) -> None:
     """Round-trip every HR/IR bank in *spec* through a scratch ``Plant.update()`` (#324).
 
     Feeds each bank twice — the second pass is the corroborating re-read that commits a
@@ -235,6 +240,10 @@ def _verify_spec_commits(spec: dict[int, dict[tuple[type[Register], int], Sequen
     landed in the plant's caches. Raises ValueError naming the failures, so a synthetic
     bank the commit guards would silently reject (e.g. an internally-impossible battery
     frame, #350) fails fast at construction instead of serving state no client can ingest.
+
+    The envelope serials mirror what the built mock will actually serve, so verification
+    replicates serve-time inputs exactly (they don't gate commits today — update() only
+    adopts them as state — but fidelity here is free).
     """
     from givenergy_modbus.pdu.read_registers import (
         ReadHoldingRegistersResponse,
@@ -264,17 +273,19 @@ def _verify_spec_commits(spec: dict[int, dict[tuple[type[Register], int], Sequen
                     register_count=len(values),
                     register_values=list(values),
                     # Envelope serials a wire decode would carry; update() reads them.
-                    inverter_serial_number=_FALLBACK_SERIAL,
-                    data_adapter_serial_number=_FALLBACK_SERIAL,
+                    inverter_serial_number=inverter_serial or _FALLBACK_SERIAL,
+                    data_adapter_serial_number=adapter_serial or _FALLBACK_SERIAL,
                 )
                 plant.update(pdu)
 
+    _empty = RegisterCache()
     failures = [
         f"0x{device_address:02x}:{reg_cls.__name__}({base + i})"
         for device_address, banks in spec.items()
+        for cache in (plant.register_caches.get(device_address, _empty),)
         for (reg_cls, base), values in banks.items()
         for i, value in enumerate(values)
-        if plant.register_caches.get(device_address, RegisterCache()).get(reg_cls(base + i)) != value
+        if cache.get(reg_cls(base + i)) != value
     ]
     if failures:
         raise ValueError(
@@ -396,7 +407,7 @@ class MockPlant:
                         )
                     cache[reg_cls(base + i)] = value
         if verify:
-            _verify_spec_commits(spec)
+            _verify_spec_commits(spec, inverter_serial=inverter_serial, adapter_serial=adapter_serial)
         return cls(devices, inverter_serial=inverter_serial, adapter_serial=adapter_serial)
 
     # -- serving ---------------------------------------------------------------
--- a/givenergy_modbus/testing/mock_plant.py
+++ b/givenergy_modbus/testing/mock_plant.py
@@ -35,6 +35,7 @@ from __future__ import annotations
 import argparse
 import asyncio
 import logging
+from collections.abc import Sequence
 from pathlib import Path
 from typing import cast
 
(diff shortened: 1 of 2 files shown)


dewet22-codex:
Starting a review now. I’ll focus on behavioural correctness, invariants in the touched modbus paths, and any downstream HASS/CLI handoff implications.

givenergy_modbus/testing/mock_plant.py
```diff
@@ -292,6 +352,43 @@ def from_sentinels(
...
+        would live. Raises :class:`ValueError` naming every register that failed to
+        commit. Only HR/IR banks exist on the read wire; spec any other register class
+        with ``verify=False`` (direct cache seeding, no round-trip).
+        """
+        devices: dict[int, RegisterCache] = {}
+        for device_address, banks in spec.items():
+            cache = devices.setdefault(device_address, RegisterCache())
+            for (reg_cls, base), values in banks.items():
+                for i, value in enumerate(values):
+                    cache[reg_cls(base + i)] = value
```
dewet22-codex, line 411: [blocking] `verify=True` still accepts values that cannot be served on the read wire. For example, `MockPlant.from_spec({0x11: {(HR, 0): [70000]}}, verify=True)` constructs a mock and stores `HR(0)=70000`, but the first `ReadHoldingRegistersRequest(device_address=0x11, base_register=0, register_count=1)` response raises from `struct.pack("H", ...)` during `resp.encode()`. Negative values behave the same way. That leaves the default verified path able to produce a mock that crashes when a client reads a fully covered range, which is exactly the failure mode `verify=True` is meant to rule out. I think this path needs to validate raw spec words as unsigned 16-bit values, or have verification actually encode/decode the synthetic responses, with a regression for the out-of-range case.

dewet22-codex (changes requested):
Thanks, this is a useful testing helper and the commit-guard verification shape makes sense. I’m requesting changes for one servability gap in the default `verify=True` path: it can still accept raw register words that cannot be encoded into a Modbus response.

givenergy_modbus/testing/mock_plant.py
dewet22: Agreed — good catch, and it was exactly the crash-on-read mode verify claims to rule out. Fixed with an unconditional range check at construction (0..0xFFFF), not just under verify: an unencodable word can never be served regardless of the verify flag, so gating the check on it would leave the same crash reachable via verify=False. Regression parametrised over 70000 / -1 / 0x10000 in both verify modes. I went with the validate-raw-words option rather than encode/decode round-tripping — the struct.pack boundary is the only place raw-word validity bites, so the direct check covers it without doubling the verification cost.

dewet22-codex:
Starting a re-review now. I’ll focus on the from_spec verify=True servability blocker and any related changes/comments on the updated head.

givenergy_modbus/testing/mock_plant.py
```diff
@@ -292,6 +352,53 @@ def from_sentinels(
...
+        Every value must fit an unsigned 16-bit register word (0..0xFFFF) regardless of
+        ``verify`` — an out-of-range word can never be encoded onto the wire, and would
+        otherwise crash ``struct.pack`` at serve time on the first client read.
+        """
+        devices: dict[int, RegisterCache] = {}
+        for device_address, banks in spec.items():
+            cache = devices.setdefault(device_address, RegisterCache())
+            for (reg_cls, base), values in banks.items():
+                for i, value in enumerate(values):
+                    if not 0 <= value <= 0xFFFF:
```
dewet22-codex: [blocking] This range check fixes the 70000 / -1 case, but it still lets non-integer values through. For example, MockPlant.from_spec({0x11: {(HR, 0): [1.5]}}, verify=True) constructs successfully, then a covered ReadHoldingRegistersRequest(..., base_register=0, register_count=1) crashes at resp.encode() with struct.pack("H", 1.5). That keeps the same verified-but-not-servable failure mode open for specs loaded from JSON/YAML or other dynamic callers. I think the validation needs to assert an integer register word as well as the 0..0xFFFF range, with a regression for a fractional value.

dewet22-codex (changes requested):
Thanks, the signed/out-of-range case from the previous review is fixed and covered in both verify modes. I found one remaining edge in the same servability class: non-integer numeric values still pass construction and fail on the first read. I left an inline note on the range check with the fractional-value reproducer.

givenergy_modbus/testing/mock_plant.py
```diff
@@ -226,6 +227,65 @@ def _make_device_serials_distinct(plant: Plant) -> None:
...
+            for (reg_cls, base), values in banks.items():
+                pdu = pdu_for[reg_cls](
+                    device_address=device_address,
+                    base_register=base,
+                    register_count=len(values),
+                    register_values=list(values),
+                    # Envelope serials a wire decode would carry; update() reads them.
+                    inverter_serial_number=_FALLBACK_SERIAL,
+                    data_adapter_serial_number=_FALLBACK_SERIAL,
+                )
```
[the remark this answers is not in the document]
dewet22-claude: Taken — verification now receives the same serials the mock will serve. Worth noting the premise is narrower than stated: envelope serials don't gate _commit_bank today (update() only adopts them as state; the serial check that does gate commits reads the bank contents at HR(13-17)), so this couldn't actually flip a verify verdict. But verification replicating serve-time inputs exactly is the right construction regardless — same principle as the double-feed — so the fidelity is worth the three lines.

givenergy_modbus/testing/mock_plant.py
```diff
@@ -226,6 +227,65 @@ def _make_device_serials_distinct(plant: Plant) -> None:
...
+                plant.update(pdu)
+
+    failures = [
+        f"0x{device_address:02x}:{reg_cls.__name__}({base + i})"
+        for device_address, banks in spec.items()
+        for (reg_cls, base), values in banks.items()
+        for i, value in enumerate(values)
+        if plant.register_caches.get(device_address, RegisterCache()).get(reg_cls(base + i)) != value
+    ]
```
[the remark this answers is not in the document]
dewet22-claude: Fixed — one empty-cache sentinel hoisted out of the comprehension and the per-device lookup done once per device rather than per register.

dewet22-codex:
Starting a re-review now. I will keep it scoped to the updated validation and tests.

dewet22-codex:
Re-review result: the head is still 23adfdc, which is the same commit as my fractional-value blocker. The unsigned-range fix is still present, but there is no update yet for non-integer raw words such as 1.5, so the existing changes-requested review remains the active blocker. I am not adding a duplicate inline review on the same line.

givenergy_modbus/testing/mock_plant.py
dewet22-claude: Fixed — the check now also requires an actual int (isinstance, alongside 0..0xFFFF): struct.pack rejects even an integral 2.0, so both fractional and whole-number floats from dynamic spec sources fail fast at construction. Regression parametrised over 1.5 / 2.0 in both verify modes.

dewet22-codex:
Starting another re-review now. I will keep this focused on the outstanding from_spec validation thread and any new commits since the last pass.

dewet22-codex (approved):
Re-reviewed 20ef8f0 and I have no blocking concerns now. The from_spec validation now rejects fractional and integral floats before serving, while uint16 integer words still serve in the probe. Focused tests pass locally: uv run --group test pytest tests/testing/test_from_spec.py.

PR #392, opened 2026-07-09
fix(client): fail-closed envelope-serial redaction in FrameRedactor

dewet22:
## What

The first-ever write-path capture (Gateway, tonight) exposed a FrameRedactor blind spot: the Gateway's `WriteHoldingRegisterResponse` envelopes carry a NUL-interrupted serial variant — prefix + NUL bytes + partial date + live unit digits — that doesn't parse as a GE serial. The envelope path used the fail-open `redact_serial`, which returns unrecognised shapes unchanged, so the real unit digits passed verbatim into the shared capture (~30 frames). Both `scan_capture_serials.py` patterns structurally miss the shape too: the NULs break the two-letter-prefix match, and the surviving tail is too short for the tail pattern.

## Fix

- **Envelope serials → `redact_serial_strict`** (blank anything unparseable). This is the same fail-closed treatment `Plant.redact()` already applies to these exact header fields (#212/#214) — the envelope has no overlapping-group ambiguity, unlike the register-payload path, which must stay fail-open by design.
- **Scanner gains a third, NUL-tolerant pattern** so the detector catches this family wherever it appears next; validated against the leaking capture (flags it) and the full committed corpus (no false positives).

The affected shared capture has been re-redacted through the fixed path (exactly the 30 frames rewritten, re-scan clean, still decodes) and the public copy taken down.

## Exposure

Nil in practice until tonight: this shape only appears on write responses, and no write-path capture had ever been taken. But the cli `capture` command and the HA integration's capture path both rely on FrameRedactor, and write-path capture requests are now circulating for the Gateway work — so I intend to cut a patch release promptly once this lands and ask the downstream integrations to bump before soliciting more write captures.

## Tests

New `test_frame_redactor_blanks_unrecognised_envelope_serial` (fabricated placeholder serials, real leak shape) — red against the fail-open path, green with strict; existing envelope test pins that recognised serials keep the date-preserving redaction. Full suite 1604 green, tox green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved serial-number redaction so any values that do not look like valid device serials are now hidden more reliably.
  * Reduced the chance of sensitive serial information appearing in captured output when data contains unusual byte patterns.
  * Expanded serial scanning to detect an additional masked pattern that could previously slip through checks.

### DIFF
--- a/givenergy_modbus/client/client.py
+++ b/givenergy_modbus/client/client.py
@@ -179,11 +179,19 @@ class FrameRedactor:
         if isinstance(pdu, LanConfigBroadcast):
             return pdu.redact().encode()
 
-        # Redact envelope serials (present on all Transparent PDUs)
+        # Redact envelope serials (present on all Transparent PDUs). FAIL-CLOSED
+        # (redact_serial_strict): an envelope value that doesn't parse as a GE serial is
+        # blanked, not passed through. The Gateway's write-response envelopes carry a
+        # NUL-interrupted serial variant that fails the pattern, and the fail-open
+        # redact_serial leaked its real unit digits verbatim (first write-path capture,
+        # 2026-07-09). Unlike the register-payload path below — which must fail open
+        # because serial groups overlap ordinary data — the envelope is unambiguous,
+        # so it takes the same strict treatment Plant.redact() applies to these exact
+        # fields (#212/#214).
         if hasattr(pdu, "data_adapter_serial_number"):
-            pdu.data_adapter_serial_number = Converter.redact_serial(pdu.data_adapter_serial_number) or ""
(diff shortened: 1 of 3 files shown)


dewet22-codex:
Starting a review now. I will focus on behavioural correctness, invariants in the touched modbus paths, and any downstream handoff implications.

dewet22-codex (approved):
Reviewed 1841cf4 and I have no blocking concerns. The strict envelope path blanks the NUL-interrupted serial before re-encoding, the recognised adapter serial remains date-redacted, and the scanner catches the synthetic NUL-interrupted shape while the committed corpus scans clean. Focused local checks passed: uv run --group test pytest tests/client/test_capture.py; uv run python scripts/scan_capture_serials.py. I noted the decoded blank is star padding rather than NUL padding, so I do not think the existing CodeRabbit test assertion note blocks this fix.

Issue #379, opened 2026-07-08
Bmu/AioBatteryModule per-cell decode ignores declared bounds (1852 °C artefact)


dewet22:
`decode_cells_temps_serial()` applies no min/max — the LUT's declared bounds are explicitly inert under the imperative decode (`hv_bcu.py`, noted in its own docstring). Wire-evidenced consequence on the committed 3ph HV fixture (#377): module 0x55's split serial (see the split-layout issue) puts `"HY"` (0x4859) at IR(110), which decodes as **`t_cell_21 = 1852.1 °C`** and flows to consumers unchecked. The same module populates only 12 of 24 temp sensors (zeros elsewhere — which decode as 0.0 °C rather than absent).

Proposal: bound-check in the shared decode so out-of-range raw values decode to `None` —
- temps ≈ [-60, 150], mirroring `BatteryRegisterGetter`'s `t_max`/`t_min` bounds;
- voltages [1.0, 5.0], already declared on the Defs but currently inert.

Affects both `Bmu` (HV) and `AioBatteryModule` (shared decode), so it needs its own small PR; `test_three_phase_from_capture.py` pins the 1852.1 artefact as characterisation today and the pin flips to `None` with the fix.

PR #395, opened 2026-07-10
fix(model): widen v_battery bound to admit HV integrated-stack voltage (AIO)

dewet22:
## What

On an All-in-One, the inverter-level `v_battery` register (IR50) carries the **HV integrated stack** voltage (~300–500 V), not an LV pack (~50 V). The Def's `max` bound was 100 V — LV-calibrated — so a real AIO stack voltage decoded to `None`, giving those owners a permanently-Unknown "Battery Voltage" entity (AberDino report via the HA side).

Same failure class as #381 (signed temps) and #379 (cell bounds): a bound calibrated on one hardware family silently suppressing a legitimate reading on another.

## Fix

`v_battery` Def `max` 100 → 600 V. LV packs unchanged, HV stacks admitted, and raw garbage still clamps (65535 → 655.35 V → None). The three-phase inverter is unaffected — it reads battery voltage from its own registers (`v_battery_lv/cv/bms/pcs`, already `max=1000`), not IR50.

Bounds here are a sanity net, not model knowledge; the splice guards remain the real corruption defence on battery data, so blunting this ceiling on LV units costs nothing meaningful.

## Evidence

- Live AIO capture: IR50 = 31348 → **313.48 V**, matching the same plant's BCU cluster voltage (314.1 V) to within 1 V.
- Committed corpus AIO fixtures read 31278 / 31314 — sane, slowly-drifting stack voltages.

## Tests

- `test_v_battery_bounds_admit_hv_stack_voltage` — parametrised: LV (53.87) unchanged, two HV cases (313.48 / 471.80) now decode, garbage (655.35) still None. Red on the two HV cases before the widening.
- Full-cycle integration fence: the `aio_redetect_clean` case now pins `v_battery == 312.78` end-to-end through the live detect→load_config→refresh cycle.

Full suite 1607 green, tox green.

## Downstream

Unblocks the hass-side row hass was about to gate off (their note asked whether to drop it permanently — answer was no, it's real, we were suppressing it). I'll confirm once this lands.

### DIFF
--- a/givenergy_modbus/model/inverter.py
+++ b/givenergy_modbus/model/inverter.py
@@ -974,7 +974,11 @@ class SinglePhaseInverterRegisterGetter(RegisterGetter):
         # as a deprecated alias on the inverter classes for a release.
         "work_time_total_hours": Def(C.uint32, None, IR(47), IR(48), max=876_000),
         "system_mode": Def(C.uint16, None, IR(49)),
-        "v_battery": Def(C.centi, None, IR(50), min=0.0, max=100.0),
+        # max 600 V (not 100): IR(50) is inverter-level battery voltage, and on an AIO it
+        # carries the HV integrated stack (~300-500 V), not just an LV pack (~50 V). The old
+        # LV-calibrated 100 V ceiling clamped a real HV reading to None (permanently-Unknown
+        # entity). 600 V admits both while still suppressing raw garbage (65535 -> 655.35).
+        "v_battery": Def(C.centi, None, IR(50), min=0.0, max=600.0),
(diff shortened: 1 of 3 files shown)


dewet22-codex:
Starting a review now. I will focus on behavioural correctness, invariants in the touched modbus paths, and any downstream handoff implications.

dewet22-codex (approved):
Reviewed 00b6028 and I have no blocking concerns. The v_battery bound change is narrowly scoped, keeps the high-end garbage clamp above 600 V, and the AIO value is pinned through the live-cycle integration case. Focused local checks passed: test_v_battery_bounds_admit_hv_stack_voltage, aio_redetect_clean, and the full test_mock_plant_integration.py file.

PR #396, opened 2026-07-10
feat(model): Plant.devices canonical topology walk (#106)

### LINKED ISSUE #106: Tracking: graph-shaped Plant abstraction for multi-device topologies

dewet22:
Revives and completes #106 as a **consumer-driven backport**: the Home Assistant integration hand-built its device registry three times over, and this lifts that proven shape into the library as one canonical topology walk. Contract reviewed and **acked byte-for-byte by the hass consumer** before implementation (identity schemes verified against their live registry — zero HA device-registry migration).

## What changed

`Plant.devices` (a derived `@property`, nothing added to `model_dump`) is now a **flat topology walk**: a root row (`parent=None`, index 0, `is_control_authority=True`) followed by child rows sorted `(device_type.value, identity)`, each with a `parent` identity reference. `PlantDevice` gained `identity`, `parent`, `firmware_version`, `is_valid`, `is_control_authority`, `data_source`; dropped `model` (verified zero external consumers — hass and cli both).

**Identity contract** (plant-independent for any device with a serial, so a pack moved between inverters keeps its identity):
- root = plant serial; LV pack / AIO module / HV BMU = own serial; HV stack = `{plant_serial}_hvstack_{addr:#04x}`; blinded managed inverter = `{serial}_managed`; meter = own serial else `{plant_serial}_meter_{addr}`; serial-less placeholders = `{plant_serial}_battery_{slot}` / `_module_{addr}` / `_hvstack_..._bmu_{n}`.
- Exactly one root and one control authority (Gateway > EMS > inverter — the #201/#373 "controls live here" rule as data).

## Design notes

- **Derived-projection architecture kept** — no first-class persistent graph nodes, no Transport abstraction, no cache-ownership migration. Phases 3–4 of the original #106 are **deferred as YAGNI with explicit triggers** (second protocol → Transport; package extraction → cache ownership); recorded for the #106 close.
- **AIO double-count resolved**: `bcu_stacks` and `aio_battery_module_addresses` describe the same physical 0x50–0x53 modules, so the walk emits the HV_STACK row but not its nested BMU rows on AIO plants — matching the consumer's stack+modules device model.
- **Robust serial source**: identities anchor on the register-first `inverter_serial` (#227), so the offline `from_caches` path (no envelope serial) still yields real identities.

## Tests

`tests/model/test_topology_walk.py` — exact golden-fixture walk pins across all five topologies (hybrid, AIO, EMS, gateway, 3ph HV) + structural-invariant property tests (one root, one authority, parents resolve, non-empty identities, determinism). Live full-cycle fence (`test_mock_plant_integration.py`) asserts the walk + identity uniqueness on the 3ph case. Full suite 1623 green, tox green (py313/py314/format/lint/build).

## Follow-ups (not blocking; flagged for the review)

- **Meter identity** is now serial-first (`{plant_serial}_meter_{addr}` only as fallback) — a small refinement beyond the acked contract; hass surfaces no meter device today so it's net-new, but worth confirming.
- **AIO-module parent**: currently the root inverter; the physically-truer parent is the stack. Identity-neutral (own serial either way), so no migration impact — a parent-ref question for hass.
- **MockPlant `_make_device_serials_distinct`** only disambiguates 0x50–0x53, so a 6-module HV stack's last two BMUs collide even on the live serve path; the live uniqueness assertion is `len(set)==len-1` with a comment pending that testing-infra fix.
- Deferred Minors: the double-decode of sub-devices per `.devices` call (perf only); the all-or-nothing AIO/BMU skip guard (no real topology exercises it).

Closes #106 on merge (with the phases-3/4 deferral recorded there).

## Summary by CodeRabbit

* **New Features**
  * Added a flat, deterministic plant topology view with one controller root and linked child devices.
  * Device entries now include stable identities, parent relationships, firmware versions, validity status, control authority, and data-source metadata.
  * Added support for clearer topology reporting across inverters, batteries, HV stacks, meters, EMS systems, and gateways.
  * Unavailable devices are represented with synthetic identities and clearly marked invalid status.

* **Tests**
  * Expanded coverage for topology ordering, hierarchy, identity handling, firmware reporting, and unavailable devices.

### DIFF
--- a/givenergy_modbus/model/devices.py
+++ b/givenergy_modbus/model/devices.py
@@ -346,24 +346,36 @@ class Inverter:
 
 @dataclass(frozen=True)
 class PlantDevice:
-    """One enumerated device on a Plant, tagged with its type and identity.
+    """One row in the #106 flat topology walk, tagged with its type and identity.
 
     A thin, immutable wrapper pairing an already-decoded typed model
     (:attr:`device`) with a generic :class:`DeviceType` discriminator and,
-    where available, a serial number and model. Consumers enumerate
-    :attr:`Plant.devices` to discover what is on a plant and read rich fields
-    through :attr:`device`.
+    where available, a serial number and firmware version. Consumers
+    enumerate :attr:`Plant.devices` to discover what is on a plant and read
+    rich fields through :attr:`device`.
 
     The wrapper exists because the concrete device models share no common
     base — the inverter facade is a plain class, batteries / meters / EMS /
     gateway are Pydantic models, an HV stack is a dataclass — so tagging them
     from the outside keeps the enumeration additive rather than mutating every
-    model. No Modbus wire address or :class:`RegisterCache` appears here:
-    identity is by serial, and the wire address (where relevant) stays
-    reachable on the underlying model (e.g. ``HvStack.device_address``).
+    model. No Modbus wire address appears as a field — identity strings may
+    embed one (e.g. an HV stack's ``{plant_serial}_hvstack_{addr:#04x}``) but
+    the wire address itself stays reachable on the underlying model where
+    relevant (e.g. ``HvStack.device_address``).
+
+    :attr:`identity` is the stable, library-computed device identity —
+    serial-anchored, and plant-independent for any device that carries its
+    own serial. :attr:`parent` links rows into the flat walk; it is ``None``
+    only for the single root row. :attr:`is_control_authority` marks the one
+    row that owns the plant's write surface.
     """
 
+    identity: str
     device_type: DeviceType
+    parent: str | None
     device: Any
     serial_number: str | None = None
-    model: Any | None = None
+    firmware_version: str | None = None
+    is_valid: bool = True
+    is_control_authority: bool = False
+    data_source: DataSource | None = None
--- a/givenergy_modbus/model/plant.py
+++ b/givenergy_modbus/model/plant.py
@@ -2194,29 +2194,34 @@ class Plant(GivEnergyBaseModel):
         # --- EMS-managed inverters: their own INVERTER rows, parented to the EMS root ---
         if ems_root is not None:
             for uinv in self.inverters:
-                serial = uinv.serial_number or ""
+                managed_serial = uinv.serial_number or ""
+                if uinv.is_blinded:
+                    identity = f"{managed_serial}_managed" if managed_serial else ""
+                else:
+                    identity = managed_serial
                 child_rows.append(
                     PlantDevice(
-                        identity=f"{serial}_managed" if uinv.is_blinded else serial,
+                        identity=identity,
                         device_type=DeviceType.INVERTER,
                         parent=root_identity,
                         device=uinv,
-                        serial_number=serial or None,
+                        serial_number=managed_serial or None,
                         firmware_version=None if uinv.is_blinded else _walk_firmware(DeviceType.INVERTER, uinv, self),
                         data_source=uinv.data_source,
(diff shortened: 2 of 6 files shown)


dewet22-codex:
Starting a review now. I’ll focus on behavioural correctness, protocol/model invariants, and compatibility with the downstream HASS/CLI consumers.

dewet22-codex (approved):
I’ve reviewed this and have no blocking concerns. The topology walk contract is pinned across the expected plant shapes, root/control-authority selection is consistent with the downstream HASS registry shape, and the serial-first identities held up in the offline `from_caches` probe without passing envelope serials.

I also checked CodeRabbit’s double-decode note. I agree it is a possible performance clean-up for the non-EMS/non-gateway path, but I don’t see it changing behaviour or invariants, so I’m not treating it as blocking for this PR. Codecov is still below its patch threshold, but the focused model/topology coverage looks broad enough for the behavioural surface being added.

Verified with `uv run --group test pytest tests/model/test_plant_devices.py tests/model/test_topology_walk.py tests/model/test_devices.py::test_plant_device_construction_and_read_through tests/model/test_devices.py::test_plant_device_optional_fields_default_to_none tests/model/test_plant_phase3.py tests/model/test_aio_battery.py::test_plant_devices_emits_battery_module_rows_under_inverter tests/client/test_mock_plant_integration.py -q` and `uv lock --check`.

PR #397, opened 2026-07-10
fix(mock): disambiguate HV BMU serials across the full 0x50-0x6F band

dewet22:
## What

`MockPlant._make_device_serials_distinct` re-tails redaction-collapsed serials so replayed multi-instance devices serve distinct identities. For the AIO/HV-module family it walked only `range(0x50, 0x54)` — a range sized for AIO's 4-module cap (`_AIO_MAX_MODULES`). HV BMU stacks share the exact same per-address IR(114-118) serial layout as AIO modules but can span up to 6 modules, so on the `three_phase_hv_a` 6-module fixture the BMUs at 0x54/0x55 kept the collapsed placeholder (`HY2336G000`) even through the live serve path.

This widens the walk to the existing `_HV_BMU_BAND_END` (0x70) constant from `plant.py`, so it covers the full HV BMU per-module band. The existing `a in caches` filter already scopes it to real, populated addresses, so an AIO still touches only its 4 and a 6-module HV stack touches its 6 — no second hardcoded per-topology constant.

## Why not derive from `capabilities`

The topology-walk gap comment suggested deriving the range from `plant.capabilities.hv_bmu_addresses` / `bcu_stacks`. That turns out not to be viable here: `_make_device_serials_distinct` runs on the mock-*server* plant built by `plant_from_capture`, whose `capabilities` is always `None` — it's only ever populated by `Client.detect()` on the client side, never during capture replay. The `a in caches` band filter is actually more precise than a capabilities-derived candidate list would be: it keys off exactly the addresses the capture proved present.

## Test

Tightens `_assert_three_phase_hv` in `test_mock_plant_integration.py` back to full identity uniqueness (`len(set(ids)) == len(ids)`) — it had been weakened on the topology-walk branch to `len(ids) - 1` to tolerate the one known collapsed pair, with a comment flagging the gap. Verified failing before the fix, passing after; full suite 1623 passed, ruff/mypy clean.

## Notes

The old docstring claimed HV BMUs "stride within a single 0x70 cache" and were out of scope — that described an earlier layout and is now stale; the real fixtures show BMUs at separate per-address IR caches, same shape as AIO modules. I've updated it to match. The 6-module stride was wire-confirmed on lamztib's HV capture, but I haven't verified stacks larger than 6 against live data — the band tolerates up to 32 addresses, which is speculative headroom rather than tested.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved device identification for high-voltage battery systems.
  * Prevented duplicate identities among battery management modules in three-phase high-voltage setups.
  * Ensured all detected modules are reported as distinct through the live serving path.

### DIFF
--- a/givenergy_modbus/testing/mock_plant.py
+++ b/givenergy_modbus/testing/mock_plant.py
@@ -43,7 +43,7 @@ from givenergy_modbus.framer import ServerFramer
 from givenergy_modbus.model.aio_battery import AioBatteryModuleRegisterGetter
 from givenergy_modbus.model.battery import BatteryRegisterGetter
 from givenergy_modbus.model.ems import EmsRegisterGetter
-from givenergy_modbus.model.plant import Plant
+from givenergy_modbus.model.plant import _HV_BMU_BAND_END, Plant
 from givenergy_modbus.model.register import HR, IR, MR, Register
 from givenergy_modbus.model.register_cache import RegisterCache
 from givenergy_modbus.pdu import (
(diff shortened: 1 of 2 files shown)


dewet22-codex:
Starting a review now. I’ll focus on behavioural correctness, protocol/model invariants, and compatibility with downstream HASS/CLI consumers.

dewet22-codex (approved):
I’ve reviewed this and have no blocking concerns. The widened mock serial disambiguation now matches the full HV BMU module band while still filtering to populated caches, and it reuses the same module serial register layout used by the model.

I also ran the focused mock-plant integration tests locally: ...............................                                          [100%]
31 passed in 5.03s (31 passed), and  is clean.

dewet22-codex:
Small correction to the approval note above: local verification was `uv run --group test pytest tests/client/test_mock_plant_integration.py -q` (31 passed) and `uv lock --check` (clean).

PR #398, opened 2026-07-10
feat(client): probe_alive() — cheap reconnect liveness gate

dewet22:
Adds `Client.probe_alive(timeout=2.0, retries=0) -> bool` — a named liveness primitive for coordinators managing reconnects to a marginal/hung dongle (hass field diagnosis on AberDino's Gen-1 Gateway).

## Why a separate method, not `detect(retries=0)`

The two-phase reconnect pattern is: while hung, cheap repeated probe; on first sign of life, a full robust detect to re-establish topology. `detect(retries=0)` *looks* like it covers this (its HR0 identity read is the gating first step and aborts fast on a hung dongle), but it conflates the two phases: the identity read shares a single `retries`, so `retries=0` makes the **recovery** detect fragile — a single dropped HR0/HR21 frame on a just-recovered, still-flaky link aborts the whole detect and loses that recovery tick. You can't have one call be both the cheap fail-fast gate and the robust recovery. (The peripheral *sweep* is separately robustifiable via `probe_retries`; the identity read is the irreducible conflict.) So the clean shape is a dedicated cheap probe → then a normal `detect(retries=3)` that no longer carries the fail-fast constraint.

## Contract

Reads the exact `HR(0,60)@0x11` read `detect` uses (proven against real hardware — not a leaner untested `HR(0,1)`). Mirrors detect's connection discipline, returning a bool instead of raising:

- **Alive** (HR0 answered) → socket **left open**, `True` — the coordinator's follow-up `detect(retries=3)` reuses the live connection (as detect-success leaves the socket for `refresh`).
- **Not alive** (timeout / `CommunicationError` / a response that left no usable HR0) → socket **closed** (detect's #274 teardown), `False` — releasing it for the dongle's quiet window, so the coordinator uniformly `connect()` → `probe_alive()` each tick with no reprobe-same-socket branch. Never raises.

Reconnect cadence/backoff stays coordinator-side (#356); the library owns only the single check. No new reconnect loop.

## Tests

`tests/client/test_detect.py` — alive→True + socket stays open; hung→False (not raised) + socket closed; answered-but-no-HR0→False + closed; the read is `HR(0,60)@0x11` with the caller's retries. Full suite 1627 green, tox green.

Rides the pending 2.12.0. hass will be notified of the exact signature to wire their reconnect loop.

### DIFF
--- a/givenergy_modbus/client/client.py
+++ b/givenergy_modbus/client/client.py
@@ -855,6 +855,49 @@ class Client:
                 _logger.exception("detect: error during connection teardown after failure")
             raise
 
+    async def probe_alive(self, timeout: float = 2.0, retries: int = 0) -> bool:
+        """Cheap reconnect liveness gate: read HR(0) once and report whether the inverter answered.
+
+        Distinct from :meth:`detect` by design. Its ONLY job is "is the inverter responding at
+        all?", so on a hung dongle it fails fast (default ``retries=0``, no peripheral sweep) and
+        its failure is free — the caller just probes again next tick. On the first success the
+        caller runs a full, robust ``detect(retries=3)`` to re-establish topology; that detect
+        stays robust because it no longer carries the probe's fail-fast constraint (which is the
+        whole reason this is a separate method rather than ``detect(retries=0)`` — one call cannot
+        be both the cheap gate and the robust recovery, since the identity read shares a single
+        ``retries``).
+
+        Mirrors :meth:`detect`'s connection discipline, returning a bool instead of raising:
+
+        - **Alive** (HR(0) came back): the socket is left **open**, so the caller's follow-up
+          ``detect`` reuses the same live connection.
+        - **Not alive** (timeout / :class:`CommunicationError`, or a response that left no usable
+          HR(0)): the socket is **closed** (mirroring detect's #274 teardown), releasing it for the
+          dongle's quiet window so the coordinator uniformly reconnects next tick. Never raises.
+
+        Reuses the exact HR(0,60)@0x11 read :meth:`detect` uses (the read proven against real
+        hardware). Reconnect cadence/backoff stays the caller's concern (#356): this owns only the
+        single check.
+        """
+        try:
+            await self.send_request_and_await_response(
+                ReadHoldingRegistersRequest(base_register=0, register_count=60, device_address=0x11),
+                timeout=timeout,
+                retries=retries,
+            )
+            alive = self.plant.register_caches.get(0x11, RegisterCache()).get(HR(0)) is not None
+        except (TimeoutError, CommunicationError):
+            alive = False
+        if not alive:
+            # Release the socket on any not-alive outcome, mirroring detect()'s teardown so a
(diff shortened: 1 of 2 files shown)


dewet22-codex:
Starting a review now. I’ll focus on behavioural correctness, protocol/model invariants, and compatibility with downstream HASS/CLI consumers.

givenergy_modbus/client/client.py
```diff
@@ -855,6 +855,49 @@ async def detect(
...
+        hardware). Reconnect cadence/backoff stays the caller's concern (#356): this owns only the
+        single check.
+        """
+        try:
+            await self.send_request_and_await_response(
+                ReadHoldingRegistersRequest(base_register=0, register_count=60, device_address=0x11),
+                timeout=timeout,
+                retries=retries,
+            )
+            alive = self.plant.register_caches.get(0x11, RegisterCache()).get(HR(0)) is not None
```
dewet22-codex: [R] Blocking: this checks the whole plant cache after the read, so a stale HR0 from an earlier healthy connection can make a fresh non-committing probe look alive. One concrete path is: detect/refresh has already populated `0x11/HR0`; the dongle then returns an all-zero full HR bank, which `_commit_bank()` rejects over non-zero cache while `send_request_and_await_response()` still returns the response; `probe_alive()` then sees the old HR0 and returns `True`, leaving the socket open. That breaks the “answered-but-no-usable-HR0 => False + close” contract in the reconnect path. I think this needs to derive liveness from the fresh response itself, or otherwise prove/stamp that this probe committed HR0, rather than reading any existing cache value.

dewet22-codex (changes requested):
I found one blocking liveness edge case around stale HR0 cache state making a fresh non-committing probe look alive. The focused tests otherwise pass locally, and CI is green, so I’ve kept the review to that single behaviour.

givenergy_modbus/client/client.py
dewet22-claude: Good catch — the stale-cache false-positive was real, and it's exactly the answered-but-no-usable-HR0 case my test missed (it only covered a never-populated cache). Fixed in 609c23a: liveness now comes from whether this probe re-stamped the HR(0,60) block's ingestion timestamp, which is set only on a successful _commit_bank — so it delegates the all-zero/CRC/error rejection to the one place that owns those guards, rather than reading a value that may predate the probe. Added a regression covering the stale-good-HR0 + rejected-fresh-all-zero-bank path.

dewet22-codex:
Starting a focused re-review now. I’ll check the stale-cache HR0 liveness blocker and any directly related changes.

dewet22-codex (approved):
I have re-reviewed 609c23a and the stale-cache liveness blocker is resolved. The probe now keys success off this HR(0,60) read re-stamping the block after a committed update, so a stale cached HR0 no longer makes a non-committing fresh read look alive.

Focused local checks passed: tests/client/test_detect.py (79 passed) and uv lock --check. Codecov is still reporting patch/project coverage red for the guarded teardown logging branch, but I do not see a remaining blocking behavioural concern in the code changes.

PR #401, opened 2026-07-12
feat: Plant.last_updated_at freshness signal + coalesce reader-EOF churn

dewet22:
Two small, independent additions that both come out of the passive-mode /
marginal-dongle work with givenergy-hass. Neither changes existing behaviour on
a healthy setup.

## `Plant.last_updated_at` (freshness signal)

`Plant.last_updated_at` returns the newest `register_block_updated_at` across
all committed banks, or `None` before the first commit. It's the honest "is the
cache being kept fresh" signal for a passive consumer: it advances whenever any
bank commits — solicited or fanned-out from a peer client — regardless of
content, so a quiet bus shows up as a signal that stops moving.

Because it keys off commit rather than value, it stays meaningful on a Gateway
(where the synthetic all-zeros inverter makes `system_time` spurious) and is
topology-agnostic. Age is consumer-derived (`now - last_updated_at`), so there's
no baked-in threshold.

hass asked for this to gate passive-mode freshness — without it, a passive tick
marks success on a stale cache and their energy integrals keep accumulating from
a frozen power value.

## Coalesce reader-EOF reconnect churn (hass#95)

A newer-chipset AIO dongle idle-reaps the TCP connection roughly every ~10s when
no client→dongle frame goes out in that window. The client reconnects instantly
and the fan-out resumes, so it's functionally harmless — but it emits thousands
of `connection lost (reader at EOF)` WARNING lines/day on a working setup
(~7000/day in the capture we looked at).

`_note_eof_drop()` throttles that log #355-style: the first drop warns, repeats
within a 5-minute window drop to DEBUG, and one re-warn per window carries the
running tally. A drop after a long quiet spell re-warns (the last-warn timestamp
has aged out), so a genuinely fresh disconnect is never swallowed — only
sustained churn is throttled. The counter lives on the `Client` so it survives
the reaped-and-reconnected sockets.

I've treated the ~10s reap itself as dongle-side and benign rather than something
to fight with keepalives — the data pointed at a rolling idle timer, and loading
a fragile dongle with continuous traffic to keep it quiet felt like the wrong
trade. This PR just stops the noise; the passive freshness signal above is what
makes the reap harmless to consumers regardless.

Both changes are test-covered (4 + 5 new tests); full suite and tox green.

## Summary by CodeRabbit

* **New Features**
  * Added a plant freshness indicator showing when register data was last updated.
* **Bug Fixes**
  * Improved handling of repeated connection drops by reducing duplicate warning messages while retaining useful diagnostics.
  * Connection loss caused by an unexpected end-of-file is now reported more clearly.
* **Tests**
  * Added coverage for plant freshness tracking and connection-drop logging behaviour.

### DIFF
--- a/givenergy_modbus/client/client.py
+++ b/givenergy_modbus/client/client.py
@@ -7,7 +7,7 @@ import warnings
 from asyncio import Future, Queue, StreamReader, StreamWriter, Task
 from collections.abc import Callable
 from dataclasses import dataclass
-from datetime import datetime
+from datetime import UTC, datetime
 from typing import Literal
 
 from givenergy_modbus.exceptions import (
@@ -79,6 +79,13 @@ _FRAME_SENT_MIN_TIMEOUT = 5.0
 # means the peer stopped ACKing — a half-open connection — so it's treated as
 # connection loss rather than left to wedge the producer until close() (hass#233).
 _DRAIN_TIMEOUT = 10.0
+
+# Re-warn cadence for the coalesced reader-EOF reconnect churn. A marginal dongle that
+# idle-reaps the TCP connection every ~10s (hass#95) would otherwise emit thousands of
+# 'connection lost (reader at EOF)' WARNING lines/day on a setup that is working fine —
+# each drop transparently recovers. The first drop warns; subsequent drops within this
+# window are demoted to DEBUG; one re-warn per window carries the running tally.
+_EOF_REWARN_SECONDS = 300.0
 
 
 class FrameRedactor:
@@ -497,6 +504,11 @@ class Client:
         self.expected_responses = {}
         self._shutting_down = False
         self._connection_lost = False
+        # Reader-EOF reconnect-churn coalescing state (#355-style, see _note_eof_drop). Persists
+        # across reconnects — a marginal dongle drops repeatedly, so the burst must be tracked on
+        # the long-lived Client, not per-connection.
+        self._eof_drop_count = 0
+        self._eof_last_warn_at: datetime | None = None
         self.network_producer_task: Task | None = None
         self.network_consumer_task: Task | None = None
         # self.debug_frames = {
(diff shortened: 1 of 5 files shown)


dewet22:
**Review notes**

Two small, independent, additive changes. Both follow existing conventions and are test-covered. Net +155/-2, 9 new tests.

**`Plant.last_updated_at`** — newest `register_block_updated_at` across committed banks, `None` when cold. Keys off commit time rather than value, so it stays honest on a Gateway (synthetic all-zeros inverter makes `system_time` spurious) and is topology-agnostic. Age is consumer-derived, no baked-in threshold. Survives `redact_serials()`/`model_copy()` since `register_block_updated_at` is explicitly copied (`plant.py:932`).

**`_note_eof_drop()`** — #355-style throttle for the "connection lost (reader at EOF)" warning an AIO dongle idle-reaps ~7000×/day. First drop warns; repeats inside a 300s window drop to DEBUG; one re-warn per window carries the tally; a drop after a long quiet spell re-warns fresh. State lives on the `Client`, not per-connection, so it survives reconnects. Mirrors the splice-burst shape at `plant.py:1591`.

**What landed well**
- Both follow established patterns (`block_age`/`_stamp_block`, `_splice_burst_note_reject`) — same window/threshold discipline, same WARNING-vs-DEBUG split.
- Test coverage is thorough: EOF side has first-drop, in-window coalesce, window-boundary re-warn (asserts `count == 3` at exactly 300s), post-quiet re-warn, plus a real-consumer integration test. Freshness side covers cold, newest-across-banks, monotonic advance, discarded-bank exclusion.
- Both commits are `feat:` → auto-classify into "✨ Added" at release time. No `docs/usage.md` change needed (no `commands.py` touch).

**Minor observations (non-blocking)**
- `last_updated_at` is O(n) per call; n is ~tens of banks polled at ~1Hz, so negligible. Flagging only in case someone later calls it in a hot loop.
- Throttle uses wall-clock, not `time.monotonic()` — consistent with the splice precedent, and a backwards NTP jump only briefly over-throttles.
- Two features in one PR; cleanly separated into two standalone commits if a split were ever wanted.

No blockers from me. Waiting on `test` / `Analyze` CI to confirm green before merge.

dewet22-codex:
Starting a review now. I’ll focus on behavioural correctness, invariants around the Modbus/client contracts, and cross-repo compatibility with the HASS/CLI consumers.

givenergy_modbus/client/client.py
```diff
@@ -1570,7 +1603,18 @@ async def _task_network_consumer(self):
             _logger.debug("network_consumer exiting on intentional shutdown")
         else:
             self.connected = False
-            _logger.warning("network_consumer: connection lost (reader at EOF)")
+            warn_now, count = self._note_eof_drop(datetime.now(UTC))
+            if warn_now and count > 1:
+                _logger.warning(
+                    "network_consumer: connection lost (reader at EOF) — %d drops in the last ~%ds, "
```
dewet22-codex: [Y] Non-blocking: this tally can include old coalesced drops from before a long quiet period. For example: warn at t0, coalesce drops at t+10s and t+20s, then no EOFs until t+1h; the next warning reports 3 drops in the last ~300s even though only the current drop happened in that window. The recovery behaviour is still fine, but I think either resetting the count when the quiet gap has elapsed, or wording this as “since the previous warning”, would keep the diagnostic honest.

dewet22-codex (approved):
I have reviewed the latest head and have no blocking concerns. The new Plant.last_updated_at property matches the committed-bank timestamp contract HASS needs for passive freshness, and the EOF coalescing keeps the existing connection-loss teardown behaviour while reducing repeated warning noise.

I left one non-blocking diagnostics note about the EOF tally wording/count after a long quiet gap. Focused local checks passed: tests/client/test_client.py, tests/model/test_plant.py, targeted EOF/freshness tests, and uv lock --check. Functional CI is green; only the Codecov threshold checks are failing.

givenergy_modbus/client/client.py
dewet22: Good catch — fixed in d234973. The tally now closes the burst when the gap since the last drop exceeds the re-warn window, so a lone drop after a long quiet spell warns fresh (count 1) rather than escalating with a stale tally. Also reworded the escalation line to "N drops since the previous warning" so it no longer claims a fixed ~300s window. Added a regression test asserting the post-quiet drop reports 1, not the stale 3.

PR #404, opened 2026-07-13
test: property-based fuzzing for framer and codec

dewet22:
Adds a hypothesis-based property-test module over the framer and codec — the library's network trust boundary. This came out of a look at where the test suite's weight actually sits: the framer/`decode_bytes`/codec path is directly network-exposed and already carries the #88 DoS-hardening, but was covered only by fixed examples. Property tests express its invariants over all inputs instead.

## What it covers

**Framer (bytes → messages)**
- `decode` never escapes as an unhandled exception — the #88 trust-boundary invariant
- marker-less garbage never grows the buffer without bound (anti-DoS)
- chunking invariance: however a stream is fragmented across reads, the same messages decode
- leading non-marker garbage is transparent to the following frame

**Codec (messages ↔ bytes)**
- read- and write-register response encode→decode round-trip fidelity, exercising the 16-bit register and string de/serialisers over arbitrary payloads

## Approach

Frame generation is seeded from the real capture corpus (403 frames mined from `tests/fixtures/captures`) and mutated around it. Random bytes almost never contain the `0x59590001` start marker, so pure-random generation gets discarded before reaching the interesting decode paths; seeding from real frames biases the generator toward structurally-valid input so mutation actually reaches `decode_bytes` and the PDU decoders.

All six properties hold at 5,000 examples each (~20k framer inputs plus ~10k round-trip PDUs). No defects surfaced — the expected-good outcome for a fuzz pass over well-exercised code; the lasting value is the regression net for future changes to the framer/codec.

## Notes

- Example count is env-driven (`GE_FUZZ_EXAMPLES`, default 300 for a fast PR pass) so a deeper sweep needs no code change.
- The module carries its own `pytest.mark.timeout(120)` because the suite-wide 1s per-test limit is aimed at unit tests — a deeper sweep would otherwise trip it mid-example and surface as a confusing `FlakyFailure` rather than an honest timeout.
- `hypothesis` added to the `test` dependency group.
- I haven't wired a nightly deeper-sweep CI job in this PR — can add it here or leave it as a follow-up.

One thing I deliberately kept out of scope: the round-trip tests exercise the wire codec, not the model-layer register *interpretation* (scale/signedness conversions). That seam is already guarded by `test_register_conventions`, so I didn't duplicate it here.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,6 +42,7 @@ dev = [
 ]
 test = [
     "bandit>=1.7.5,<2",
+    "hypothesis>=6.100,<7",
     "mypy>=2.0.0,<3",
     "pytest>=9.0.0,<10",
     "pytest-asyncio>=1.0.0,<2",
--- a/tests/test_framer_properties.py
+++ b/tests/test_framer_properties.py
@@ -1,0 +1,279 @@
+"""Property-based fuzzing of the framer's trust boundary.
+
+The framer is the library's network trust boundary: :meth:`Framer.decode`
+consumes raw, potentially-hostile bytes and must never raise, never grow its
+buffer without bound, and always yield the same messages regardless of how the
+byte stream is chunked (#88). Those are *properties* over all inputs, not a
+handful of examples, so we assert them with Hypothesis.
+
+Generation is biased towards structurally-valid frames — random bytes almost
+never contain the ``0x59590001`` start marker and so get discarded before
+reaching the interesting decode paths. We seed from the real capture corpus
+(``tests/fixtures/captures``) and mutate around it, which drives coverage into
+``decode_bytes`` and the PDU decoders.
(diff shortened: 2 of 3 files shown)


dewet22-codex:
Starting a review now. I’ll focus on behavioural correctness, Modbus/client invariants, and downstream compatibility with the HASS/CLI consumers.

dewet22-codex (approved):
I’ve reviewed this head and have no blocking concerns. The new properties are scoped to the framer trust boundary and codec round-trips, Hypothesis stays in the test dependency group, and the local focused plus broader checks were clean. Gemini’s field-assertion and event-loop notes look reasonable as possible polish, but I don’t see them as blockers: byte-level idempotence already covers those fields, and the suite remains fast at the default depth.

PR #405, opened 2026-07-13
ci: nightly deep fuzz sweep for framer/codec properties

dewet22:
Follow-up to #404. That PR added the property tests but they only run at a shallow example count (`GE_FUZZ_EXAMPLES` default 300) so per-PR CI stays fast — which meant the deeper sweep only ever ran when someone bumped the count by hand locally. This wires it into CI so it actually runs somewhere.

## What it adds

- **`tox -e fuzz` env** — runs just `tests/test_framer_properties.py`, cranks `GE_FUZZ_EXAMPLES` (default 5,000, overridable), no coverage. Usable locally too: `GE_FUZZ_EXAMPLES=20000 tox -e fuzz`.
- **`Nightly Fuzz` workflow** — 04:30 UTC daily (offset from `dev.yml`'s weekly 06:30 Monday run) plus manual dispatch with a custom example count.

A nightly run explores fresh random inputs against whatever landed on `main` that day, so a regression in the framer/codec trust boundary surfaces within a day rather than a week (or whenever someone next fuzzes locally).

## Notes

- **Kept separate from `dev.yml` on purpose.** Its cadence differs — nightly here vs the weekly cross-OS/newer-Python safety net there. Folding it in would mean gating every job on which cron fired, which reads worse than a small dedicated file.
- **No injection surface.** The `workflow_dispatch` examples input is bound through `env:` (not interpolated into `run:`) and consumed only as an `int()`-parsed value, so a malformed input fails closed with `ValueError`.
- Read-only permissions, pinned action SHAs — matches `dev.yml`.
- Verified locally: `tox -e fuzz` at the default 5,000 runs all six properties green in ~28s.

The nightly cron won't fire until this is on `main`; the `workflow_dispatch` trigger lets it be run on demand from the Actions tab once merged.

### DIFF
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -1,0 +1,39 @@
+name: Nightly Fuzz
+
+on:
+  schedule:
+    # Nightly deep property-based fuzz sweep of the framer + codec (04:30 UTC,
+    # offset from dev.yml's weekly 06:30 Monday run). Per-PR/weekly CI runs these
+    # property tests at a shallow example count for speed; this runs them deep
+    # against the latest main so a regression in the wire trust boundary surfaces
+    # within a day of landing rather than a week.
+    - cron: '30 4 * * *'
+
+  # Allows a manual run (optionally with a custom example count) from the Actions tab.
+  workflow_dispatch:
+    inputs:
+      examples:
+        description: 'Hypothesis examples per property'
+        required: false
+        default: '5000'
+
+# Least privilege: read-only, like dev.yml. No coverage upload here.
+permissions:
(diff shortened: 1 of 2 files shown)


dewet22-codex:

dewet22-codex (approved):
I have reviewed this head and have no blocking concerns. The new fuzz tox env runs the property module at the intended deeper default, and the workflow keeps the manual example-count input in env rather than interpolating it into shell. CI is green, and local checks passed: uv run --group test tox -e fuzz, uv run --group test pytest tests/test_framer_properties.py, and uv lock --check.

PR #408, opened 2026-07-14
test: add first HYBRID_GEN2 golden-master fixture (hass#293)

dewet22:
Our first-ever **Gen2** device capture — crg-n's 3.6 hybrid from [hass#293](https://github.com/dewet22/givenergy-home-assistant/issues/293). Adds it as a golden-master fixture so Gen2, previously untested, now has regression coverage.

## What it pins

- **Classification** — `0x2003` / ARM `920` → `Model.HYBRID_GEN2` at `0x11` (the Gen2 discriminator is `arm_fw // 100 == 9`).
- **The 2.12.2 routing fix** — daily battery energy at **alt1** (`IR36`/`IR37`) → 4.7 / 5.5 kWh, with GEN1's alt2 daily reading 0.0. This is the live evidence behind #406.
- **Deferred total** — lifetime battery-energy total stays an honest `None`; the alt1 total register reads an implausible 0 and the alt2/alt3 `HR(4100+)` candidates are never polled.
- **`IR44 == PV generation` on GEN2** — the DC-coupled PV fields are real and sun-shaped, so the manifest hybrid field-identity holds on Gen2, not just Gen1. That's the main de-risking value here.

## Characterisation flag (not asserted-correct)

This two-string PV unit reports `num_mppt == 0` and `num_phases == 0`. I've pinned it so a later decode change trips the test, but I can't yet tell whether that's a Gen2 decode gap or GE simply reporting 0 here — it needs a second Gen2 to disambiguate. Called out rather than silently accepted.

## Redaction

Serials were already redacted at source (`givenergy-cli capture` zeroes the trailing unit digits). I re-ran `FrameRedactor` over the frames as belt-and-braces (a no-op here) and verified **zero** unredacted serials remain — envelope (inverter + dongle) and every register serial-group. Manufacture dates (prefix + YYWW) are preserved, per the existing fixture convention.

## Not included (possible follow-up)

I kept this to the offline golden-master. Wiring the capture into the live full-cycle MockPlant fence (`test_mock_plant_integration.py`) would give the strongest Gen2 coverage, but a 60s passive single-cycle capture likely won't answer detect()'s full request sequence, so it'd need the `RefreshPartiallySucceeded` tolerance path — worth its own look rather than bundling here.

Test-only; full suite + tox green.


dewet22-codex:

dewet22-codex (approved):
I have reviewed this head and have no blocking concerns. The fixture replays through the real framer and Plant.update path, pins HYBRID_GEN2 classification and the #406 daily-energy routing, and keeps the unconfirmed lifetime totals as None. I also checked the committed capture for non-redacted serial-like tokens and found none.

Local checks passed: uv run --group test pytest tests/model/test_fixture_golden_master.py, uv run --group test ruff check tests/model/test_fixture_golden_master.py, uv lock --check. CI is green too.

PR #414, opened 2026-08-03
fix: suppress v_pv*/i_pv* where PV is metered AC-side (hass#281)

dewet22:
Raised from [givenergy-hass#281](https://github.com/dewet22/givenergy-hass/issues/281): an AC-coupled unit has no DC string, yet reports a live, climbing `p_pv1`. The power figure is real — but the accompanying *voltage* is mains, not a string voltage. Same class of quiet mislabelling as IR44 before #267: plausible-looking and wrong.

`v_pv1`/`v_pv2`/`i_pv1`/`i_pv2` now decode to `None` on `Model.AC` and `Model.ALL_IN_ONE`, via a new `FIELD_IDENTITY` row reusing the mechanism `ir44_inverter_out` already established.

The **power and energy** members are deliberately untouched. They carry real generation and are the only such figure an AC-metered install has.

## Evidence

The AIO case is conclusive. Across all 49 states of the `aio_a` fw612 capture:

- `v_pv1` is **bit-identical** to `v_ac1` — not "within about a volt", the same value — tracking together across nine distinct voltages (245.5 → 246.2).
- `i_pv1` takes just **two** distinct values (3.0, 4.0) against **24** distinct `p_pv1` values (860–971 W). So `p_pv1 ≠ v_pv1 × i_pv1`; 245.7 × 3.0 = 737 W against a reported 862 W.

Neither is a coherent string measurement. The DC controls behave completely differently — V×I reconciles with `p_pv1` on GEN1, GEN2 and HYBRID_HV_GEN3, and `v_pv1` sits 43–142 V clear of mains:

| model | v_pv1 | v_ac1 | delta | p/(V×I) |
|---|---|---|---|---|
| GEN1 hybrid | 372.2 | 243.5 | 128.7 | ~1.0 |
| HYBRID_HV_GEN3 | 127.6 | 244.8 | −117.2 | 1.02 |
| **ALL_IN_ONE** | **245.7** | **245.7** | **0.00** | **1.17** |

Both directions are pinned, plus the raw `IR(1) == IR(5)` equality on the fixture so a later capture can contradict the reading rather than having to argue with a derived field.

## Two scope calls worth flagging

**`Model.AC` is included on a structural argument, not corpus evidence.** An AC-coupled unit has no DC string for these registers to describe, which I find convincing on its own. But every AC capture we hold reads `p_pv1 == 0` — across 690 and 825 samples at both `0x11` and `0x31`, there is not a single generating AC sample. So our fixtures cannot corroborate it directly. The one live sample (a GIV-AC3.0 on hass#281) has `v_pv1` ~1.6 V from mains rather than exactly equal, which is a weaker signature than the AIO's exact match. Flagging in case that gap matters more to a reviewer than it does to me.

**`AC_3PH` is deliberately absent despite being `is_ac_coupled`.** The row is applied by `SinglePhaseInverter._apply_field_identity`, and `ThreePhaseInverter` carries no model validators at all — so listing `AC_3PH` would read as handled while doing nothing. Extending the routing to three-phase is its own piece of work; I'd rather leave the gap visible than paper it over.

## Behaviour change

Consumers reading `v_pv1`/`v_pv2`/`i_pv1`/`i_pv2` on AC or AIO models will now get `None` where they previously got a mains-derived number. Worth an upgrade note on the release.

## Not established

The metering path — whether the figure comes from a generation CT on a separate solar inverter's output or from something internal — is still unknown, and I haven't guessed at it. It matters because it determines whether a zero reading means "no generation right now" or "no CT fitted". The pinned `i_pv1`-at-3.0 A is a clue but not an answer.

## Verification

tox green (py311, py314, format, lint); 1665 tests pass.

### DIFF
--- a/givenergy_modbus/model/inverter.py
+++ b/givenergy_modbus/model/inverter.py
@@ -1065,7 +1065,7 @@ class SinglePhaseInverter(  # type: ignore[valid-type,misc]
         source field actually carries one, and re-validating a dumped model (where
         e_pv_generation_* is already None) passes the moved values through untouched.
         """
-        from givenergy_modbus.model.manifest import ir44_is_inverter_output
+        from givenergy_modbus.model.manifest import ir44_is_inverter_output, pv_string_vi_is_ac_derived
 
         if not isinstance(values, dict):
             return values
@@ -1085,6 +1085,13 @@ class SinglePhaseInverter(  # type: ignore[valid-type,misc]
                 if values.get(src) is not None:
                     values[dst] = values.pop(src)
                     values[src] = None
+        if pv_string_vi_is_ac_derived(model, arm_fw=int(arm_fw)):
+            # Suppressed rather than relocated: v_ac1 already carries the voltage these
+            # registers echo, so there is no destination field worth adding. The power
+            # and energy members of the PV family are left alone — they are the only
+            # generation figure an AC-metered install has (hass#281).
+            for field in ("v_pv1", "v_pv2", "i_pv1", "i_pv2"):
+                values[field] = None
         return values
 
     def _battery_energy(self, direction: str, metric: str) -> float | None:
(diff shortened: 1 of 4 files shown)


dewet22-codex (approved):
I’ve reviewed the model-specific field routing and the evidence behind suppressing the AC-derived V/I readings while retaining the power and energy figures. The implementation and coverage look sound to me. I also ran the full 1,665-test suite locally, together with the format, lint, type, security, package and documentation build checks; all passed.

PR #415, opened 2026-08-03
test: assert the HV Gen3 fixture's decoded-frame count and CRC cleanliness

dewet22:
Post-merge review catch on #412 (CodeRabbit reviewed after the merge landed).

`test_every_response_frame_decodes` claimed all 17 response frames decode, but only asserted the device topology. A frame dropped for an already-populated device leaves both the device set and the populated count untouched — so the fence would have stayed green while the fixture quietly tested less. That's the same hole as the #371 `load_config` fence, and a fence that doesn't bite is worse than no fence because it reads as reassurance.

It now counts decoded responses directly.

## One correction to the review's framing, worth recording

CodeRabbit suggested folding this into a single verified count. That isn't sufficient on its own: **a bad CRC does not reduce the decoded count.** The PDU still decodes and is flagged `crc_failed`, with the guard refusing the commit downstream. I checked rather than assumed:

| capture state | (decoded, errors, crc_failed) | caught by |
|---|---|---|
| pristine | (17, 0, 0) | — |
| one frame removed | (16, 0, 0) | count |
| one CRC byte flipped | (17, 0, **1**) | CRC assertion only |

So counting alone would have missed a botched CRC regen — precisely what the fixture README's "clean" claim rules out. CRC is asserted separately for that reason, with the reasoning in the docstring so the next person doesn't collapse the two.

The README frame count is disambiguated to 17 requests + 17 responses; it read 34 while the test said 17, which is the inconsistency that surfaced this.

## Verification

tox green (py314, format, lint); 1660 tests pass. Both failure modes confirmed to trip the fence by hand before committing.

### DIFF
--- a/tests/fixtures/captures/hybrid_hv_gen3_a/README.md
+++ b/tests/fixtures/captures/hybrid_hv_gen3_a/README.md
@@ -7,7 +7,7 @@ in [`../README.md`](../README.md).
 
 | File | Vantage |
 |---|---|
-| `givhy80g3hv_hass295_120s.log` | HYBRID_HV_GEN3 passive HA capture, 120 s (34 frames) |
+| `givhy80g3hv_hass295_120s.log` | HYBRID_HV_GEN3 passive HA capture, 120 s (34 frames: 17 requests + 17 responses) |
 
 ## Topology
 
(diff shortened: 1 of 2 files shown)


dewet22-codex (approved):
I reviewed the decoded-frame fence and the separate CRC-cleanliness assertion. The split is warranted: removing a response reduced the result to (16, 0, 0), while flipping a CRC byte retained 17 decoded responses and raised the CRC-failure count to 1. I also ran the full 1,660-test suite locally, together with the format, lint, type, security, package and documentation build checks; all passed.

